PR #46, opened 2020-06-10
Katacoda click to load

josh-heyer:
Demonstrate a couple of different options for Katacoda embedding:

1. Click to load for the full UI, inline
2. Click to load for the "panel" UI (terminal window pinned to the bottom of the page)

### DIFF
--- a/src/components/katacoda.js
+++ b/src/components/katacoda.js
@@ -1,9 +1,11 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { Helmet } from 'react-helmet';
 
 const Katacoda = ({account, 
         scenario, 
-        "data-katacoda-layout": layout, 
+        clickToShowText = '',
+        panel = false,
+        layout = undefined, 
         port = undefined, 
         lang = undefined, 
         token = undefined, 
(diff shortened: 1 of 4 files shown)


src/components/katacoda.js
```diff
@@ -57,15 +61,34 @@ const Katacoda = ({account,
...
+                    e.dataset.lang = "shell";
+                window.katacoda.init();
+            }
+        };
+
     return <>
         <Helmet>
             <script src="https://katacoda.com/embed.js" />
         </Helmet>
+        <button type="button" onClick={initKata} className={`btn btn-secondary ${!shownClass && 'd-none'}`}>{clickToShowText}</button>
```
edabot, line 86: This hides the button, but can be a problem for the panel type. The button hides when you open the panel. But if you close the panel, the button is still hidden and there's no way to open it again without refreshing the page.

josh-heyer: Unfortunately, this butts into how Katacoda handles "closing" the panel - it destroys it. So while we can re-show it using another call to `window.katacoda.init()`, that also re-creates the environment (e.g., you start over). 

It might be possible to override this behavior & actually just *hide* the panel (similar to how I'm hiding the embedded environment with click-to-show), but it'd require deeper hooks into Katacoda's implementation, and resulting fragility.

bradnoble:
Can the button stay visible and onclick either builds or destroys katacoda?

josh-heyer:
It could, yeah. Is that something that'd be helpful?

bradnoble:
to me, it seems nice to have the trigger stay visible and to only load
katacoda on request

PR #60, opened 2020-07-14
Rs tweaks wo docs

edabot:


cero-miedo:
dupe, with content removed

PR #69, opened 2020-08-04
add additional metadata: product, platform and tags

josh-heyer:
For the exploratory work that @epbarger is doing


epbarger:
@josh-heyer This is great, thanks! This should be enough for me to mess with for a bit.

So I was thinking that `product` would be one of `ark, bart, cds, efm, epas, pem`, so that the products match up to existing docs products. But maybe that's not super useful? Platform tag is awesome - likely we'd want to add that to docs as well.

josh-heyer:
Platform might be more useful in the future - just don't have any EDB-platform-specific topics here right now.

PR #71, opened 2020-08-17
basic issue templates for problems & requests

josh-heyer:
Instead of putting these in contribute, include them in the GH UI. 

TODO: link to these *from* contribute


josh-heyer:
(Pushing immediately so that I can grab links)

PR #73, opened 2020-08-18
Link to issue report templates w/ page path prefilled in title on topic pages

josh-heyer:
The goal here is to make it simple for folks to report issues (or suggest changes) on specific pages by providing links to issue templates at the bottom of the page:

Know of a way this page could be better? Report a problem or suggest an addition!

Those two links contain the path (not URL) of the page being viewed prepopulated as the title, along with references to the [specific template](https://github.com/rocketinsights/edb_docs_advocacy/tree/master/.github/ISSUE_TEMPLATE) with which to pre-fill the body of the report.

@epbarger I'd like your input on the approach here - I'm pretty much just copying the logic for the "Edit this page" button, but there may be a better way of including this (relatively simple) logic.

@cero-miedo any opinions on the (currently extremely sparse) design?

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -187,12 +187,17 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
       'https://github.com/rocketinsights/edb_docs_advocacy/edit/master/advocacy_docs' +
       doc.fields.path +
       (doc.fileAbsolutePath.includes('index.mdx') ? '/index.mdx' : '.mdx');
+    const githubIssuesLink = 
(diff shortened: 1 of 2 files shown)


epbarger:
@josh-heyer this looks good to me! Could you submit this as a PR to edb_docs though? Otherwise I'll have to copy and paste these files to edb_docs manually, which is fine, but I don't want to forget about it :). Once that PR is merged into edb_docs I can sync docs into advocacy.

josh-heyer:
Moved to https://github.com/rocketinsights/edb_docs/pull/134 per @epbarger 's request

PR #83, opened 2020-09-14
New Katacoda components

epbarger:


josh-heyer (approved):
Good to merge so far as I can see!

PR #97, opened 2020-10-08
FYI component

cero-miedo:
`<ForYourInfo>The vast majority of the discussion will focus on features, functionality, and techniques that apply equally to both PostgreSQL and EDB Postgres Advanced Server (EPAS); however, it will also touch on a couple of features that are only available in EPAS. These will be clearly noted.</ForYourInfo>`

MDX component to enable messages to be played to users at the top of articles.


cero-miedo:
Outdated: Closing

PR #148, opened 2020-11-30
added vagrant vm scripts

leo-jin-edb:
This is a PR for an auto-provisioned VM for launching EDB documentation server locally in development mode.  The value of this PR is to provide a frictionless way for other developers who are not familiar with the edb docs architecture to install the server and immediately be able to work on their docs without having to understand the technical details.  

I suspect we'll need another iteration to fully realize the value proposition described above.  For example, it would be nice to allow developers to drop their md or mdx files into a shared folder (between VM guest and the host) on the host and be able to see the server pick it up on the VM guest and render the user facing HTML.   I didn't want to march further down this route without proper consultation and feedback from the doc team first.


package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "clean": "gatsby clean",
     "clean-develop": "gatsby clean && gatsby develop",
-    "develop": "gatsby develop",
+    "develop": "gatsby develop -H 0.0.0.0",
```
epbarger, line 9: This will expose the application to the whole network, correct? Is there a way to avoid doing that?

epbarger (commented):
I haven't used Vagrant in a while so I can't say I'm super familiar with the configuration, but this all looks good to me! I love the idea of streamlining the setup process for people.

Just curious, is there a reason you're reaching for Vagrant over Docker here? I was previously thinking about introducing a Dockerfile since that seems like it would be quite straightforward for this app (especially with the existing github action workflows). I'm wondering if the performance would be better over Vagrant too - spinning up `gatsby develop` with the product docs set loaded is reasonably intensive. Vagrant seems like it's up to the task too though

package.json
leo-jin-edb: ah good point, this is not a problem within a vm but it'd be a problem if someone would to run this directly on the host machine.   I could instead add a new script target?  maybe `develop-expose` or `develop-vm`, something descriptive.  or we can just run gatsby directly.

leo-jin-edb:
@josh-heyer  hey great question.  Here's a generally accepted POV - vagrant is better suited for local development due to its ability to support fast file sharing such as nfs (**use case:** write code using vscode on host machine but execute on the vagrant box), better environment isolation at development to minimize chances of host corruption, among other developer specific features.  Docker allows file sharing as well via volumes but vagrant has more local dev focused features.  I think we need both vagrant and dockerfile where:

- docker can be used across CI/CD, as well as used as a local service for developing EDB docs. 
- vagrant can be used by EDB core application devs, plugin devs and contributors like myself.

So more or less similar technology but targeting different audiences.

Here's a blog that explains further and better than me 😄  https://medium.com/faun/vagrant-vs-docker-which-one-to-use-for-development-122c99ad563f

package.json
leo-jin-edb: awesome, thanks for the feed back, please see this commit. https://github.com/EnterpriseDB/docs/pull/148/commits/6b495d35695b18ee974717c4d46c2f2ab3999873

.vm_provision/playbook/roles/xdoc/tasks/main.yml
```diff
@@ -0,0 +1,40 @@
+---
+- name: Clone xdoc source code
```
josh-heyer, line 2: This is the only bit that really bugs me: right now to get this going, you have to clone the repo onto your host, only for it to be then cloned *again* in the VM. That's inefficient, both of time and of storage.

Worse: it doesn't work twice. If you try to reprovision the VM, this will fail. 

A few ideas here...

- Provide a script to pull down JUST the vagrantfile + provisioning scripts from the repo (faster to get going)
- Instead of cloning here, just copy everything from `/vagrant` (but ONLY IF `~/code` doesn't already exist) then git fetch / pull to update it.
- Just work entirely in `/vagrant`? More notes on that later...

josh-heyer (changes requested):
I spent way too much time playing with this this evening, testing it against both VirtualBox and Hyper-V (since I'm on Windows, got both anyway). 

Pretty solid little setup. Only big concern I noted inline - would really like to avoid maintaining two copies of the repo for each instance. My first thought was that it'd make sense to just work entirely out of /vagrant (the share of the repo as it was pulled on the host machine) - but YMMV here; perf does take a hit. One other annoyance there is that file change notifications don't get propagated, so even though you can edit on the host gatsby won't automatically rebuild - which negates a lot of the advantage of working in a synced folder to begin with. I did stumble across [a plugin that purports to address this](https://stackoverflow.com/questions/21190584/vagrant-shared-folder-take-advantage-of-inotify-over-nfs) but didn't mess with it to see if it works.

Which approach makes sense here may depend largely on how folks use it. It's certainly comfortable enough to just work entirely within the VM right now, as long as you don't try to work on the host as well and expect that to do anything. The simple solution might just be "download this one script, let it do the rest" - that'd also make for a great demo!

OH, one last note: I found this neat trick over on SO - when sshing in, it's possible to forward a port, thus eliminating the need to run gatsby against all interfaces:

```
vagrant ssh -- -L 8000:localhost:8000
```

.vm_provision/playbook/roles/xdoc/tasks/main.yml
leo-jin-edb: @josh-heyer thanks a lot for the feedback, let me address them and submit changes.

epbarger:
To continue with that @josh-heyer said last night... personally I would consider being forced to solely use the filesystem inside vagrant a non-starter. Ideally I wouldn't have to change my development practices at all, besides actually running things inside vagrant.

I'll give setting this up a shot next update!

leo-jin-edb:
@epbarger @josh-heyer Thank you both again for the excellent constructive feedback.  This is why I love getting PRs up for review as soon as possible so we can have dialogs like this.    I made several changes with the latest commit. 

I garnered two issues reading through your feedback and both are excellent suggestions.

1.  Pulling source code inside of the VM makes no sense for we already have the source code locally.

2.  When I make an edit in my editor on the host, I should see gastby pick it up in dev mode and automatically recompile modified assets with latest changes on the guest machine.

Solutions:

1.  For issue #1,  instead of pulling from git, I modified the shared folder settings in "vagrantfile" to sync to the root folder of the xdoc project on host with folder `/home/vagrant/code` on guest.

**Pro**:  No more cloning inside the VM again, we simply use the already cloned source code on host;  No changes needed to your existing local dev flow, just make changes on host and see it appear in guest.

**Con** Initial provisioning for the box takes longer now because `yarn` is executed against a shared folder that requires every operation to be synced back to the host.  With NFS protocol, the initial `vagrant up` took about 3 - 5 min on a macbook pro. 

2.  For issue #2, thanks to @josh-heyer's suggestion, I was able to make this work using the  'vagrant-notify-forwarder' plugin and I've listed this plugin as part of the vagrantfile.   To test this, do the following:

1. `rm -rf node_modules && rm -rf .vagrant` (this is not needed if it's a fresh clone)
2. `vagrant destory` to kill the current vm instance
3. `vagrant up` to reload the vm with the plugin.
4. wait for it to finish
5. on host type `yarn run vm-ssh` to run the vagrant ssh command with forwarded 8000 port.  (another good find by @josh-heyer !!)
6. once you logged in `cd code` and `yarn develop`
7. after the server is up, make a change in vscode and see the server automatically reload. 

**Con** The only con I see is the initial `yarn develop` launch takes a bit longer due to file syncing but once it's launched, subsequent recompile is pretty fast.

gsong:
@leo-jin-edb Please excuse my mid-way intrusion into this thread. It seems the primary goal for this PR is to minimize the friction of contribution for new developers to the project, right?

I believe both Docker and Vagrant can achieve that goal (feel free to disagree).

With that in mind, do you feel there are significant differences in adopting a Docker based solution vs. a Vagrant based solution? Are EDB developers more familiar with one vs. the other, or it would be about the same from a knowledge / experience perspective?

leo-jin-edb:
@gsong  hi, Josh asked the same thing, here's my previous reply.

> @josh-heyer hey great question. Here's a generally accepted POV - vagrant is better suited for local development due to its ability to support fast file sharing such as nfs (**use case:** write code using vscode on host machine but execute on the vagrant box), better environment isolation at development to minimize chances of host corruption, among other development specific features. Docker allows file sharing as well via volumes but vagrant just has more local dev focused features. I think we need both vagrant and dockerfile where:
> 
> * docker can be used across CI/CD, as well as used as a local service for developing EDB docs.
> * vagrant can be used by EDB doc core application devs (actually work on the nodejs source code), plugin devs and contributors like myself.
> 
> So more or less similar technology but targeting different audiences.
> 
> Here's a blog that explains further and better than me 😄 https://medium.com/faun/vagrant-vs-docker-which-one-to-use-for-development-122c99ad563f

.vm_provision/playbook/roles/xdoc/tasks/main.yml
leo-jin-edb: I've removed this, see the newest commit

PR #154, opened 2020-12-04
Fix capitalisation of "Barman"

amenonsen:
The Barman docs use "Barman" consistently. I think there might be some remnant uses of "BaRMan" in external webpages, but we shouldn't propagate it.


amenonsen:
Duplicates #153, closing.

PR #164, opened 2020-12-15
Content conversion of latest product releases, plus tidy up on content issues

cero-miedo:
* EPAS 13
* EFM 4.1
* PEM 8.0


epbarger (approved):
Built fine locally, looks good!

PR #542, opened 2021-01-15
Update sidebar navigation based on new IA

matt-edb:
Based on: https://trello.com/b/fYNGgukt/product-portfolio-for-edb-docs-20-launching-end-of-january?menu=filter&filter=label:For%20launch%20Jan%2028

Note: 
* Some nav items may need stub pages that link out to external sites (e.g. PostgreSQL, pgAdmin, etc).
* `EDB Connectors` is now an aggregation of the 4 previous discrete connectors
* `Foreign Data Wrappers` only includes Hadoop (but there may be more?) but not sure if we need a more generic index page for this section that lists just Hadoop
* Slony and Ark are now omitted, not sure how intentional that was
* Now that we have `PostgreSQL` under `Databases`, do we need the persistent `PostgreSQL Docs` link anymore?
* I don't know what `source` means in this file, so it was omitted in some cases where I was creating new sections


epbarger:
`source` is used to show/hide the link depending on if a specific source is loaded or not. eg `source: 'epas'` would only show the link if epas is loaded. You can see all the source names if you look in `build-sources.json`. If you don't want to hide/show something you can just omit the `source` key altogether, like you've done

epbarger (approved):
This looks good to me! Besides some of the links being non-functional, which I assume is intended for now.

Issue #421, opened 2021-01-07
Feedback on /epas/13/epas_qs_windows/index.mdx


sheet20:
Image name is missing at the bottom for all the images on this page and all the images must be centered.

PR #589, opened 2021-01-19
Image name added to the bottom of the image

sheet20:
Image name that was missing added to the bottom of the image


cero-miedo:
Fixes #421

cero-miedo (approved):

PR #670, opened 2021-01-22
clean up various markdown tables so that they parse properly

josh-heyer:
GFM tables require pipe characters in cells to be escaped EVEN IF WITHIN INLINE CODE; Ref: https://github.github.com/gfm/#example-200

The juicer scripts did not do this correctly; this resulted in tables with implicit columns, which the renderer would silently drop. Round-tripping through the parser -> stringifier would retain the columns, but now the table would look odd with cell contents split across columns and headers not matching up. 

This PR attempts to mitigate this by escaping pipe characters correctly; all changes are introduced by a script (normalizeTables.js) that looks for this specific issue, along with a related issue where pipe characters were apparently used to indicate paragraph breaks.

Need this to be in before I run the full markdown normalization - the script for that (normalizeMarkdown.js) is included for reference, though the changes it will introduce are not.


package.json
```diff
@@ -71,7 +71,17 @@
     "prettier": "^1.19.1",
     "pretty-quick": "^3.1.0",
     "react-refresh": "^0.8.3",
-    "typescript": "^3.9.3"
+    "typescript": "^3.9.3",
```
epbarger, line 74: these should be alphabetized ideally

scripts/normalizeMarkdown.js
```diff
@@ -0,0 +1,387 @@
+// run: node scripts/normalizeMarkdown.js "product_docs/**/*.mdx"
```
epbarger, line 1: I think it would be nice if these scripts were in a `scripts/normalization` directory or something similar, what do you think?

epbarger: We should also extend the precommit hook in package.json to cover these files

epbarger (approved):
LGTM, I would just make sure prettier touches these files for future commits :)

josh-heyer:
Reasonable suggestions - done!
And yeah, once we have this mess under our belts, doing [some] of this automatically would be a good idea

PR #673, opened 2021-01-22
Normalize Markdown

josh-heyer:
There are a lot of changes here; you might want to spot-check a few, but I wouldn't worry about reading all of them. The purpose of this branch is to change a lot of files without changing anything substantial in the output, to prepare for a subsequent set of changes that *will* change the output.

Lemme back up a bit...

## The point of this monstrocity

Most of these MDX files were generated originally by Pandoc, then modified with a set of scripts in a process colloquially known as the "juicer". Its aim was to bring content from various sources of varying provenance and allow them to be treated the same. 

One problem: the ultimate destination for these files is Gatsby, which transforms them into HTML using a toolset known as MDXjs which in turn is built on another known as "remark". And Remark's idea of what proper Markdown looks like is... Slightly different from that of Pandoc. Not *extremely* different, just different enough that there *could* be problems if we aren't careful.
   
And, in fact, there are problems.

Fortunately, a lot of those problems can be fixed rather easily. Indeed, some of them can be fixed *automatically* with a bit of care - that Remark tool allows reading and writing Markdown files, just like Pandoc - so it'd be super convenient to just put a little filter in between the "reader" and the "writer" bit and fix up stuff that way.

Except... The first time through that pipeline, every file *changes*. Maybe just a little, maybe a lot. Mostly just whitespace that doesn't have any real effect, in some cases something more substantial. 

So before I go doing any of that, I fed all of the product MDX files through Remark, just reading them in & writing them back out again, until the results were stable. The script I used is included - scripts/normalize/markdown.js. 

## Verification

I tested this by running a production build before making any changes, capturing the HTML output, then running the normalization, building again, and comparing the HTML generated by both builds. I iterated on this until I got the differences down to a reasonable set of 29 files containing a few minor whitespace and list numbering changes, mostly the latter. I reviewed these by hand, and the results looked fine to me: making numbered lists actual lists without repeated numbers seems like a positive change.

Related: #670

## Sanity check

I'd appreciate anyone with a bit of time looking at the scripts I wrote for this; they're pretty quick'n'dirty, so more eyes looking for mistakes is probably a good thing.

For the content itself, I'm mostly looking for spot-checks: if you scan down the list of changes & see anything that looks dodgy, wrong, or that you just don't understand... Call it out. Otherwise, don't stress about it - there's no point in trying to go through 3K files by hand, that's why I used the verification approach detailed above.

I'm gonna probably let this bake until Monday, then merge it in so I can finally fix all those broken links.

## Reference material

- [The CommonMark spec](https://spec.commonmark.org/current/)
- [The Github-Flavored Markdown spec](https://github.github.com/gfm/)


product_docs/docs/ark/3.4/ark_admin/04_ark_registering_with_pem.mdx
```diff
@@ -43,28 +42,26 @@ If you select `ENABLED`:
 
 -   Any Amazon Role or Azure Group that is accessible by the Ark user that is deploying the console is added to the PEM server as a group role. The PEM role created for the project, role, or group is not a login role. To simplify management, each role that represents a project, role, or group is also a member of the PEM `ark-team` group role.
 
-![*The PEM Group Role dialog*](images/ark_registering_with_pem_group_role_dialog.png)
+![\*The PEM Group Role dialog\*](images/ark_registering_with_pem_group_role_dialog.png)
```
cero-miedo, line 45: Introduces backslash characters inside image ALT tags

product_docs/docs/ark/3.4/ark_admin/11_ark_securing_ark.mdx
```diff
@@ -57,16 +52,14 @@ Before using the private key, you must modify the permissions on the keyfile. Us
 
 > `chmod 0600 <ssh_key_file>.pem`
 
-Where &lt;ssh\_key\_file&gt;.pem specifies the complete path and name of the EDB Ark ssh private key file.
+Where &lt;ssh_key_file>.pem specifies the complete path and name of the EDB Ark ssh private key file.
```
cero-miedo, line 55: Looks like we're catching `&gt;` but missing `&lt;`

epbarger:
So GitHub doesn't show any of the scripts because of the 3000 file limit, so I'm reviewing the scripts locally.

*all scripts*
- In general your code style is pretty different than the typical node code I read. I don't think this is a problem at all right now, and I'd rather you keep things your way since you're the only one in here right now, and this is pretty "deep" work where reorganizing too much right now might hurt more than it helps. Still, before this is put aside, I think it would be worth linting and cleaning some stuff up. Things like the deeply nested functions within functions, usage of var and `function` when arrow functions would work, consistent whitespace, etc. Overall the scripts are still quite readable as is though.

* in `markdown`*
- maybe the comments on 43-47 are no longer needed if you're taking argument input? if they're still helpful leave them though
- I would move the inline `require('remark-stringify')` on line 38 to the top with the other includes

* in `tables`*
- it would be nice if there was more description around the regexes and replacement. Particularly lines 77-81 are a bit hard to understand what exactly is happening. Maybe a well named function could help? Also an option to turn the regexes into well named constants, like the `tableFmtRegez`

*in `mast-embedded-hast`*
- I would move module.exports to the bottom where most people would expect it.
- I think moving the long conditional on line 30 into a function with a clear name would be helpful. You also may be able to use some `||` to clean this up a bit. IE `(hast.children || []).length > 1` or similar.

epbarger (approved):

product_docs/docs/ark/3.4/ark_admin/11_ark_securing_ark.mdx
josh-heyer: Doesn't matter; HTML in markdown has to start with less-than; if it doesn't, it's parsed as Markdown. Ref: [HTML blocks](https://spec.commonmark.org/0.29/#html-blocks), [raw HTML](https://spec.commonmark.org/0.29/#raw-html)

The writer here avoids a lot of unnecessary escapes - note also the lack of backslash escaping on the underscores - this was actually kinda necessary in the original Markdown, but everyone hated it so most parsers will avoid it somehow; CommonMark lays out some sensible rules, but the examples are even better [intraword emphasis examples](https://spec.commonmark.org/0.29/#example-373)

product_docs/docs/ark/3.4/ark_admin/04_ark_registering_with_pem.mdx
josh-heyer: So, this is strickly-speaking unnecessary: the spec allows formatting in alt text, but recommends that they be ignored (stripped during render). The parser here apparently attempts to preserve them as *text*, under the assumption that they were intended to be literal. 

Of course, they weren't - this was a caption, not alt text, and the asterisks were intended to italicize the caption. So... We probably need two decisions here:

- Do we want to allow formatting in captions? (we could, uh, just style captions however we want) - if so, they should be preserved so that they can be re-rendered properly by whatever.
- CAN we render formatting in captions? (this is a question for @epbarger ) - if not, I should probably just strip this entirely.

Meanwhile, I'll take a look at what's there; I should be able to block escaping, but may not want to if we're gonna just strip these.

josh-heyer: Ok, these are no longer escaped. We can decide down the road what we want to do with them.

josh-heyer:
Yeah, quick'n'dirty script is dirty ;-) 
I'll clean these up progressively if we keep using them; made the corrections to specific issues / suggestions Evan brought up.

PR #697, opened 2021-01-25
Updated links in footer

matt-edb:
Put actual links in the footer instead of dead links.

### DIFF
--- a/src/components/footer.js
+++ b/src/components/footer.js
@@ -1,5 +1,4 @@
 import React from 'react';
-import { Link } from 'gatsby';
 
 const TimestampLink = ({ timestamp, githubFileLink }) => {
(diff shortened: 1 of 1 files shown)


epbarger (changes requested):
Because these links are going outside of the app, the `<Link>` component should become a regular `<a>` tag. For example, the first line should become `<a className="text-muted mx-2" href="https://www.enterprisedb.com/gdpr"> GDPR </a>`

Because `<Link>` will end up unused, you can remove the import by deleting the entire `import { Link } from './';` line

matt-edb:
Okay, proper link format is now in place and <Link> import removed.

epbarger (approved):
thanks Matt!

PR #699, opened 2021-01-25
add note about syncing with the develop branch

josh-heyer:
This may need some elaboration, lemme know what you think @shannon7wallace


josh-heyer:
Eh @gsong you wanna have a look here & drop some wisdom into it?

cero-miedo (approved):

PR #704, opened 2021-01-25
<Link> improvements

epbarger:
- make protocol relative urls use `<a>` tags
- force trailing backslash outside of `URL` to avoid browser specific issues


josh-heyer (approved):
Looks good, this should get around the issues I was running into trying to add the slash to the relative path.

PR #705, opened 2021-01-25
preparing for link rewriting

josh-heyer:
This is another run of the normalization introduced by #673 - it catches a few files that were updated or added since then. I've tested this locally, it produces no differences in the final output.


josh-heyer:
Shotgunning this through so I can get links fixed!

PR #706, opened 2021-01-25
[WRONG] rewrite relative links

josh-heyer:
This fixes a bunch of broken links, as well as rewriting all relative links in the product docs to comply with the relative path strategy introduced by #703


josh-heyer:
Ran (and committed!) an old version of the script!

PR #707, opened 2021-01-26
rewrite relative links

josh-heyer:
This fixes a bunch of broken links, as well as rewriting all relative links in the product docs to comply with the relative path strategy introduced by #703

Fixes several distinct classes of bugs:

1. target IDs that were mixed-case (valid in RST, not valid in HTML)
2. IDs that refer to headings (Gatsby follows Github's strategy for generating these IDs from the heading text, while RST did something else)
3. IDs that reference a target or heading in a sibling file 
4. A combination of 1 or 2 and 3
5. incorrect relative paths
6. formerly-correct relative paths that are broken by #703 (and would also have been broken with JavaScript disabled)

Most of these links can be tested right here on GitHub: page-relative links will work without fuss, while others will require you to edit the path, replacing the last "/" with ".mdx". Only links intended to go to an index.mdx file will require you to guess at this fact.

But I've tested them all locally, and will do so again after merging this.

Heads-up @cero-miedo this may cause conflicts with work being done locally.


epbarger (approved):
Looks great, really awesome work!

PR #749, opened 2021-01-27
adding stubs for open source offerings

bradnoble:
- postgresql
- barman
- pgbackrest
- pgadmin

links from the home page will also work


epbarger (approved):
Looks good to me!

PR #752, opened 2021-01-27
Update README.md

matt-edb:
Made instructions for installation and setup clearer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,37 +1,60 @@
-# <img src="static/images/edb-docs-logo-disk-dark.svg" alt='EDB Docs' width="350">
+<img src="static/images/edb-docs-logo-disk-dark.svg" alt='EDB Docs' width="200">
 
 ![Deploy Main to Netlify](https://github.com/EnterpriseDB/docs/workflows/Deploy%20Main%20to%20Netlify/badge.svg)
 ![Deploy Develop to Netlify](https://github.com/EnterpriseDB/docs/workflows/Deploy%20Develop%20to%20Netlify/badge.svg)
 ![Update PDFs on Develop](https://github.com/EnterpriseDB/docs/workflows/Update%20PDFs%20on%20Develop/badge.svg)
 
-This repo contains the Gatsby application that powers EDB's documentation website, as well as the advocacy content. The site pulls additional content from other repos, in a process called 'sourcing'.
+This repo contains the React/Gatsby application that powers the EDB Docs website. The site pulls [Markdown](https://www.markdownguide.org/) content from several repos in a process called "sourcing", and then renders it all into high-performance markup. You can install the application on your local computer for easy editing, viewing, and eventually publishing to the GitHub repo.
 
-## Installation
-1. (MacOS) Install the [homebrew package manager](https://brew.sh/), if it's not already installed.
-2. Clone the repo with GitHub desktop, or on the command line with `git clone https://github.com/EnterpriseDB/docs.git`. If you do not have git installed, install it with (MacOS) `brew install git`.
-    - If you have not set up your local git with GitHub, you may want to setup an SSH key on your computer to authenticate you to GitHub. See [GitHub's SSH Docs](https://docs.github.com/en/github-ae@latest/github/authenticating-to-github/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent) for help doing that!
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,37 +1,60 @@
...
 
-## Running Without a Local Installation (Recommended for Windows)
+1. Install [Homebrew](https://brew.sh/), if it's not already installed. (Use `brew -v` to check.)
 
-If you wish to work with Docs without installing the prerequesites locally, you can do so from within a Docker container using VSCode. See: [Working on Docs in a Docker container using VSCode](README_DOCKER_VSCODE.md)
+1. Install Git using Homebrew with `brew install git`, if it's not already installed.  (Use `git --version` to check.)
+
+1. Clone the [EDB Docs GitHub repo](https://github.com/EnterpriseDB/docs) to your local machine using [GitHub Desktop](https://desktop.github.com/), or via the Terminal with `git clone https://github.com/EnterpriseDB/docs.git`
+
+1. Set up an SSH key in GitHub, if you haven't done so already. (Go to [GitHub's SSH Keys page](https://github.com/settings/keys) to check.) If you don't have an SSH Key set up yet, you'll need to set one up to authenticate you to GitHub. See [GitHub's SSH docs](https://docs.github.com/en/github-ae@latest/github/authenticating-to-github/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent) for more information.
```
epbarger, line 18: If this is going to be a full number item, I think it should be above the `clone` step, and then the clone step should use `git clone git@github.com:EnterpriseDB/docs.git`, which would only work if they set their SSH key up correctly

README.md
```diff
@@ -1,37 +1,60 @@
...
+
+1. Install [Node.js version 14 LTS](https://nodejs.org/en/download/). We recommend using Node version 14 LTS (the Long Term Support release) as version 15 is not compatible with some of our dependencies at this time.
+
+    - If you already have Node installed, you can verify your version by running `node -v` in the cloned repo directory.
+
+    - If you already have Node version 15 installed, you may want to consider using Node Version Manager (NVM) for a simpler way to manage multiple versions of Node.js. Follow the [directions to install NVM](https://github.com/nvm-sh/nvm#installing-and-updating), then run `nvm install` in the cloned repo directory, followed by `nvm use 14` to use version 14 LTS.
```
epbarger, line 24: Should this just be "a different version of node" instead of Node 15? Also I think suggesting `nvm use` vs `nvm use 14` makes more sense, because `nvm use` is aware of what version the project needs, vs `nvm use 14` just always loading version 14. Not a big deal, but I'm sure eventually we'll update to 16 lts

README.md
```diff
@@ -1,37 +1,60 @@
...
+
+1. Install Python 3 with `brew install python3`, if it's not already installed. (Use `python3 -V` to check that you have version 3.6 or higher.) Python is not needed for the core Gatsby system, but is required by several source scripts.
+
+1. Navigate to the cloned repo directory in your Terminal, if you haven't already done so.
```
epbarger, line 28: This should be higher up, probably right after cloning the repo

README.md
```diff
@@ -1,37 +1,60 @@
...
+
+1. Install Yarn with `npm i -g yarn`. Yarn is the package manager we're using for this project, instead of NPM.
+
+1. Install Gatsby with `npm i -g gatsby-cli`. Gatsby is the software that powers the EDB Docs site.
+
+1. Install all required packages by running `yarn`.
+
+1. Pull the shared icon files down with `git submodule update --init`
+
+1. Select which sources you want with `yarn config-sources`, and then running `yarn pull-sources` if prompted (see section below for details)
```
josh-heyer, line 38: I believe explicitly running yarn pull-sources is no longer necessary (@epbarger ?) - might benefit more here from a bit of mildly prescriptive guidance, such as 
```suggestion
1. Select which sources you want with `yarn config-sources` - this will configure the products that will be built, and automatically pull them down. To speed up the next steps, choose only the products you intend to test or edit - you can change these options at any time (see below for details)
```

josh-heyer (approved):
Looks good to me. Just one comment (inline)

README.md
epbarger: Yeah the wording isn't quite right, `yarn config-sources` will launch right into `pull-sources` if needed, but you don't need to explicitly run `pull-sources` unless you're manually futzing around

matt-edb:
Okay, I've updated the README based on your feedback. Thanks guys!

PR #782, opened 2021-01-29
html tables to markdown tables

josh-heyer:
New plugin for normalization script: table-formatter.js

Re-writes HTML tables as Markdown, if it thinks it can get away with it. Pretty-prints the HTML otherwise.

Currently, this means it will avoid rewriting tables that have colspan / rowspan on cells / rows. I was actually pleasantly surprised by how rare this is - just one instance in the current versions. (bart/2.6.1/bart_user/03_using_bart/02_managing_backups_using_a_retention_policy/04_managing_the_backups_based_on_the_retention_policy.mdx)

This PR includes the scripts, and the re-written files that exist in current versions. It can be run against older versions as well (and I've tested this) but that's a lot more to review. 

I've taken steps to preserve paragraphs in cells, by replacing them with `<br /><br />`. However, this WILL strip out the following:

- blockquote formatting
- list formatting
- other block-level formatting
- other HTML that has no representation in Markdown
- column widths
- custom CSS styling
- any IDs on or within cells **this needs to be watched for carefully** - it may break links somewhere else in the page!


cero-miedo (approved):
This looks awesome. Going to test with recent content changes.

Thanks so much for this.

PR #807, opened 2021-01-29
Make index link targets bigger

matt-edb:


epbarger (approved):
LGTM. Next time you need to make a change here, ping me and we'll build a component so you don't need to duplicate styles everywhere

PR #809, opened 2021-01-30
All links to EDB Docs 1.0 from current versions rewritten

### LINKED ISSUE #801: Some generic links that still point to /edb-docs but shouldn't
See the first link on this page, which takes the user to the EFM docs on Docs 1.0:
https://www.enterprisedb.com/docs/efm/latest/efm_quick_start/

The link isn't tied to an older version, so there's no reason to send the user to Docs 1.0.

Maybe we can catch and rewrite outbound links that include `/edb-docs` in the URL but don't include a version.

josh-heyer:
As I expected, this was a fair bit of manual work. However, it wasn't as bad as I had feared: due to the work that @epbarger put in on the redirects, I was able to re-write most deep links without significant effort. This took care of 47 links.

Shallow links - those that simply pointed to /edb-docs - were a bit more work. In the majority of cases, the intended context for the link was represented in the text that preceded it; either in the same sentence, or in a preceding paragraph. I extracted this text, and with the help of a few infernal regexps was able to compile a short list of titles for various documentation sections; I then looked these up and built a table of mappings between the titles and paths within our documentation. This took care of another 103 links.

This left about 13 links that I had to figure out by hand and re-write. 

I also re-wrote the *text* of links where it contained `https://www.enterprisedb.com/edb-docs` to point to `https://www.enterprisedb.com/docs`

The script which assisted me in this task (along with Evan's redirects) can be found in scripts/normalization/lib/docs1-link-rewriter.js - it runs as a part of my normalization script, along with the checker/rewriter for fragment-links, table conversion, etc.

I ran this on the current versions found in `/product_docs`; it can be run on older product versions, but will need similar hand-holding or manual work to complete.

This PR fixes #801


scripts/normalize/lib/docs1-link-rewriter.js
```diff
@@ -0,0 +1,5204 @@
...
+    if (result.length) return result.join(' ');
+
+    // nothing there? Go up a level and do the same thing, but only look back one node
+    if (ancestors.length > 1)
+      return getContext(parent, ancestors.slice(0, -1), 1);
+    return "";
+  }
+}
+
+const titleMap = {
```
epbarger, line 152: I should have thought to send you my redirects metadata file, it has this information in a fairly similar format  😬

scripts/normalize/lib/docs1-link-rewriter.js
```diff
@@ -0,0 +1,5204 @@
...
+  'BART Installation Guide': '/bart/latest/bart_inst/',
+  'BART Installation and Upgrade Guide': '/bart/latest/bart_inst/',
+  "PEM Administrator's Guide": '/pem/latest/pem_admin/',
+  'PEM Agent Guide': '/pem/latest/pem_agent/',
+  'PostgreSQL core documentation': 'https://www.postgresql.org/docs/',
+};
+
+
+const urlMap = {
+  /* these are unique to this file */
```
epbarger, line 185: This works, but I wouldn't be upset if you wanted to break this out into a `.json` file specific for this script ;)

epbarger (approved):
LGTM, nice work

Issue #773, opened 2021-01-28
Consolidate sidenav links at bottom left (and the bottom of the site index)


bradnoble:
- PostgreSQL docs can be removed (in favor of the stub for PostgreSQL)
- Contribute and "authoring" need to be combined

bradnoble:
Live on staging

PR #822, opened 2021-02-01
Bn consolidate footer

bradnoble:
- #773


epbarger (approved):
Looks good to me! I took the liberty of removing the index file in the `supported-open-source` folder - I don't think actually want a file in there, since you can only navigate to it directly the way things are currently structured

Issue #844, opened 2021-02-02
https://www.enterprisedb.com/docs/?


marclinster:
## Summary
Search does not appear to be working on any of the pages. The page appears to reformat, but no serach results are displayed.
This happens on MacOS 10.15.7 with Safari  14.0.2

bradnoble:
Recreated on Safari.

bradnoble:
hey @marclinster, this is updated to work on Safari. Please verify when you have a minute, and close if you agree.

epbarger:
Closing this since the other issue was resolved, and these ended up being the same root problem

Issue #843, opened 2021-02-02
https://www.enterprisedb.com/docs/efm/latest/


marclinster:
## Summary
The version drop down is not working (that appears to be the problem throughout the site)
This was observed on MacOS 10.5.7 with Safari 14.0.2

bradnoble:
Recreated on Safari.

bradnoble:
hey @marclinster, this is updated to work on Safari. Please verify when you have a minute, and close if you agree.

marclinster:
Works now

PR #847, opened 2021-02-02
Fix Safari JS issues

epbarger:
- Replace lookbehind regex in Link with simple string concat - Safari doesn't support lookbehind


josh-heyer (approved):
Solid!

bradnoble:
Addresses #844 and #843

PR #848, opened 2021-02-02
Fix duplicate path prefixing

epbarger:
It seems that by using a custom component for links in the mdx, the pathPrefix (`/docs`) gets applied twice. Found this issue referenced [here](https://github.com/ChristopherBiscardi/gatsby-mdx/issues/377). Simple solution is to strip the pathPrefix in our custom <Link> component.

This PR also changes `yarn build` and `yarn serve-build` to use the `--prefix-paths` option, which should make issues like this easier to catch in local builds. I think this also makes sense from a "minimize differences between dev and prod" perspective too.


josh-heyer (approved):
Looks reasonable (didn't test). 
This means that we can write either, `[epas link](/docs/epas/latest)` or `[epas link](/epas/latest)` and get the same results *as long as we never change the link prefix*, but the former will break if we do (and, I assume, locally for `gatsby develop`) - is that right?

epbarger:
Yeah that's correct. We shouldn't be using `/docs` as a prefix for relative links though, since like you said that removes the flexibility of changing the prefix, and wouldn't work in development

PR #850, opened 2021-02-02
Add admonitions, and basic example page in playground

epbarger:
May need more styling, I just went with the most basic bootstrap styling and a simple default icon to replace the default

Screen Shot 2021-02-02 at 4 01 21 PM

### DIFF
--- a/advocacy_docs/playground/1/01_examples/admonitions.mdx
+++ b/advocacy_docs/playground/1/01_examples/admonitions.mdx
@@ -1,0 +1,23 @@
+---
+title: Admonitions
+---
+
+!!!important IMPORTANT TITLE HERE
+You can have [markdown](www.google.com) in here, neat
+!!!
+
+!!!tip TIP TITLE HERE
+You can have [markdown](www.google.com) in here, neat
+!!!
+
+!!!note NOTE TITLE HERE
+You can have [markdown](www.google.com) in here, neat
+!!!
+
+!!!warning WARNING TITLE HERE
+You can have [markdown](www.google.com) in here, neat
+!!!
+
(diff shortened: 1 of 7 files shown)


epbarger:
Used the info icon from https://github.com/primer/octicons/tree/master/icons . I assume we'll probably want to use our own icons, but this is an example of how to do the custom icons.

josh-heyer:
Tested it for a bit, seems to work fairly well. Only one concern: the syntax is slightly different from MKDocs / python-markdown, which uses indentation to capture the block:

```
!!!Note "This is an outrage!"
   Details on the outrageousness

   Whatever else

Not part of the note
```

josh-heyer:
See the commit I just pushed. It adds some more tests and references a forked plugin: https://github.com/josh-heyer/remark-admonitions

FWIW, the repo I was trying to remember earlier is https://github.com/EnterpriseDB/cloud-native-postgresql

The two plugins in use are Admonitions, and definition lists

epbarger:
Really nice work @josh-heyer! IMO it may make sense to give ownership of your forked repo to EDB, unless you plan to try and get your changes merged into the original repo. I'll leave that to you though

@cero-miedo anything in particular styling wise you want to do here? I'm assuming we don't actually want all of these in use, should we pick the 2 or 3 we want to have available and style them more if necessary?

cero-miedo:
@epbarger Looking at the styling of this now

josh-heyer:
FWIW: the keywords currently used in the Cloud Native Operator docs are: note, important, hint, warning, seealso

cero-miedo:
I've made my tweaks. Feel free to merge!

cero-miedo (approved):

PR #852, opened 2021-02-02
Fix the tables I broke

josh-heyer:
Due to a bug in Remark, pipe characters in code in table cells were not escaped properly (and are not un-escaped properly when parsed). This resulted in me setting a trap for myself in #782 - they were converted without escapes, leading to them becoming corrupt on the next normalization pass.

The first thing I did here was create a plugin for Remark to work around the poor handling of pipes-in-code. 

Then I manually re-converted the affected tables, with the plugin in place. Did a bit of other cleanup as well, as some of the affected tables had been modified.

...yeah, another PR with a bunch of random crap in it. See b9fa7bb for *just* the table correction.

Next step is to add the Remark plugin to Gatsby to fix the *rendering* side of things.


epbarger (approved):
interested to see this plugin!

epbarger:
Also this seems to contain a lot of "Note" fixes, I assume that's intended?

josh-heyer:
> 
> 
> Also this seems to contain a lot of "Note" fixes, I assume that's intended?

Yeah, I donno why but the format I used for notes markdown did NOT round-trip well. This parser is ... starting to irritate me :)

PR #853, opened 2021-02-02
work around bug in Remark wrt pipes in code in tables

josh-heyer:
This is a hack around https://github.com/remarkjs/remark/issues/583
Pipe characters within inline code in table cells need to be backslash escaped to avoid breaking the table, but, inline code parser doesn't know this, so gotta fix it up after the fact. The inline code stringifier apparently doesn't know that they need to be escaped either, so gotta fix that TOO.

This plugin does both of those: the transform step handles the former, while the monkeypatch to the compiler
handles the latter.

For Gatsby, only the former matters - but for the normalization scripts, the latter is also critical.

Both reference the plugin from src/plugins/code-in-tables.js so that I don't let the two diverge (hence the Remark plugin vs. a Gatsby Remark plugin)


cero-miedo (commented):
LGTM

PR #893, opened 2021-02-05
Rewrite notes as admonishments

josh-heyer:
This follows from #850 which adds "admonition" support: it reverses #828 as well as a number of manual edits to use this syntax.

Note: none of the normalization scripts can be used after this is merged, until I update the admonitions plugin to both parse *and* stringify admonitions. The standard remark parser sees these as *code*, and will seriously trash them if asked to write them out.


scripts/normalize/lib/table-formatter.js
```diff
@@ -69,16 +70,22 @@ function convertTable(hastTable)
 
   // this does all the heavy lifting
   mdast = hast2mdast(hastTable, options);
-  
-  // we kinda need paragraphs in cells; simulate with linebreak elements
+  const br = {type: 'jsx', value: '<br />'};
```
epbarger, line 73: Did you mean to check in these table-formatter changes?

epbarger (approved):
Looks good, nice work!

scripts/normalize/lib/table-formatter.js
josh-heyer: Yeah; every time I touch this I find something new I'd missed in the table conversion. This time around: some tables had nested *lists* - the new results are not awesome, but... Good enough for now, I think. Was half-way through the rewrite when I saw this, so... Eh.

Also found that some of the original tables had complicated formatting based on nested divs. That all got stripped (or, probably, wasn't generated at all) by the original pandoc import. Frankly don't know what to do about that, or if it's worth doing anything. Realized looking at the pgAdmin stuff yesterday that Docs 1.0 was using Sphinx, not pandoc - gotta figure that's probably the root cause of a lotta weirdness here.

PR #922, opened 2021-02-10
quick sketch of the form a kubernetes operator tutorial might take

josh-heyer:
Did this for a quick demo: https://drive.google.com/file/d/1FOEiavqJSo_SJ5p5FU3CuFPlPpy-kUny/view?usp=sharing

Not intended to be used directly, just a proof of concept. Sharing here for George so we're on the same page with what's been shared with that team.


josh-heyer:
Superseded by #1094

PR #926, opened 2021-02-11
import first pgBackRest docs

pgstef:
- Imported latest docs from pgbackrest-docs.
- Bumped latest release tag

Idea is to leave the docs here and remove the old pgbackrest-docs repo I guess.

Note: I'm not sure that the images in the use cases are rendered correctly.


cero-miedo (approved):
All looks good to me 👍 

I'm having problems merging this on the CLI, I think it's an oversight on my end. I'll double check with the team when they get online.

PR #935, opened 2021-02-15
Card deck enhancements: Frontmatter-configurable rendering, sort order

josh-heyer:
There are two rendering modes for the "card deck" index generated for pages in the advo section: simple and full, with the latter being used ONLY when exactly 4 directories deep. Three if you count not rendering one at all if less than 4 levels deep in the directory hierarchy. This is... A pain to remember. 

Also, the card decks ignore the sort order used in the sidebar.

Both of these keep tripping me up, so this PR attempts to fix them both.

1. Add an optional `indexMode` key to frontmatter, with 3 possible values: "none", "simple", "full". If omitted, current "guess based on depth" behavior remains; if specified, then:

	- none == no card deck
	- simple == card deck without icons or children
	- full == card deck with icons and children (if any)

2. Sort the things!

Combined, these provide a lot more control over and utility for index appearance.

before

after

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -83,6 +83,7 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
             iconName
             originalFilePath
             productStub
+            indexMode
             katacodaPages {
               scenario
               account
@@ -390,6 +391,13 @@ exports.createSchemaCustomization = ({ actions }) => {
 
     type Frontmatter {
       originalFilePath: String
+      indexMode: TileModes
+    }
+
+    enum TileModes {
+      none
+      simple
+      full
     }
   `;
   createTypes(typeDefs);
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -83,7 +83,7 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
             iconName
             originalFilePath
             productStub
-            indexMode
+            indexCards
             katacodaPages {
               scenario
               account
@@ -100,6 +100,7 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
           fields {
             docType
             path
+            depth
             product
             version
             topic
@@ -391,7 +392,7 @@ exports.createSchemaCustomization = ({ actions }) => {
 
     type Frontmatter {
       originalFilePath: String
-      indexMode: TileModes
+      indexCards: TileModes
     }
 
     enum TileModes {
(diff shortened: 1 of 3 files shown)


src/templates/learn-doc.js
```diff
@@ -35,30 +35,43 @@ const ContentRow = ({ children }) => (
...
+    Full: 'full',
+  };
+  if (mode === modes.None) return null;
+
   const { path } = mdx.fields;
-  const depth = path.split('/').length;
-  if (depth === 4) {
+
+  if (!mode) {
+    const depth = path.split('/').length;
```
epbarger, line 58: `depth` is actually available from graphql now as a field, so `node.fields.depth` should work here

src/templates/learn-doc.js
```diff
@@ -35,30 +35,43 @@ const ContentRow = ({ children }) => (
...
+    if (depth === 4) mode = modes.Full;
+    else if (depth >= 5) mode = modes.Simple;
+  }
+
+  if (Object.values(modes).includes(mode)) {
+    const colSize = mode === 'simple' ? 4 : 6;
```
epbarger, line 64: Maybe it makes sense to store these widths with the mode names? Maybe the default depth stuff too. Something like ...

`Full: { name: 'full', defaultDepth: 4, columnWidth: 6 }`

Just to keep the little bits "configuration" closer together. What do you think?

gatsby-node.js
```diff
@@ -83,6 +83,7 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
             iconName
             originalFilePath
             productStub
+            indexMode
```
epbarger, line 86: I think `indexCards` might be a better name for this, unless we anticipate latching on more behavior to this value. Otherwise it's probably nicer to have the flexibility

epbarger (approved):
Overall strategy looks good, just left some comments on possible improvements. Thanks for making this change :)

src/templates/learn-doc.js
josh-heyer: does if I add it to the query... Did both (page and nav) to keep them comparable.

src/templates/learn-doc.js
josh-heyer: Moved into CardDecks - this was a weird option to push up, only reason I can think of to change it is if design of the cards themselves change.

PR #948, opened 2021-02-16
aim for saner breaks on card decks

josh-heyer:
Eh, @cero-miedo mind weighing in on this? I'd just copied the existing breaks in #935 but... Didn't bother to check if they made sense. When I did, I noticed stuff like this:

3 unreadable columns

That's at a 1024px window size (Chrome, Windows 10) - kiiinda sketchy. Even at 1278 (half my larger screen) 3 columns is pushing it for longer words like "Recommended", but at least most of the text is readable. 

This PR specifies a couple of breaks for both simple and full cards to try & keep 'em readable as long as possible.

card-breaks

### DIFF
--- a/src/components/card-decks.js
+++ b/src/components/card-decks.js
@@ -62,7 +62,9 @@ const CardDecks = ({ cards, cardType = 'simple' }) => {
         return (
           <Col
             key={card.fields.path}
-            md={cardType === 'full' ? 6 : 4}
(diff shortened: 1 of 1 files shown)


epbarger (approved):
lgtm, @cero-miedo thoughts?

cero-miedo (approved):
👍

PR #964, opened 2021-02-19
Production Deploy: 2021-02-19

cero-miedo:
* content updates


epbarger:
just a note that I think we want to be following https://github.com/EnterpriseDB/docs/tree/develop/docs/proposals/branch-and-release#release-process in the future

PR #965, opened 2021-02-19
Cloud Native Operator, frontmatter `navigation`, and other gatsby-node work

epbarger:
- Add the cloud native operator docs at `/kubernetes/cloud_native_operator/`. Since we aren't pulling from their repo yet, I've included the raw docs in `temp_kubernetes`.
- Add CNO conversion script that can be run with `python3 scripts/source/source_cloud_native_operator.py`. This is converting the files to mdx, adding frontmatter (including the `navigation` frontmatter pulled from `mkdocs.yaml`), and fixing `yaml` links to work.
- Implemented `navigation` frontmatter, which replaces `left-navs.js`. This frontmatter lets you specify the order you want the navigation presented, as well as create sections.
- Replaced the majority of `navLinks` usage with new `navTree`, which powers the navigation. `navLinks` was a complete array of all navigation nodes under a specific product/version or advocacy topic, so it was a lot of data. Many of the components that used it needed to build their own tree from the data. This has all been stripped out in favor of a centralized `navTree` built during the `createPages` step of `gatsby-node`. The one remaining component that uses `navLinks` is `CardDesk`, since the `full` variant requires data that the current version of `navTree` does not have available. I'm definitely going to be working more on the `navTree` in the not-so-distant future, as there is more room to clean this stuff up.

test deployment @ https://edb-docs-test.netlify.app/docs/


gsong:
@epbarger The description sounds good. Is there something specific that you feel like needs a second (or third) pair of eyes on?

epbarger:
> @epbarger The description sounds good. Is there something specific that you feel like needs a second (or third) pair of eyes on?

Nothing I'm particularly concerned about, but all the action is in `gatsby-node` and `gatsby-node-utils` if you wanted to look in there

PR #966, opened 2021-02-19
Update PostgreSQL index.mdx

scott-edb:
New stub for PostgreSQL


scott-edb:
Note the comment for the final line, there's a reference to the EDB Language Installer Pack that needs a home

epbarger (approved):

PR #975, opened 2021-02-23
Basic Barman tutorial

josh-heyer:
The goal here is to demonstrate the most basic functionality of Barman via an interactive walk-through, covering installation, configuration, and core operations. Used a single-machine scenario to streamline the process, but admittedly this is NOT a recommended configuration; we need to be careful to make that clear.

If useful, this could be part of a series covering more advanced (and ... more realistic) scenarios.

Results can be viewed in the temp deployment here: https://edb-docs-adv-josh-barma-fcapab.herokuapp.com/supported-open-source/barman/01_single-server-streaming/

@sebasmannem I'd be interested in your feedback here, both in terms of utility and suitability.

@bradnoble @gsong I'd be interested in your feedback & suggestions in terms of flow, accessibility and general stability.

@epbarger small change you might be interested in in  src/components/code-block.js (I guess it's been a while since I've used multiple, executable languages in a single tutorial - single worked, comma-separated list had broken at some point)


epbarger (approved):
code related changes look good to me

gsong:
EDB_Docs_-_Backup_and_Recovery__Single-Server_Streaming

@matt-edb @cero-miedo Do you have suggestions of how to make the UX better here? My opinion:

1. The explainer text should probably read something like:
    > You can follow along the tutorial by [loading an interactive terminal]().

    Making the link act the same as the button. Also, would the user care what technology is backing the interactivity?
1. I'm not sure why we need a whole explainer to a button. The button should be self evident, especially since it should be the supplemental CTA (to the link in the explainer text).
1. I have to agree with @bradnoble that the button doesn't look like a button, it looks like a callout similar to the admonitions.

matt-edb:
Yep, agree. @cero-miedo can you update us on the status of this when you get a sec?

josh-heyer:
FWIW, I put the "Katacoda" mention in there mostly to help socialize other folks in the company as to what we're using ;-)
Can probably be dropped.

gsong:
@josh-heyer I feel like we're not really demonstrating the "streaming" nature of the backup. After the initial snapshot backup, I would make a change to the `actor` table, e.g. `UPDATE actor SET first_name='SANDY' WHERE actor_id=23;`.

Then when we restore, we should make sure that the row reflects the change. This will require [using the get-wal option](http://docs.pgbarman.org/release/2.12/#get-wal).

I can demonstrate this via Zoom.

josh-heyer:
Awesome feedback, @gsong - thanks! I'll go through & address these tomorrow morning.

cero-miedo:
> EDB_Docs_-_Backup_and_Recovery__Single-Server_Streaming
> 
> @matt-edb @cero-miedo Do you have suggestions of how to make the UX better here? My opinion:
> 
> 1. The explainer text should probably read something like:
>    > You can follow along the tutorial by loading an interactive terminal.
>    
>    
>    Making the link act the same as the button. Also, would the user care what technology is backing the interactivity?
> 2. I'm not sure why we need a whole explainer to a button. The button should be self-evident, especially since it should be the supplemental CTA (to the link in the explainer text).
> 3. I have to agree with @bradnoble that the button doesn't look like a button, it looks like a callout similar to the admonitions.

Having a text link feels like a regression to me. The button is a call to action. It's bright coloured and deliberately breaks up text and draws the eye to it. We want it to stand out. The explainer text is there to set expectations before beginning the interaction. There are different flavours of Katacoda embed, the expectation setting for this one is that a terminal window appears at the bottom of your browser window. I don't think this is an altogether expected interaction, hence setting expectations for users.

Katacoda Running

To remove confusion once the tutorial has been enabled, my suggestion is to remove the "enabled state" outline button, and provide instruction on how to dismiss the tutorial once loaded.

bradnoble:
w/r/t the tutorial trigger button question: @cero-miedo have you tried an "on/off" affordance like we see in the dark mode toggle?

josh-heyer:
Superseded by #996

PR #979, opened 2021-02-23
slight copy adjustments on the index

bradnoble:
- Removed italics around mentions of pgBackRest (we don't italicize product names anywhere else)
- switched from "Postgres" to "PostgreSQL"
- put the link to the .org in a Markdown admonition (!!! Note...), per the new standard
- removed language in the intro about pgBackRest being "as easy as could be" as this is for users and not marketing per se

### DIFF
--- a/advocacy_docs/supported-open-source/pgbackrest/index.mdx
+++ b/advocacy_docs/supported-open-source/pgbackrest/index.mdx
@@ -1,5 +1,5 @@
 ---
-title: 'pgBackRest Guide'
+title: 'pgBackRest'
 description: 'Introduction to pgBackRest, a reliable PostgreSQL backup & restore tool'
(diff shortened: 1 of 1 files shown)


epbarger (approved):

pgstef:
@bradnoble
Please don't merge any change before @sebasmannem or me would approve it please.

_Postgres_ was used on purpose. Meaning PostgreSQL + EPAS.
_pgBackRest_ was also intended on purpose. If you change it there, you'll have to change it in all the files. Now we're in an inconsistency state.

So please, revert this or correct it.

epbarger:
My mistake, I jumped the gun on merging this

pgstef:
> My mistake, I jumped the gun on merging this

Don't worry, PR #984 should fix it ;-)

PR #992, opened 2021-02-25
Homepage tweaks for launch

cero-miedo:


src/pages/index.js
```diff
@@ -117,6 +148,17 @@ const Page = () => (
...
               Barman
             </IndexCardLink>
+
+            <IndexCardLink
+              to="/supported-open-source/barman/single-server-streaming/"
+              className="nested-link"
+            >
+              Single Server Streaming
+              <span className="new-thing" title="Interactive Katacoda Tutorial">
+                {' '}
```
epbarger, line 158: What's this accomplishing? Could padding or the `&:before content` replace this? Not a big deal, just curious

src/pages/index.js
```diff
@@ -117,6 +148,17 @@ const Page = () => (
             <IndexCardLink to="/supported-open-source/barman/">
               Barman
             </IndexCardLink>
+
+            <IndexCardLink
+              to="/supported-open-source/barman/single-server-streaming/"
+              className="nested-link"
```
epbarger, line 154: This is fine for now, but I could see it becoming a property of IndexCardLink, like `<IndexCardLink nested={true}` or similar. Especially if we're going to use this style more

epbarger (approved):

src/pages/index.js
cero-miedo: I think the linter added that when I committed :(

PR #997, opened 2021-02-26
Remove the Attention component

epbarger:


josh-heyer (approved):
Donno about all those parens around parameters, but the rest looks good! ;-P

cero-miedo (approved):

PR #1002, opened 2021-02-26
develop ← main after: release/2021-02-26

cero-miedo:


gsong:
@cero-miedo We don't need this PR, since the release branch was already merged into `develop`, which in turn was merged into `main`.

We can talk about the intent of `main`, `develop`, and release branches in a future discussion as a team.

Issue #1009, opened 2021-03-01
Regarding /kubernetes/cloud_native_operator/expose_pg_services.mdx


marclinster:
The link under 'Quickstart' is broken
https://www.enterprisedb.com/docs/quickstart/ gets a 404

PR #1029, opened 2021-03-03
Fixing #1009

cero-miedo:


cero-miedo:
Stale, fixed in George's latest CNO PR

PR #1033, opened 2021-03-03
remove dev sources requirement

epbarger:


cero-miedo (approved):
Looks good to me, would like to get @josh-heyer's take before we merge :)

gatsby-config.js
```diff
@@ -65,12 +66,23 @@ const sourceToPluginConfig = {
...
+      result[source] = true;
+      return result;
+    }, {});
+
+    if (gracefulFs.existsSync(sourceFilename)) {
+      console.log(
+        `${ANSI_BLUE}###### Sourcing from ${sourceFilename} #######${ANSI_STOP}`,
+      );
+      console.log(
+        `${ANSI_GREEN}Note that ${sourceFilename} is no longer strictly required - the full set of docs will be loaded in its absence.${ANSI_STOP}`,
```
josh-heyer, line 81: Maybe be a bit more prescriptive here: the intention is to give folks a heads-up that they might need to take action to get the sources that they need, so might as well be explicit about that action. 
```suggestion
        `${ANSI_GREEN}Note: ${sourceFilename} is no longer required; delete it to load the full set of docs.${ANSI_STOP}`,
```

josh-heyer (approved):
One suggestion inline; otherwise looks fine.

PR #1038, opened 2021-03-04
Fix Katacoda closing on link hover in development

epbarger:
Thanks to @gsong for helping figure out what was going on here!

The components passed into `MDXProvider` were changing somehow on link hover - memoizing the components object seems to fix the issue.

### DIFF
--- a/src/components/code-block.js
+++ b/src/components/code-block.js
@@ -130,7 +130,7 @@ const OutputPre = ({ content }) => (
   </div>
 );
 
-const CodeBlock = ({ children, katacodaPanelData, ...otherProps }) => {
+const CodeBlock = ({ children, codeLanguages, ...otherProps }) => {
   const childIsComponent = !!children.props; // true in normal usage, false if raw <pre> tags are used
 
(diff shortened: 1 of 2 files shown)


josh-heyer (approved):
Looks good - much less of a minefield!

I wonder, do you see of a way by which we could avoid losing the panel when navigating via the right-hand ToC? e.g., /supported-open-source/barman/single-server-streaming/step02-backup-setup/ -> /supported-open-source/barman/single-server-streaming/step02-backup-setup/#configuration

epbarger:
> Looks good - much less of a minefield!
> 
> I wonder, do you see of a way by which we could avoid losing the panel when navigating via the right-hand ToC? e.g., /supported-open-source/barman/single-server-streaming/step02-backup-setup/ -> /supported-open-source/barman/single-server-streaming/step02-backup-setup/#configuration

:+1:, will merge

Didn't realize the TOC links closed it. I'll mess with it a bit - I suspect reducing the url in the links there to just the anchor might fix it.

PR #1047, opened 2021-03-09
Containerize PDF build tools

### LINKED ISSUE #1043: Dockerize local pdf building
# Goal

Replace `yarn build-pdf /pdf/path` with a new command that builds the pdf within a docker contianer. It needs to take in a path to a doc directory, and sync the resulting file back to the host filesystem (2-way volume?).

We want to avoid users needing to setup `pandoc` and `wkhtmltopdf` on their local machine.

# Dependencies

- python 3 >= 3.6
- wkhtmltopdf 0.12.6 (latest version)
- pandoc (latest version)
  
# Example

`yarn build-pdf-docker /product_docs/docs/epas/13/epas_inst_linux/` should spin up the container if needed, run `python3 scripts/pdf/generate_pdf.py /product_docs/docs/epas/13/epas_inst_linux/`, display the output of the script on the screen for debugging, and ultimately create the pdf at `product_docs/docs/epas/13/epas_inst_linux/epas_v13_epas_inst_linux_documentation.pdf`. The pdf should be available on the host filesystem.

gsong:
Delegate PDF generation to a Docker container. There's no change to how you run the commands `yarn build-pdf` and `yarn build-all-pdfs`. 🐢 The first time you run either command in this branch, it'll be slow, as Docker has to pull the base Ubuntu image and build the `edb/x-docs-pdf-builder` image.

## How to Test

* Run `yarn build-pdf` and / or `yarn build-all-pdfs` commands as you did before.
* Run `yarn build-all-pdfs-ci` to simulate what would be done in CI.

## Changes

* Docker specs
* Add a new npm script `build-all-pdfs-ci`, which  is the same as `build-all-pdfs` before. Used in CI only.
* Clean up `generate_pdf.py` a bit:
  * Use `subprocess.run` instead of `os.system` for safety.
  * Various Python best practices.
  * Black format to community style guide.

## Limitations

* The `--open` and `--html` flags won't work properly, since the job is run inside a Docker container.
* Not using the Docker container in CI.

Closes #1043.


docker/docker-compose.build-pdf.yaml
```diff
@@ -0,0 +1,13 @@
+version: '3.8'
+
+services:
+  pdf-builder:
+    build:
+      context: ../
+      dockerfile: docker/images/Dockerfile.pdf-builder
+    image: edb/x-docs-md-to-pdf
```
gsong, line 8: @josh-heyer @cero-miedo Is there an established EDB Docker image namespace and image naming and tagging convention?

README.md
```diff
@@ -40,13 +40,7 @@ We recommend using MacOS to work with the EDB Docs application.
 
 ### Installation of PDF / Doc Conversion Tools (optional)
```
epbarger, line 41: Given this still references PDF conversion, should we add a note about installing Docker too?

package.json
```diff
@@ -15,8 +15,9 @@
     "build": "gatsby build --prefix-paths",
     "serve-build": "gatsby serve --prefix-paths",
     "update-icons": "git submodule update --init --remote && node scripts/createIconTypes.js && node scripts/createIconNames.js",
-    "build-pdf": "python3 scripts/pdf/generate_pdf.py",
-    "build-all-pdfs": "for i in product_docs/docs/**/*/ ; do echo \"$i\"; python3 scripts/pdf/generate_pdf.py ${i%} || exit 1; done",
+    "build-pdf": "docker-compose -f docker/docker-compose.build-pdf.yaml run --rm --entrypoint scripts/pdf/generate_pdf.py pdf-builder",
+    "build-all-pdfs": "for i in product_docs/docs/**/*/ ; do echo \"$i\"; yarn build-pdf ${i%} || exit 1; done",
```
epbarger, line 19: Curious how much overhead there realistically is in calling `yarn build-pdf` in a loop, now that it's dockerized.

package.json
```diff
@@ -15,8 +15,9 @@
     "build": "gatsby build --prefix-paths",
     "serve-build": "gatsby serve --prefix-paths",
     "update-icons": "git submodule update --init --remote && node scripts/createIconTypes.js && node scripts/createIconNames.js",
-    "build-pdf": "python3 scripts/pdf/generate_pdf.py",
-    "build-all-pdfs": "for i in product_docs/docs/**/*/ ; do echo \"$i\"; python3 scripts/pdf/generate_pdf.py ${i%} || exit 1; done",
+    "build-pdf": "docker-compose -f docker/docker-compose.build-pdf.yaml run --rm --entrypoint scripts/pdf/generate_pdf.py pdf-builder",
```
epbarger, line 18: I don't feel too strongly one way or another on this - but there is at least one person on the docs team who is already used to building PDFs with this command. Since we've changed it to be dockerized, they'll have to change their setup to install docker etc. Just something to keep in mind

epbarger (approved):
Tried this out locally, looks good to me! Left some comments but I'm happy with this as is.

README.md
gsong: Good point.

package.json
gsong: I think we should deprecate running pandoc and wkhtmltopdf locally completely, except for power users.

I'm going to add an issue to address to get rid of needing pandoc locally as well.

package.json
gsong: In my very unscientific testing:

* `build-all-pdfs-ci`: 8 minutes
* `build-all-pdfs`: 10 minutes

So I feel it's acceptable.

cero-miedo:
👍

cero-miedo (approved):

package.json
epbarger: Yeah that's fair :+1:

package.json
epbarger: Yeah that seems fine, was just curious

PR #1055, opened 2021-03-11
fix: adjust Pandoc CSS to use specified coding font

gsong:
## What Changed?

- Fixed coding font spec in Pandoc CSS. I assume we want to use the specified font of `Source Code Pro` for both `pre` and `code`, right?


epbarger (approved):
seems right to me

PR #1066, opened 2021-03-15
allow using docker from within docker-hosted dev env

josh-heyer:
Now that @gsong has demonstrated the use of containers to reduce tangential dependencies with #1047, I need my preferred environment to support this. Unfortunately, my preferred environment was already running everything in a container. 

This change adjusts the VSCode container definition to allow the use of Docker commands ("Docker in Docker"). Could have also gone with a lighter-weight [Docker-from-Docker](https://github.com/microsoft/vscode-dev-containers/blob/master/script-library/docs/docker.md) approach, but would've required special handling for VSCode's "Clone Repository in Container Volume" option, which is a handy option but not really worth special-casing; this should behave more or less the same to just running stuff locally.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -28,3 +28,12 @@ RUN if [ "${INSTALL_NODE}" = "true" ]; then su vscode -c "source /usr/local/shar
 RUN su vscode -c "source /usr/local/share/nvm/nvm.sh \
     && npm install -g gatsby-cli \
     && npm install -g yarn" 2>&1
+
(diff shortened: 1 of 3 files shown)


gsong:
Is `.devcontainer/` a VSC specific thing?

josh-heyer:
Yep: https://code.visualstudio.com/docs/remote/containers

gsong:
Is there a reason for editor specific settings to make it in to the repo?

If it's VSC specific, I don't feel like it belongs in the repo. We should probably gitignore `.vscode/` as well.

josh-heyer:
Don't have a strong opinion on .vscode, but .devcontainer allows opening
and building without any other local setup - see:
docs/README_DOCKER_VSCODE.md

gsong (approved):

PR #1094, opened 2021-03-19
Cloud Native PostgreSQL Operator live quickstart

josh-heyer:
## What's New?

A live (in-browser) demonstration of installation, configuration and deployment - essentially the combined contents of the CNP Quickstart, Install and Comdms pages.

This doesn't even get into the good stuff yet, but we gotta start somewhere - and I think a demo that takes literally 5 minutes to work through and leaves you with a 3-node cluster running ain't too bad for whetting the appetite!

View the demo here: https://edb-docs-adv-feature-jo-ukdchu.herokuapp.com/kubernetes/cloud_native_postgresql/interactive/installation_and_deployment/

## What Changed?

This creates a section for interactive content that will be merged into the CNP docs during import. Lacking a more entertaining idea, I created the path `merge_sources/kubernetes/cloud_native_postgresql` and modified the import script to merge the contents of this path with the transformed documentation from the CNP repo (`source_cloud_native_operator.py` - see [sync instructions](https://github.com/EnterpriseDB/docs/blob/develop/docs/how-tos/sync-cnp-docs.md) for details).

I also modified that import script to add the "interactive" directory to the navigation, and a blurb highlighting the existence of a demo as the 2nd paragraph in quickstart.mdx. This is... somewhat fragile, but I don't care to over-engineer this part until / unless it becomes necessary - and right now, this is all that's being merged!

View the modified Quickstart page here: https://edb-docs-adv-feature-jo-ukdchu.herokuapp.com/kubernetes/cloud_native_postgresql/quickstart/ (and note the addition of an "Interactive Demos" section to the side nav)

## Checklist

**Content**


josh-heyer:
@gsong could you take a look at this and suggest a better organization (one that would allow clean merging with future imports from the CNP repo)? 

I approached this draft by moving Quickstart into its own subdirectory, with minor modifications to paths and the addition of a link to the demo; that looks clean enough I think, but is probably a bear to try to automate. Some initial ideas:

- Create separate subdirectory for demo(s) that would be ignored by CNP pulls (cross-linking might get awkward)
- Transfer ownership of Quickstart to this repo (again, would have to work out how linking would work)
- Transformation of Quickstart on import (more complicated import, but could allow CNP to continue to be autonomous)
- ???

gsong:
At first blush, here's my thought:
* Host the additional content in a separate directory of this repo that's ignore by Gatsby.
* Upon import / update of external content source, the import process:
  * Copies the additional content into the content-proper area
  * In-place modifies the upstream files to reference the additional content

The in-place modification could be tricky but do-able, especially if we can inject some kind of comment place-holder upstream. We can choose to inject our custom snippet which links to new content at any stage of the import process.

josh-heyer:
Eh, @gsong - I've mostly taken your suggestion and applied it - seems reasonable, only potential pitfall is editing stuff in the wrong place. 

Have a look?

josh-heyer:
@cero-miedo could you take a look at the blurb at the start of the Quickstart page here: https://edb-docs-adv-feature-jo-ukdchu.herokuapp.com/kubernetes/cloud_native_postgresql/quickstart/ (the admonition titled, "Live demonstration") and let me know your thoughts? Not sure it needs anything TBH, but might tie into the work you've been doing with Katacoda iconography elsewhere.

josh-heyer:
@gbartolini I think you or someone from your team might be interested in looking over the demo I'm adding here, make sure I'm not misrepresenting anything in your work! Direct link: https://edb-docs-adv-feature-jo-ukdchu.herokuapp.com/kubernetes/cloud_native_postgresql/interactive/installation_and_deployment/

cero-miedo:
I've merged in Evan's feature branch which allows us to signal that content is interactive.

## Link from homepage

image

## Header style added for Katacoda enriched content

image

* Button copy has also been renamed. Confusingly, we had called out "Interactive Tutorials" in the button, but referred to them as "Demos" around the site. Converging on Demo.

I've marked this as work in progress as I don't like the behaviour in the menu tree:
image
We can mark the parent as containing Demos, but it looks off when opened:
image

There are no interactive elements in the parent, and it looks clumsy stacking the badges.

josh-heyer:
Yeah, the sidebar thing is... awkward. Also means the text of the top-level item is "Interactive DemosDemos" (lack of space there is also a problem for other titles with this badge).

advocacy_docs/kubernetes/cloud_native_postgresql/interactive/index.mdx
```diff
@@ -0,0 +1,7 @@
+---
+title: 'Cloud Native PostgreSQL Interactive Demonstrations'
+navTitle: 'Interactive Demos'
+product: 'Cloud Native Operator'
+indexCards: full
+showInteractiveBadge: true
```
gsong, line 6: When running the `scripts/source/source_cloud_native_operator.py` script, the `showInteractiveBadge` front matter directive is not generated.

Same thing with `advocacy_docs/kubernetes/cloud_native_postgresql/interactive/installation_and_deployment.mdx`.

scripts/source/source_cloud_native_operator.py
```diff
@@ -37,42 +37,62 @@ def index_frontmatter():
...
-    new_file_path = file_path.replace(".md", ".mdx")
+    new_file_path = file_path.with_suffix(".mdx")
 
     with open(new_file_path, "w") as new_file:
         with open(file_path, "r") as md_file:
             copying = False
-            gh_relative_path = file_path.replace("temp_kubernetes/build/", "src/")
+            quickstart = file_path.name == "quickstart.md"
+            paragraph = 0
+            gh_relative_path = "src/" + str(
```
gsong, line 54: [nit] `Path('src') / file_path...`

scripts/source/source_cloud_native_operator.py
```diff
@@ -37,42 +37,62 @@ def index_frontmatter():
...
+            )
 
             for line in md_file:
-                if copying:
-                    new_file.write(rewrite_yaml_links(line))
-                if line.startswith("#") and not copying:
+                if not line.strip():
+                    paragraph = paragraph + 1
+
+                    if quickstart and paragraph == 2:
```
gsong, line 62: [q] This feels a little fragile. Is it not possible to inject a comment in the original source, then search and replace based on that?

scripts/source/source_cloud_native_operator.py
```diff
@@ -37,42 +37,62 @@ def index_frontmatter():
...
+            paragraph = 0
+            gh_relative_path = "src/" + str(
+                file_path.relative_to("temp_kubernetes/build/")
+            )
 
             for line in md_file:
-                if copying:
-                    new_file.write(rewrite_yaml_links(line))
-                if line.startswith("#") and not copying:
+                if not line.strip():
```
gsong, line 59: [q] We can shortcut based on `quickstart` right here, since it's the only file we care about.

scripts/source/source_cloud_native_operator.py
```diff
@@ -37,42 +37,62 @@ def index_frontmatter():
...
+                        line = """
+!!! Tip "Live demonstration"
+    Don't want to install anything locally just yet? Try a demonstration directly in your browser:
+
+    [Cloud Native PostgreSQL Operator Interactive Quickstart](interactive/installation_and_deployment/)
+
+"""
+                elif copying:
+                    line = rewrite_yaml_links(line)
+                elif line.startswith("#") and not copying:
```
gsong, line 72: [q] We don't need `not copying` condition here, right?

advocacy_docs/kubernetes/cloud_native_postgresql/interactive/index.mdx
josh-heyer: Yep - these should have been modified in the merge_source directory in order for the changes to be preserved.

scripts/source/source_cloud_native_operator.py
josh-heyer: It is fragile; I was aiming to keep this fairly simple initially though.

For the future, I'd like to not have any content *in* the import script itself. Comment + keyword + content in the source material would be responsible for tracking docs-specific stuff, e.g.

```yaml

nav:
 - a
#docs-content - q
 - b
```

```markdown
blah blah blah

<!-- docs-content
!!! Admonition
	that should only appear after import
-->

This would offer a lot more flexibility going forward; may be overkill right now.

gsong: Can you add an issue for improvement, in case we come across another similar use case?

josh-heyer: Good plan! https://github.com/EnterpriseDB/docs/issues/1224

josh-heyer:
Gabriele gives approval for this; just need @cero-miedo to work out the design issues. Suggestions:

- Remove sidenav badges for now; too intrusive. We'll need to sit down and grind out a design there.
- Add specialized look for admonition

Robert's call - he'll own this from now until merge (I'm out next week but ping me if there are any questions)

cero-miedo (approved):

PR #1095, opened 2021-03-19
Visual Enhancements

cero-miedo:
## What Changed?

* orange fill for icon in card decks
* improving legibility of search results, badge counts in quick search and advanced search
* dark mode for both types of search
* added a border to the search input to make it easier to see


epbarger (approved):
Looks good!

PR #1197, opened 2021-04-02
CNP v1.2.0 docs update

gsong:
Just as the title says. 😉


cero-miedo (approved):

josh-heyer (approved):
Changes look good. Realized while reviewing that we'd missed an admonition type: "hint". Addressing this in #1207

PR #1281, opened 2021-04-21
build: add .dockerignore to minimize build context

gsong:
Since we don't rely on any local files for the PDF builder image, we don't need to include them in the build context.


gsong:
@josh-heyer @cero-miedo @frago12 Can one of you approve this? Let me know if you have a question what it's for.

PR #1320, opened 2021-04-28
docs: remove subsection PDFs, which are no longer generated

gsong:
These PDFs cannot be updated automatically using the current PDF
generation script, so no reason to keep them around.

See discussion starting at https://edb.slack.com/archives/CU5QEU5L7/p1619477783152800


josh-heyer (approved):
👍

PR #1236, opened 2021-04-13
EPAS 12 Backfill

robert-stringer:
## What Changed?

This backfills EPAS 12.

## Checklist

Please check all boxes that apply (`[ ]` is unchecked, `[x]` is checked)

**Content**
- [x] This PR adds new content
- [ ] This PR changes existing content
- [ ] This PR removes existing content

PR #1327, opened 2021-05-04
Content/epas/12/backfill redux

gsong:
Replaces #1236.


robert-stringer (approved):

robert-stringer:
@gsong On initial inspection, this looked fine. However, after merging into develop, it created problems when the PDF generation was happening. Our canary in the coal mine, missing images was throwing errors.

After looking closer at the files present inside `product_docs/docs/epas/12`, it looks like *only* the edited files are included.

EPAS 13 was used as the base, edits were then made. It looks like we might be missing the unedited files that make up the rest of the document sets.

gsong:
@robert-stringer Was there an explicit commit which was the copy of EPAS 13 to EPAS 12 that I missed somewhere?

PR #1328, opened 2021-05-04
Updates found while installing on Windows for the first time.

jericson-edb:
## What Changed?

* Without Git installed the clone step fails (silently?).
* I think the "isolated" is now called "unique" now.

## Checklist

**Content**


robert-stringer:
Thanks for the PR, @jericson-edb!

@josh-heyer Do you want to check this out?

josh-heyer (approved):
Looks reasonable - good catch!

PR #1370, opened 2021-05-13
Add PDF link component

frago12:
## What Changed?

Add PDF link component

**Examples**
```jsx
<Archive title="Download PDF" href="link-to-file.pdf" />
```

## Checklist

**Content**

### DIFF
--- a/src/components/archive.js
+++ b/src/components/archive.js
@@ -10,10 +10,10 @@ export default function Archive({ title, href, ...props }) {
   );
 }
 
-const PdfIcon = ({ className }) => (
+const PdfIcon = () => (
   <Icon
     iconName="PDF"
-    className={`fill-orange position-relative top-minus-2 ${className}`}
+    className="fill-orange position-relative top-minus-2"
     width="16"
     height="auto"
   />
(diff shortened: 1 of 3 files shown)


src/components/archive.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
```
gsong, line 1: Is this still necessary? I know you no longer need to do this with CRA.

src/components/archive.js
```diff
@@ -0,0 +1,20 @@
...
+
+export default function Archive({ title, href, ...props }) {
+  return (
+    <Link to={href} title={title} className="small" {...props}>
+      <PdfIcon /> {title}
+    </Link>
+  );
+}
+
+const PdfIcon = ({ className }) => (
```
gsong, line 13: What's the `className` prop about?

src/components/archive.js
frago12: this is Gatsby, so yes, we need the import

src/components/archive.js
frago12: I think it can be removed since no extra styling is needed for this case

frago12: removed

PR #1386, opened 2021-05-18
Fix HTML in titles (hopefully for the last time)

josh-heyer:
## What Changed?

Move anchor links *above* titles; the ToC plugin does not care for HTML *in* titles. Arguably a bug in that plugin, since it really should know better - but... HTML *in* Markdown titles is a bit unusual, and parsing that properly is not a small thing.

Strictly-speaking, we don't need these anchor links at all *for this repo* - the ToC plugin generates its own (via remark-autolink-headers). But they're presumably useful for other targets, they might be useful for cross-linking, and it's nice to be consistent.

There's a fix for this in https://github.com/EnterpriseDB/cloud-native-postgresql/pull/298 - I should've verified the status of that first, since got bit by this for the 1.3.0 release as well.

## Checklist

**Content**

### DIFF
--- a/advocacy_docs/kubernetes/cloud_native_postgresql/api_reference.mdx
+++ b/advocacy_docs/kubernetes/cloud_native_postgresql/api_reference.mdx
@@ -51,7 +51,9 @@ Below you will find a description of the defined resources:
 - [WalBackupConfiguration](#WalBackupConfiguration)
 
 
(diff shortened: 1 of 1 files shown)


jericson-edb (approved):
I noticed that, thought it was weird, planned to ask and . . . got distracted by something else. :-(

Anyway. Looks good to me.

PR #1394, opened 2021-05-19
Updates for DF-229: OCL 13.1.4.2

abhilasha-narendra:
## What Changed?

Updated the Installation topics for DF-229.

## Checklist

**Content**


drothery-edb (commented):
When I looked at the File view I saw that the formatting was a bit off in the Creating a Repository Configuration File section. I think the ">"s need to be removed before the command text. This is in the unchanged content, but we might as well clean it up.

drothery-edb (approved):

Issue #1224, opened 2021-04-08
Process latent Docs-specific content within comments on import of external source material 


josh-heyer:
## The scenario 

Content is being pulled from an external source - for example, the Cloud Native PostgreSQL Operator docs. Upon import, Docs-specific content is being merged into it - for example, live demos. 

## The problem 

Cross-links between external content and Docs-specific content should be exposed upon import, but not rendered in the source repo or other targets (where they will not function). But, ideally these links would be integral to the source material (along with associated supporting content and metadata such as descriptions, YAML, etc) so that their nature and position is not a surprise to authors. 

## The solution

Create functionality to be used by the import scripts that will process specially-constructed comments and surface them in the imported documents / metadata.

### Example:

A source file named `mkdocs.yml` is used to define the navigation in the source material, and migrated into Frontmatter during the import. YAML comments prefixed with `docs-content` are used for navigation items that should only be exposed upon import:

```yaml

nav:
 - "a page in the source"
#docs-content - "a page merged during import"
 - "different page in the source"
```

A source file named `quickstart.md` contains content relevant to the product. HTML comments prefixed with `docs-content` are used to contain additional content which will direct readers to a live demo that will be merged in during import:

```markdown
Original paragraph text blah blah blah

<!-- docs-content
!!! Admonition
	that should only appear after import
-->

blah blah more original text
```

(Example inspired by this PR: https://github.com/EnterpriseDB/docs/pull/1094, which uses a rather more clunky method of merging)

## When should this be implemented?

When there exists more than a single page that could make use of it!

PR #1403, opened 2021-05-20
Fix conversion error on md file with leading whitespace

josh-heyer:
## What Changed?

Conversion script was not handling markdown files where whitespace preceded the first heading. This is not a problem for Markdown, but MDX needs to lead with the frontmatter - no whitespace!

This case is handled now. However, there remains a possibility for a file that has *non-whitespace* preceding the first header; this can be handled with additional effort, but for now I've just added a warning to the script to call this out if it is detected.

A more robust approach would be needed anyway for #1224, so unless this potential problem shows up in the wild a fix can wait until that issue is processed.

## Checklist

**Content**

### DIFF
--- a/advocacy_docs/kubernetes/cloud_native_postgresql/monitoring.mdx
+++ b/advocacy_docs/kubernetes/cloud_native_postgresql/monitoring.mdx
@@ -1,4 +1,3 @@
-
 ---
 title: 'Monitoring Instances'
 originalFilePath: 'src/monitoring.md'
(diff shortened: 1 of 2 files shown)


josh-heyer:
oops, saw another issue here; stand by

josh-heyer:
Ink wasn't even dry on this before I saw another problem: a link that spans two lines and wasn't rewritten. Well, that didn't take long - we've hit the limits of a simple line-based regex converter. For now, I've manually patched the file, but I'll need to rewrite this script to handle markdown more intelligently before we run it again.

jericson-edb (approved):
Seems sane. ;-)

PR #1406, opened 2021-05-21
Removed the note

ManjushaVaidya:
## What Changed?

- Removed the note as currently, we have documented only the installation on Windows.

## Checklist

**Content**


abhilasha-narendra (commented):
Looks good

drothery-edb:
I think the note is still relevant because the topic discusses general platform support and isn't specific to which installers we have documented. I think we should leave it in. Let's discuss.

ManjushaVaidya:
I have removed the note about RHEL, CentOS, and OL platforms as the existing guide has information about Windows installation only.

drothery-edb (approved):

PR #1435, opened 2021-06-01
Minor tweaks to windows guide

sheet20:
## What Changed?

Minor content tweaked in window guide.

## Checklist

**Content**


sheet20:
@abhilasha-narendra, @drothery-edb: Please merge the content with develop branch.

abhilasha-narendra (commented):
Looks good.

drothery-edb (approved):
I don't think it is worth changing, but in the future we should reconsider using a bulleted list when there is only one item in the list. Typically a bullet is not necessary.

PR #1459, opened 2021-06-11
Cleanup disk space before building out.

### LINKED ISSUE #1458: Cleanup disk space before GitHub Actions
Looks like we're [running out of disk space when building](https://github.com/EnterpriseDB/docs/actions/runs/928623441). The short-term solution is to [cleanup unused components](https://github.com/curoky/cleanup-disk-action). In the future, we should see if we can make the repo smaller. (Moving the PDFs to LFS will help.)

jericson-edb:
## What Changed?

* Added a new clean disk job to all GitHub Actions to resolve #1458.

### DIFF
--- a/.github/workflows/deploy-develop.yml
+++ b/.github/workflows/deploy-develop.yml
@@ -7,6 +7,10 @@ jobs:
   build-deploy:
     runs-on: ubuntu-latest
     steps:
+    - name: Cleanup disk
+      uses: curoky/cleanup-disk-action@v2.0
+      with:
+        retain: 'python,node'      
     - uses: actions/checkout@v2
       with:
         ref: develop
--- a/.github/workflows/deploy-main.yml
+++ b/.github/workflows/deploy-main.yml
@@ -7,6 +7,10 @@ jobs:
   build-deploy:
     runs-on: ubuntu-latest
     steps:
+    - name: Cleanup disk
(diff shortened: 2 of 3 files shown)


gsong:
@jericson-edb Were you able to test that this actually works?

josh-heyer (approved):
The only thing I'd watch for here is build times - it's *possible* this will add time to staging builds. If that's not onerous (and, naturally, if this actually works) then 👍

jericson-edb:
I tested it locally using act, but that doesn't seem like the most useful test. Yesterday I realized I can test Actions on my fork. Running [a test](https://github.com/jericson-edb/docs/runs/2850654368?check_suite_focus=true) shows a 23G decrease:

```
-------------Disk info before cleanup----------------
Filesystem      Size  Used Avail Use% Mounted on
/dev/root        84G   45G   40G  54% /
. . .
-------------Disk info after cleanup----------------
Filesystem      Size  Used Avail Use% Mounted on
/dev/root        84G   22G   62G  27% /
```

In this case we weren't close to using the entire disk, but the free disk space varies significantly from one run to the next. Sometimes there isn't quite enough space. For reference, I checked after the test and it used ~10G:

```
Filesystem      Size  Used Avail Use% Mounted on
/dev/root        84G   32G   52G  38% /
```

I think our production releases are bigger, however. At any rate, I have tested it now.

Issue #1290, opened 2021-04-22
Single-file content products is generating blank PDFs


gsong:
The root cause of this issue is because `generate_pdf.py` assumes that any paths with 4 elements in the input path, e.g. `product_docs/docs/bdr/3.6`, is a `fullProductPdf`.

<https://github.com/EnterpriseDB/docs/blob/develop/scripts/pdf/generate_pdf.py#L138-L139> then removes the first file from the file entries to be converted to PDF, in this case the only file in this directory.

<details>
<summary>I believe all these products suffer from this problem</summary>

```
product_docs/docs/bart/2.4
product_docs/docs/bart/2.5.1
product_docs/docs/bart/2.5.2
product_docs/docs/bart/2.5.3
product_docs/docs/bart/2.5.4
product_docs/docs/bart/2.5.5
product_docs/docs/bart/2.5.7
product_docs/docs/bdr/3.6
product_docs/docs/efm/3.10
product_docs/docs/efm/3.6
product_docs/docs/efm/3.7
product_docs/docs/efm/3.8
product_docs/docs/efm/3.9
product_docs/docs/epas/10
product_docs/docs/epas/11
product_docs/docs/epas/12
product_docs/docs/epas/9.5
product_docs/docs/epas/9.6
product_docs/docs/hadoop_data_adapter/2.0.5
product_docs/docs/hadoop_data_adapter/2.0
product_docs/docs/jdbc_connector/42.2.12.1
product_docs/docs/jdbc_connector/42.2.8.1
product_docs/docs/jdbc_connector/42.2.9.1
product_docs/docs/migration_toolkit/53.0.1
product_docs/docs/net_connector/4.0.10.1
product_docs/docs/net_connector/4.0.10.2
product_docs/docs/net_connector/4.0.6.1
product_docs/docs/net_connector/4.1.3.1
product_docs/docs/net_connector/4.1.5.1
product_docs/docs/ocl_connector/12.1.2.1
product_docs/docs/ocl_connector/13.1.4.1
product_docs/docs/odbc_connector/12.0.0.1
product_docs/docs/odbc_connector/12.0.0.2
product_docs/docs/odbc_connector/12.2.0.1
product_docs/docs/pem/7.10
product_docs/docs/pem/7.11
product_docs/docs/pem/7.12
product_docs/docs/pem/7.13
product_docs/docs/pem/7.14
product_docs/docs/pem/7.15
product_docs/docs/pem/7.6
[...]

PR #1487, opened 2021-06-16
Update PDF chapter numbering algorithm and add tests

gsong:
### Tests for File Picking and Chapter Numbering

Add tests for file finding and chapter numbering:

- [x] No content index should not generate PDF
- [x] Root level index with content should generate PDF
- [x] No content index with subdirectories should have proper chapter numbers
- [x] Root index _and_ subdirectories should have proper chapter numbers

### Top-Level Peers to Non-Empty Index Are Chapter Numbered Properly

[Per discussion](https://github.com/EnterpriseDB/docs/issues/1290#issuecomment-861161063), top-level peers to non-empty `index.mdx` should get their own major chapter numbers.

I've staged updated files in [Google Drive](https://drive.google.com/drive/folders/1eLSeYFWlgr6IU6OedAXEFbqRoA8utr5S?usp=sharing).

### Generate Raw Text for PDF Diffs

In some cases, visual PDF diffs are hard to read (such as when there were outright missing chapters). `compare_pdfs.py` now has an optional `--text` flag which can be used to parse out raw text from the PDFs for comparison.

## Notes

If you would like to run the tests:

1. Create a Python 3.8+ virtual environment
2. `pip install -r requirements.txt`
3. `pytest`

For visual diffs of raw PDF text, I personally use the following tools:

* <https://www.perforce.com/products/helix-core-apps/merge-diff-tool-p4merge>
* <https://github.com/dandavison/delta>


gsong:
@josh-heyer @jericson-edb @abhilasha-narendra @robert-stringer Did I miss any other combinations?

gsong:
⚠️ Rebased.

jericson-edb (approved):
Great work! It's good we have the start of a test framework.

josh-heyer (approved):
Added one test case. Looks solid. Works like a charm!

abhilasha-narendra:
This seems to be the perfect solution to our PDF issue. Thanks a lot @gsong

PR #1492, opened 2021-06-17
Path typo from user feedback

jericson-edb:
## What Changed?

- User pointed out that the path was wrong by just a `.`.

## Checklist

**Content**


jericson-edb:
Fixed instead by [this PR](https://github.com/EnterpriseDB/docs/pull/1499).

PR #1495, opened 2021-06-18
Ohter PDF-build-breaking issues

josh-heyer:
## What Changed?

Couple more bad images paths from #1488 also another github paste that wasn't actively causing issues but still shouldn't be there.

## Checklist

**Content**


jericson-edb:
Uff. I thought I got them all, but I clearly should have tested before doing a PR.

PR #1500, opened 2021-06-21
Label-triggered draft deploys

josh-heyer:
Enables creation of draft deployments associated with a PR by adding a "deploy" label. Will then build anew each time the associated branch is synchronized (i.e., every time it is pushed to, provided GitHub is able to merge with the base branch).

I've been testing this for a while on the cloud branch (#1531) and also a couple of long-lived branches for the Docs team. It is... Good enough for now. At very least, it beats manually deploying a branch.

@jericson mind having a look at these to ensure they're compatible with your recent updates to the other deployment workflows?


.github/workflows/deploy-draft.yml
```diff
@@ -0,0 +1,86 @@
...
+    - name: Adjust file watchers limit
+      run: echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
+
+    - uses: actions/setup-node@v1
+      with:
+        node-version: '14.x'
+    - name: Install yarn
+      run: sudo npm -g install yarn
+    - name: Yarn install
+      run: yarn install --immutable
```
frago12, line 41: what do you think about having Yarn installed in the package.json as a dev dependency and then just use it here  `npx yarn install --immutable`?

.github/workflows/deploy-draft.yml
```diff
@@ -0,0 +1,86 @@
...
+        ALGOLIA_API_KEY: ${{ secrets.ALGOLIA_API_KEY }}
+        ALGOLIA_APP_ID: ${{ secrets.ALGOLIA_APP_ID }}
+        ALGOLIA_INDEX_NAME: edb-docs-staging
+        INDEX_ON_BUILD: false
+
+    - name: Netlify deploy
+      run: |
+        sudo yarn global add netlify-cli
+        netlify deploy --dir=public --json | tee deploy_results.json
+        echo "NETLIFY_DRAFT_URL=`jq -r '.deploy_url' deploy_results.json`" >> $GITHUB_ENV
```
frago12, line 73: Same as above, we could add `netlify-cli` as a dev dependency to the package.json or use this other action? https://github.com/marketplace/actions/netlify-actions

frago12 (changes requested):

josh-heyer:
Thanks for the suggestions, @frago12 - with that plus a bit of extra care in cache preservation, these are looking a lot shinier!

.github/workflows/deploy-draft.yml
```diff
@@ -0,0 +1,88 @@
...
+        
+    - uses: actions/checkout@v2
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+        fetch-depth: 0 # fetch whole repo so git-restore-mtime can work
+
+    - name: Adjust file watchers limit
+      run: echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
+
+    - uses: actions/setup-node@v1
```
jericson-edb, line 31: I switched to @v2 here.

jericson-edb (approved):
Looks good! 

I'm thinking we might want to [package some of our action steps](https://docs.github.com/en/actions/creating-actions/creating-a-composite-run-steps-action) to so that we can avoid duplicating code. (But this is scope creep for now!)

Issue #1508, opened 2021-06-28
Migration Portal documentation should not a Version drop-down 


drothery-edb:
## Summary
The Migration Portal has a dropdown listing all the recent releases. Since this is a hosted web app the concept of versions does not apply. Users cannot go back to an earlier version. Prasant, the PO, has requested we just have a label showing the latest version.

## Where did you see the problem?
The Version dropdown
<img width="292" alt="dropdown" src="https://user-images.githubusercontent.com/83650384/123678795-c6baab80-d814-11eb-86d2-35641d19c863.png">

## Expected behavior
We should just display the latest version as a label.

drothery-edb:
@prashantagar, I created this GitHub for the request you made to only show the latest version. The related PR is https://github.com/EnterpriseDB/docs/pull/1510. I requested you and Josh as reviewers.

PR #1510, opened 2021-06-28
Migration Portal 3.2.0 version consolidation

drothery-edb:
## What Changed?

This PR addresses https://github.com/EnterpriseDB/docs/issues/1508. 

It removes the 3.0.1 and 3.1.0 version folders in the repo so that only the latest version is indicated on Docs 2.0. It also includes a comment in the front matter of the index file explaining the versioning strategy.

I also added a pointer to the MP wiki What's new page for users who might want to know what was new in earlier releases/pushes. (Note I couldn't add an actual link because that page on the Migration Portal wiki does not have a unique URL.)

**Content**
- [ x] This PR adds new content


josh-heyer (approved):
I've added a commit to enable redirects from the removed versions to 3.2.0. Otherwise looks good to me.

prashantagar (approved):
Looks good to me.
Thank you Dee Dee and Josh.

PR #1542, opened 2021-07-09
Updates to optional tools section

drothery-edb:
Added more detail about how to create local PDFs and split out the RST to MDX section

## Checklist

**Content**


gsong:
@drothery-edb In the future, please refer to https://github.com/EnterpriseDB/docs/blob/develop/docs/agreements/branch-and-release.md#use-cases for branch naming strategies. 😊

drothery-edb:
@gsong, which branching use case would this type of change be? Add Documentation for Currently Released Product (Or Product Version)? Feature? BugFix? Should I have created an issue to associate it with? Or was my most egregious faux pas that I used underscores instead of slashes?

PR #1544, opened 2021-07-09
style: better MD formatting

gsong:
Suggested changes to #1542.

@drothery-edb I've made these markdown formatting changes:

1. In general, avoid using the trailing double space trick if possible. If you're in a bulleted block, just use blank line and indentation to keep everything as a block.
2. Prefer code fence blocks.


drothery-edb (approved):
@gsong, thank you so much for these comments. Great learning experience for the MD newbie.

PR #1549, opened 2021-07-12
Archive component

frago12:
## What Changed?

Added Archive component to allow the download of deprecated docs

**Examples**
```jsx
<Archive
  title="Installation Guide"
  path="/docs/epas/9.6/EDB_Postgres_Advanced_Server_Installation_Guide_v9.6.pdf"
/>
```


src/components/archive.js
```diff
@@ -0,0 +1,25 @@
...
+  const url = `https://github.com/EnterpriseDB/docs-archive/raw/${environment}${path}`;
+
+  return (
+    <Link to={url} title={title} className="w-100 d-block" {...props}>
+      <PdfIcon /> {title}
+    </Link>
+  );
+}
+
+const PdfIcon = ({ className }) => (
```
gsong, line 18: We don't need any props, right?

src/components/archive.js
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+
+import Icon from "./icon";
+import Link from "./link";
+
+const environment = process.env.GATSBY_ENVIRONMENT_BRANCH;
+
+export default function Archive({ title, path, ...props }) {
+  const url = `https://github.com/EnterpriseDB/docs-archive/raw/${environment}${path}`;
```
gsong, line 9: I would consider adding a leading slash to `path` if the user doesn't supply one.

gsong (approved):

src/components/archive.js
frago12: right!

src/components/archive.js
frago12: 👍

PR #1559, opened 2021-07-14
Added appendix to Thales Guide and sidebar logo to guide template and Thales Guide

sethijs:
This is a reimport of the Partner Docs template and Thales Guide


josh-heyer (approved):
Looks good, just needs Partner_Docs -> partner_docs - we'll do that after merge.

PR #1562, opened 2021-07-15
Download PDFs from Github repo

frago12:
Download PDFs directly from the archived-docs repo

### DIFF
--- a/src/components/archive.js
+++ b/src/components/archive.js
@@ -14,8 +14,8 @@ export default function Archive({
   const updatedPath = !path.startsWith("/") ? `/${path}` : path;
   const url = `https://github.com/EnterpriseDB/docs-archive/raw/${environment}${updatedPath}`;
 
-  return variant === "inverse" ? (
-    <InverseVariant {...{ title, url }} {...props} />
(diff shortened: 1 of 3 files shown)


src/components/archive.js
```diff
@@ -5,22 +5,43 @@ import Link from "./link";
...
+export default function Archive({
+  variant = "default",
+  title,
+  path = "",
+  ...props
+}) {
   const updatedPath = !path.startsWith("/") ? `/${path}` : path;
   const url = `https://github.com/EnterpriseDB/docs-archive/raw/${environment}${updatedPath}`;
 
+  return variant === "inverse" ? (
```
gsong, line 17: Maybe instead of "inverse", "title-first"?

gsong (approved):

josh-heyer (approved):

src/components/archive.js
frago12: I prefer to keep the `variant` prop as is (will update the variant name as suggested), I don't think this component needs to know if the primary link is a PDF or not.

PR #1563, opened 2021-07-16
Fix "undefined" in archive paths

josh-heyer:
## What Changed?

Apparently `GATSBY_ENVIRONMENT_BRANCH` is undefined (at least in local and staging builds). Re-use the APP_ENV approach already used elsewhere to determine which archive branch to link to.

## Checklist

**Content**

### DIFF
--- a/src/components/archive.js
+++ b/src/components/archive.js
@@ -3,7 +3,8 @@ import React from "react";
 import Icon from "./icon";
 import Link from "./link";
 
-const environment = process.env.GATSBY_ENVIRONMENT_BRANCH;
+const isProduction = process.env.APP_ENV === "production";
+const environment = isProduction ? "main" : "develop";
 
 export default function Archive({
   variant = "default",


src/components/archive.js
```diff
@@ -3,7 +3,8 @@ import React from "react";
 import Icon from "./icon";
 import Link from "./link";
 
-const environment = process.env.GATSBY_ENVIRONMENT_BRANCH;
+const isProduction = process.env.APP_ENV === "production";
+const environment = isProduction ? "main" : "develop";
```
frago12, line 7: The idea of `GATSBY_ENVIRONMENT_BRANCH` is to point it to a specific branch if needed, mainly for development purposes, but I guess `APP_ENV` is enough? I'm ok using `APP_ENV`. If we do so we should remove the `netlify.toml` file and update `onPostBuild` hook on `gatsby-node.js` file so that it doesn't copy the netlify file to the static resources

josh-heyer: If we can make it work, great - would be nice to have elsewhere! But right now the links on staging are invalid, e.g. https://edb-docs-staging.netlify.app/docs/epas/10/

frago12: Gotcha, I'm fine with this then, will follow this up next week to see if we can get `GATSBY_ENVIRONMENT_BRANCH` working.

Thanks for checking this.

frago12 (approved):

jibachang:
你们真能整

PR #1592, opened 2021-07-22
Docs/pem reorg

gsong:
Attempted fix of [pem_8_rel_consolidation](https://github.com/EnterpriseDB/docs/tree/pem_8_rel_consolidation) branch.

The changes that we care about:

```
R096    product_docs/docs/pem/7.16/index.mdx    product_docs/docs/pem/7/index.mdx
A       product_docs/docs/pem/7/pem_rel_notes/01_716_rel_notes.mdx
A       product_docs/docs/pem/7/pem_rel_notes/02_715_rel_notes.mdx
A       product_docs/docs/pem/7/pem_rel_notes/03_714_rel_notes.mdx
A       product_docs/docs/pem/7/pem_rel_notes/04_713_rel_notes.mdx
A       product_docs/docs/pem/7/pem_rel_notes/05_712_rel_notes.mdx
A       product_docs/docs/pem/7/pem_rel_notes/index.mdx
R098    product_docs/docs/pem/8.0/index.mdx     product_docs/docs/pem/8/index.mdx
R099    product_docs/docs/pem/8.1.0/pem_ent_feat/04_pem_erd_tool.mdx    product_docs/docs/pem/8/pem_ent_feat/04_pem_erd_tool.mdx
R096    product_docs/docs/pem/8.1.0/pem_ent_feat/17_monitoring_BDR_nodes.mdx    product_docs/docs/pem/8/pem_ent_feat/17_monitoring_BDR_nodes.mdx
R099    product_docs/docs/pem/8.1.0/pem_inst_guide_linux/04_installing_postgres_enterprise_manager/06_configuring_the_pem_server_to_use_kerberos_authentication.mdx     product_docs/docs/pem/8/pem_inst_guide_linux/04_installing_postgres_enterprise_manager/06_configuring_the_pem_server_to_use_kerberos_authentication.mdx
R076    product_docs/docs/pem/8.0/pem_online_help/05_toc_pem_management_basics/11_managing_database_objects/21_synonym_dialog.mdx       product_docs/docs/pem/8/pem_online_help/05_toc_pem_management_basics/11_managing_database_objects/23_synonym_dialog.mdx
A       product_docs/docs/pem/8/pem_rel_notes/08_810_rel_notes.mdx
A       product_docs/docs/pem/8/pem_rel_notes/09_801_rel_notes.mdx
A       product_docs/docs/pem/8/pem_rel_notes/10_800_rel_notes.mdx
A       product_docs/docs/pem/8/pem_rel_notes/index.mdx
```

[...]

PR #1603, opened 2021-07-26
added redirects for 7 and 8

drothery-edb:
Added redirects from minor and patch release versions to major versions - part of version consolidation project 

See #1592


josh-heyer (approved):
Enthusiastic 👍

Issue #1602, opened 2021-07-26
Change the bang notation for factorial to `factorial()`


jericson-edb:
The EPAS documentation sometimes uses an exclamation point to represent factorial. For instance:

> To illustrate the effect of setting the CPU rate limit for resource groups, the following examples use a CPU-intensive calculation of 20000 factorial (multiplication of 20000 * 19999 * 19998, etc.) performed by the query `SELECT 20000!;` run in the psql command line utility.

(https://www.enterprisedb.com/docs/epas/latest/epas_guide/10_edb_resource_manager/)

But that notation is deprecated and will be removed in Postgres v14. Fortunately, we can use `factorial()`, which has been available since PG8.

PR #1605, opened 2021-07-26
Fix for #1602

jericson-edb:
## What Changed?

Fixed some examples of using `!` instead of `factorial()`. I used:

```
find product_docs/ -name \*.mdx | xargs grep '[0-9]!'
```

to find them. Note that two more files (`product_docs/docs/epas/1[23]/epas_compat_spl/02_spl_programs/07_subprogra
ms_subprocedures_and_subfunctions/02_creating_a_subfunction.mdx`) have that pattern, but they are merely printing out the results of `factorial()` in a human-readable way. So those are safe.

## Checklist

**Content**

### DIFF
--- a/product_docs/docs/epas/12/epas_guide/10_edb_resource_manager.mdx
+++ b/product_docs/docs/epas/12/epas_guide/10_edb_resource_manager.mdx
@@ -340,7 +340,7 @@ edb=# SELECT rgrpname, rgrpcpuratelimit FROM edb_resource_group;
 
 Changing the `cpu_rate_limit` of a resource group not only affects new processes that are assigned to the group, but any currently running processes that are members of the group are immediately affected by the change. That is, if the `cpu_rate_limit` is changed from .5 to .3, currently running processes in the group would be throttled downward so that the aggregate group CPU usage would be near 30% instead of 50%.
 
-To illustrate the effect of setting the CPU rate limit for resource groups, the following examples use a CPU-intensive calculation of 20000 factorial (multiplication of 20000 \* 19999 \* 19998, etc.) performed by the query `SELECT 20000!`; run in the `psql` command line utility.
+To illustrate the effect of setting the CPU rate limit for resource groups, the following examples use a CPU-intensive calculation of 20000 factorial (multiplication of 20000 \* 19999 \* 19998, etc.) performed by the query `SELECT factorial(20000)`; run in the `psql` command line utility.
 
 The resource groups with the CPU rate limit settings shown in the previous query are used in these examples.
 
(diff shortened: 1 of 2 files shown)


product_docs/docs/epas/12/epas_guide/10_edb_resource_manager.mdx
```diff
@@ -340,7 +340,7 @@ edb=# SELECT rgrpname, rgrpcpuratelimit FROM edb_resource_group;
 
 Changing the `cpu_rate_limit` of a resource group not only affects new processes that are assigned to the group, but any currently running processes that are members of the group are immediately affected by the change. That is, if the `cpu_rate_limit` is changed from .5 to .3, currently running processes in the group would be throttled downward so that the aggregate group CPU usage would be near 30% instead of 50%.
 
-To illustrate the effect of setting the CPU rate limit for resource groups, the following examples use a CPU-intensive calculation of 20000 factorial (multiplication of 20000 \* 19999 \* 19998, etc.) performed by the query `SELECT 20000!`; run in the `psql` command line utility.
+To illustrate the effect of setting the CPU rate limit for resource groups, the following examples use a CPU-intensive calculation of 20000 factorial (multiplication of 20000 \* 19999 \* 19998, etc.) performed by the query `SELECT factorial(20000)`; run in the `psql` command line utility.
```
jericson-edb, line 343: Looking at it, I think the semicolon should be inside of the code. It's a necessary part of the command. Also I'm not sure what convention we are following for built-in functions. Should they be in all-caps? If so, this might be better written:

> performed by the query `SELECT FACTORIAL(20000);` run in the `psql` command line utility.

jericson-edb (commented):
Don't want to let the perfect be the enemy of the good enough, but I noted a couple of things that might be improved in my changes.

product_docs/docs/epas/12/epas_guide/10_edb_resource_manager.mdx
gsong: +1 to capitalizing reserved keywords in SQL statements, that's been a community convention for a long time now.

josh-heyer: We should be careful to distinguish between keywords and identifiers here. Keywords are case-insensitive; a common convention is to upper-case them in listings. 

Identifiers are case SENSITIVE, but internally PostgreSQL converts them to lower-case when parsing. That's important, because they *behave* as though they were case-insensitive by default, but... They don't *have* to be case-insensitive: if quoted, case will be preserved. So if I create a `"Factorial"` function, and try to reference it via `factorial()` or `FACTORIAL()`, it will fail. Convention here is to lower-case identifiers - including built-in functions, tables, etc. - in listings.

Ref: https://www.postgresql.org/docs/13/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS

josh-heyer: Oh, and definitely move the semicolon into the code span (or omit it entirely) - it makes no sense as a sentence construct, and was almost certainly intended to be a literal quote from the psql command :)

josh-heyer (approved):

drothery-edb:
@jericson, thank you for responding to this feedback. You rock.

PR #1626, opened 2021-07-29
Update 02_high_availibility.mdx

aswright491:
## What Changed?

I removed repetitive content towards the end and edited the first paragraph. CNPO uses primary so I think we should stick with that language.

## Checklist

**Content**

### DIFF
--- a/product_docs/docs/edbcloud/beta/overview/02_high_availibility.mdx
+++ b/product_docs/docs/edbcloud/beta/overview/02_high_availibility.mdx
@@ -2,11 +2,14 @@
 title: "High Availability"
 ---
 
-EDB provides optional high availability aimed to reduce downtime in the event of a zone or instance becoming unavailable. High availability clusters are configured with one _primary_ and two _replicas_ with synchronous streaming replication. Clusters are configured across availability zones in regions with availability zones.
+EDB provides optional high availability to minimize downtime in case of failures. High availability clusters are configured with three database instances - one _primary_ and two _replicas_ - with streaming replication from the primary to the replicas. In cloud regions with availability zones, instances are provisioned across multiple availability zones to provide fault tolerance in the face of datacenter failure.
+
+Incoming client connections are always routed to the current primary. In case of failure of the primary, a standby replica will automatically be promoted to primary and new connections will be routed to the new primary. When the old primary recovers, it will re-join the cluster as a secondary.
+
+ By default, transactions must synchronously commit to one replica before acknowledgement - in PostgreSQL terms, `synchronous_commit` is set to `on` and `synchronous_standby_names` is set to `ANY 1 (replica-1, replica-2)`. Customers can modify this behavior on a per-transaction, per-session, per-user, or per-database basis with appropriate `SET` or `ALTER` commands.
 
 * Replicas are usually called _standby servers_ and can also be used for read-only workloads.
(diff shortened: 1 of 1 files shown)


skrivov-edb (approved):
shorter is better

product_docs/docs/edbcloud/beta/overview/02_high_availibility.mdx
```diff
@@ -2,21 +2,11 @@
...
-EDB Cloud provides built-in high availability aimed to reduce downtime in the event of a zone or instance becoming unavailable.
-An EDB Cloud _cluster_ is a PostgreSQL cluster made up of a single _primary_ and an optional number of _replicas_.
-Streaming replication is used to ensure that the primary and replicas have the same data.
+EDB provides optional high availability aimed to reduce downtime in the event of a zone or instance becoming unavailable. High availability clusters are configured with one _primary_ and two _replicas_ with synchronous streaming replication. Clusters are configured across availability zones in regions with availability zones.
 
-
-* The primary cluster is often called the _leader_.
 * Replicas are usually called _standby servers_ and can also be used for read-only workloads.
-* In case of temporary or permanent unavailability of the primary, a standby server will become the leader.
+* In case of temporary or permanent unavailability of the leader, a standby replica will become the primary.
```
ba-edb, line 8: We want to standardize on "primary", right, not "leader"?

product_docs/docs/edbcloud/beta/overview/02_high_availibility.mdx
```diff
@@ -2,21 +2,11 @@
 title: "High Availability"
 ---
 
-EDB Cloud provides built-in high availability aimed to reduce downtime in the event of a zone or instance becoming unavailable.
-An EDB Cloud _cluster_ is a PostgreSQL cluster made up of a single _primary_ and an optional number of _replicas_.
-Streaming replication is used to ensure that the primary and replicas have the same data.
+EDB provides optional high availability aimed to reduce downtime in the event of a zone or instance becoming unavailable. High availability clusters are configured with one _primary_ and two _replicas_ with synchronous streaming replication. Clusters are configured across availability zones in regions with availability zones.
```
ba-edb, line 5:
```suggestion
EDB provides optional high availability to minimize downtime in case of failures. High availability clusters are configured with three database instances - one _primary_ and two _replicas_ - with streaming replication from the primary to the replicas. In cloud regions with availability zones, instances are provisioned across multiple availability zones to provide fault tolerance in the face of datacenter failure.

Incoming client connections are always routed to the current primary. In case of failure of the primary, a standby replica will automatically be promoted to primary and new connections will be routed to the new primary. When the old primary recovers, it will re-join the cluster as a secondary.

 By default, transactions must synchronously commit to one replica before acknowledgement - in PostgreSQL terms, `synchronous_commit` is set to `on` and `synchronous_standby_names` is set to `ANY 1 (replica-1, replica-2)`. Customers can modify this behavior on a per-transaction, per-session, per-user, or per-database basis with appropriate `SET` or `ALTER` commands.
```

product_docs/docs/edbcloud/beta/overview/02_high_availibility.mdx
ba-edb: I included this verbiage in my suggestion above.

PR #1627, opened 2021-07-29
Reliably detect branch name & use 'develop' for edit links on production

josh-heyer:
## What Changed?

1. Avoid pointing folks to an editor on the `main` branch: this will not have the most recent changes (if changes are pending a production build), and will tend to encourage PRs against main (which should be avoided except in emergency / hotfix situations). 
2. Reliably detect the current branch name, both for local builds and builds made against PRs via Github Actions. Use this for edit links except when on production (as noted above). Don't shell out for this info unless necessary (non-GHA builds).
3. Use the same branch name / head sha for the /docs/build-info page (which I had forgotten existed until just now, but which should probably be consistent here!)

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -43,6 +43,7 @@ const gitData = (() => {
   }
   if (!branch)
     branch = process.env.APP_ENV === "production" ? "main" : "develop";
+  if (!sha) sha = "";
 
   branch = branch
     .trim()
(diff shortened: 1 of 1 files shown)


jericson-edb (approved):
Pointing people to the right place _and_ getting better detection of the current branch seem like wins to me.

PR #1642, opened 2021-08-02
Remove beta logo

frago12:


frago12:
@josh-heyer the interactive demo on the homepage is showing the logo with the beta flag, do we want to redo it?

image

josh-heyer:
We should probably swap that out - it's been up for a while now. Any suggestions on a new topic to feature there, @drothery-edb ?

drothery-edb:
@josh-heyer , are you asking about the banner at the top of every page that says: This is the redesigned EDB Docs site. For older product docs, go to the EDB Docs Archive? Can we remove the banner altogether and just use it with updated text when needed?

josh-heyer:
No, the "featured topic" banner on the homepage - that's where the animation above lives.

josh-heyer:
Merge when comfortable, @frago12 !

PR #1644, opened 2021-08-02
Build/george/upgrade pandoc version

gsong:
Upgrade Pandoc to v12.14.1


jericson-edb (approved):
Seems like a good idea. :-)

PR #1655, opened 2021-08-04
Update 04_responsibility_model.mdx

drothery-edb:
Change as discussed with Kelly

Moved contents of the Shared Responsibility section on the Security topic to the Responsibility Model topic.


kelpoole (changes requested):
I have made several updates that I committed to your branch which changed the third person references of customer to you and your as discussed.  I'm not sure if it should be "EDB and you" or "you and EDB" so requesting that you verify the additional changes.  Also removed a redundant sentence created by moving the paragraph between pages

PR #1657, opened 2021-08-04
Update 03_security_compliance.mdx

drothery-edb:
Change as discussed with Kelly

Removed contents of the Shared Responsibility section on the Security topic that was moved to the Responsibility Model topic in PR #1655

- [x ] This PR removes existing content


kelpoole (approved):

kelpoole:
Removed content that was moved to another section

PR #1662, opened 2021-08-05
Update 03_create_cluster.mdx

drothery-edb:
more tidying and added a link the versioning topic. I'll need to check that link in the next deployment.


kelpoole (changes requested):
Recommended changes:
remove note syntax vs just sentence fo password
change type of Postgres to **Postgres Type**
change Version to be **Version**
should the numbers inside 12 be letters (a, b,c)
public/private networking can we use verbiage more llike high availablilty?

drothery-edb:
Comments are incorporated in PR #1663

PR #1678, opened 2021-08-06
Added headers to last procedure

drothery-edb:
Added functional higher level steps to the Create your Subscription procedure to support linking


kelpoole (approved):
I have changed some referenced from subscription to account after talking with Aislinn earlier.  this also aligns better with the topic heading

PR #1679, opened 2021-08-06
Update 03_create_cluster.mdx

drothery-edb:
Fixed list spacing and added links


drothery-edb:
@josh-heyer, I used the full url's for the links I just added. Can you make them relative for me? Also, can you check that I am formatting the embedded ordered lists correctly? Ideally the nested ones would be lowercase alphas, but I'd settle for roman. They show up ok in the preview but it seems in the deployments they display as regular ordinal numbers.

josh-heyer (approved):

PR #1714, opened 2021-08-11
CNP 1.7.1 import

josh-heyer:
## What Changed?

- Import of the docs for Cloud Native PostgreSQL 1.7.1

## Checklist

**Content**

### DIFF
--- a/advocacy_docs/kubernetes/cloud_native_postgresql/samples/cluster-example-initdb.yaml
+++ b/advocacy_docs/kubernetes/cloud_native_postgresql/samples/cluster-example-initdb.yaml
@@ -9,8 +9,11 @@ spec:
     initdb:
       database: appdb
       owner: appuser
+      postInitSQL:
+        - create table numbers (i integer)
+        - insert into numbers (select generate_series(1,10000))
       options:
       - "-k"
-      - "--locale=en_US"
+      - "--locale=en_ZA.utf8"
   storage:
     size: 1Gi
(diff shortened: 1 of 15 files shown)


advocacy_docs/kubernetes/cloud_native_postgresql/samples/cluster-example-initdb.yaml
```diff
@@ -9,8 +9,11 @@ spec:
     initdb:
       database: appdb
       owner: appuser
+      postInitSQL:
+        - create table numbers (i integer)
+        - insert into numbers (select generate_series(1,10000))
       options:
       - "-k"
-      - "--locale=en_US"
+      - "--locale=en_ZA.utf8"
```
jericson-edb, line 17: This seems odd. For reference, en_ZA is [English South Africa](https://metacpan.org/dist/DateTime-Locale/view/lib/DateTime/Locale/en_ZA.pod). I doubt it's a problem, but I'm curious why it was changed to that.

jericson-edb (approved):
Not a problem, but I am curious about the South Africa locale change.

advocacy_docs/kubernetes/cloud_native_postgresql/samples/cluster-example-initdb.yaml
josh-heyer: Huh... That originated with this PR, might wanna drop a question in: https://github.com/EnterpriseDB/cloud-native-postgresql/pull/593

PR #1721, opened 2021-08-11
Hide product version and show banner for EDB Cloud pages

frago12:
* Added 2 new keys to the frontmatter of the MDX pages: `hideVersion` and `showBanner`.
* `hideVersion` is expected to be `true` or `false`.
* `showBanner` accepts a `string` ('edbcloud' for this case). The idea is that in the future, we can show banners with different content if needed.
* The 2 new keys are available in the `navTree` so that we only need to define their value on the top-level page.

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -126,6 +126,7 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
               codelanguages
             }
             hideVersion
+            displayBanner
             directoryDefaults {
               description
               prevNext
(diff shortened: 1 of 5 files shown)


josh-heyer (changes requested):
Looks pretty good! You're gonna want to add a `hideVersion` check in doc.js line 280-282 as well, to remove the version indicator next to the title.

drothery-edb:
@josh-heyer once this gets merged in, we can merge edbcloud-beta after first posting our intention in the upm-docs channel.

josh-heyer (approved):
246 as well, but... Less important now than later

PR #1733, opened 2021-08-12
Update 10_edb_resource_manager.mdx

lucianobotti:
We might need to clarify some part of the docs on `cpu_rate_limit`. Docs said:
- "the value assigned to cpu_rate_limit should typically be less than or equal to 1."
- and also "The valid range of the cpu_rate_limit parameter is 0 to 1.67772e+07"

The value could be more than 1.0 on a multicore system. So for example 1.0 would mean to used 100% of one CPU core, and 2.0 would mean to use 100% of 2 CPUs. That understanding was confirmed on the `eng-db-server` slack channel (https://edb.slack.com/archives/CDTA6PCJV/p1628781278238200) 

Please feel free to change the wording if needed.

## What Changed?

Please list, at a high level, what changed in your branch. If you changed content, include the product, guide, and version as applicable.

**Examples**

- Added more details about `cpu_rate_limit` as it was not clear enough the meaning of values above 1.0.

## Checklist

**Content**


drothery-edb:
Moved change to https://github.com/EnterpriseDB/docs/pull/1745

PR #1745, opened 2021-08-13
Incorporated feedback from PR#1733

drothery-edb:
Added this sentence to EDB Resource Manager topic in epas guide v11, 12, 13

On multicore systems, you can apply the `cpu_rate_limit` to more than one CPU core by setting it to greater than 1. For example, if `cpu_rate_limit` is set to 2.0, you use 100% of two CPUs.


drothery-edb:
fyi @sheet20

lucianobotti (approved):
Agree on that, is the same meaning but with a better wording! Thanks for including it!

drothery-edb:
@josh-heyer , I added you as a reviewer because even though Lucian approved it, it said merging is blocked.

josh-heyer (approved):

josh-heyer:
Added @lucianobotti to the ACL

PR #1747, opened 2021-08-13
Bugfix/jon/1673 GitHub actions lfs

jericson-edb:
## What Changed?

* Added `lfs: true` option to the checkout step on each GitHub Action. (See https://github.com/EnterpriseDB/docs/runs/3325337804?check_suite_focus=true)


josh-heyer (approved):
Oh... Yeah, probably need that 😬

PR #1753, opened 2021-08-16
Minor style edits to Raising Azure Resource Limits

drothery-edb:
Remade changes to links
spelled out "8"
Removed stray quotation mark

- [x ] This PR changes existing content


aswright491 (approved):
@drothery-edb these look good i'd like to get the changes out sooner rather than later, I don't think we also need Guang to review

PR #1767, opened 2021-08-18
Give new frontmatter options a type so that they can be completely optional

josh-heyer:
## What Changed?

This builds on #1721 by making the new frontmatter options optional: with types defined, Gatsby won't try to infer types and will therefore be content if the options are not used anywhere in the codebase. 

My immediate need for this involves being weary of Heroku builds failing, but I do foresee a time when the banner won't be needed anywhere and thus removing it entirely might be useful.

## Checklist

**Content**


frago12 (approved):

frago12:
@josh-heyer good point, thanks for taking care of this

PR #1774, opened 2021-08-19
Remove Heroku support

josh-heyer:
## What Changed?

This stopped working with the move to LFS anyway, and has become more confusing than helpful for product docs authors. Netlify has worked better for some time now; might as well drop this.

I've already disabled the pipeline integration on the Heroku side; once this is merged, I'll remove the webhooks and thousand or so environments that Heroku left littering this repo.

## Checklist

**Content**


frago12 (approved):

frago12:
agree with @jericson, Netlify has been working fine, so I think we'll be Ok with it

PR #1776, opened 2021-08-20
rename directories to drop mdx extension

josh-heyer:
## What Changed?

EPAS 12 and 11 docs had a directory named 03_installing_edb_plus.mdx - this is not a great naming scheme, particularly as EPAS 13 just calls this 03_installing_edb_plus. Let's go with that name. 

## Checklist

**Content**


sheet20:
Looks good.

Issue #1538, opened 2021-07-08
Tracking: tasks to complete before retiring Docs 1.0 


josh-heyer:
- [] OpenAPI HTML gen
- [] Enable rel:canonical header (ref: #163 - this was disabled at some point during release)
- [] Merge archive component (ref: #1368)
- [] Update extant links to docs 1.0 within the docs themselves
- [] Test Netlify redirects
- [] Update legacy redirects
- [] Send legacy redirects to IT
- [] Remove beta label on logo
- [] Remove beta banner ("This is the redesigned EDB Docs site. For older product docs, go to theEDB Docs Archive")

frago12:
@josh-heyer The archive component was merged into `develop` as part of #1549

PR #1785, opened 2021-08-23
Legacy redirect updates

josh-heyer:
## What Changed?

Background: https://github.com/EnterpriseDB/docs/blob/develop/docs/overviews/legacy-redirects/README.md

This is a massive update because we haven't run this since March. Fortunately, we're nearing a point where this should also be close to the last update for this metadata. Changes in this PR:

- Updates to legacy redirect script & metadata
- Addition of version mapping metadata
- Updates to generated legacy redirects
- Updates to manual legacy redirects
  (PostgreSQL installers, Language Pack installers)

Of particular note:

1. This update adds mappings from the so-called "old style" docs in 1.0 (those that were imported from an even older documentation system) to 2.0 docs (in particular, EPAS 11 and 12). I mapped these based on heuristics (primarily title matching), and the mappings are as a result somewhat incomplete; I believe a majority of pages will be mapped, but some number of extant URLs will be missing from redirects.
2. This update maps ALL versions of the PostgreSQL installers on 1.0 to the (unversioned) documentation under Docs 2.0. Some of these were old-style docs on 1.0 (so see note 1 above). Same for the Language Pack docs.

As this touches much of the product documentation tree, it will likely require some careful merging of other pending PRs. When doing so, please defer to the LegacyRedirects and LegacyRedirectsGenerated frontmatter from this PR; if unable to do so (lots of files moved around for instance), let me know so that I can generate the necessary redirects.

Related to: #1538

## Checklist

**Content**


jericson-edb (approved):
I did a few spot checks and it all seems reasonable. I suppose the consequence of a mistake is that it will confuse Google and the unlucky people who have old links.

PR #1826, opened 2021-09-08
Update index.mdx

EFM-Bobby:
## What Changed?

Partly am just testing editing, but it's more correct to say that efm doesn't validate the addresses rather than "verify" the content. EFM will make sure the content is of the right format, for instance.

## Checklist

**Content**


drothery-edb (approved):
Looks good. Go ahead and merge your changes (or I can if you prefer).

EFM-Bobby:
@drothery-edb I don't have write access to make the merge (which is probably a Good Thing for devs in general).

PR #1852, opened 2021-09-15
Clean up Netlify redirects, write guidance for usage

josh-heyer:
## What Changed?

This started as a quick fix for a few files that were consolidated in the EPAS 13 docs, and turned into a tweak to Frontmatter redirect functionality, a review of existing Netlify redirects, and an expansion of the guidance available for using redirects.

When you get a chance,

- @jericson would you mind reviewing the changes / additions to README.md, docs/how-tos/avoid-breaking-links.md, and _redirects for accuracy and narrative flow?
- @drothery-edb and @nidhibhammar could you review the same for applicability to your work and unanswered questions?
- @frago12 would you mind doing a quick sanity-check of my changes to gatsby-node.js and gatsby-util.js?

### DIFF
--- a/src/constants/gatsby-utils.js
+++ b/src/constants/gatsby-utils.js
@@ -1,4 +1,5 @@
 const fs = require("fs");
+const path = require("path");
 
 const sortVersionArray = (versions) => {
   return versions.sort((a, b) =>
@@ -251,9 +252,17 @@ const findPrevNextNavNodes = (navTree, currNode) => {
   return prevNext;
 };
 
-const configureRedirects = (toPath, redirects, actions, config = {}) => {
+const configureRedirects = (
+  toPath,
+  toIsLatest,
+  redirects,
+  actions,
+  config = {},
+) => {
   if (!redirects) return;
   redirects.forEach((fromPath) => {
+    // allow relative paths in redirects
+    fromPath = path.resolve("/", toPath, fromPath).replace(/\/*$/, "/");
     actions.createRedirect(
       Object.assign(
         {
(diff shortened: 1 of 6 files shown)


src/constants/gatsby-utils.js
```diff
@@ -265,6 +274,20 @@ const configureRedirects = (toPath, redirects, actions, config = {}) => {
...
+      actions.createRedirect(
+        Object.assign(
+          {
+            fromPath: replacePathVersion(fromPath),
+            toPath: replacePathVersion(toPath),
+            redirectInBrowser: true,
+            isPermanent: false,
+            force: true,
+          },
+          config,
```
frago12, line 287: not a required change, maybe this could be done using ES6 syntax?

```javascript
actions.createRedirect({
              ...config,
              fromPath: replacePathVersion(fromPath),
              toPath: replacePathVersion(toPath),
              redirectInBrowser: true,
              isPermanent: false,
              force: true,
})
```

frago12 (approved):

src/constants/gatsby-utils.js
josh-heyer: That reverses the behavior! Right now, "config" can be used to *override* whatever properties are set in this function - but this could still be done using spread syntax, e.g. 

```javascript
{
  fromPath: replacePathVersion(fromPath),
  toPath: replacePathVersion(toPath),
  redirectInBrowser: true,
  isPermanent: false,
  force: true,
  ...config,
}
```

...of course, this functionality is never used so far as I can tell - this method is only called once, and no config arg is provided. I have... No idea what the intent was TBH. I might be better off dropping it entirely; would simplify the code.

drothery-edb (approved):
Looks good. We didn't have any comments on first blush, but may have some later.

nidhibhammar (approved):

src/constants/gatsby-utils.js
frago12: 👌

PR #1904, opened 2021-10-05
updating doc-release-tag-format table

drothery-edb:
Adding column for PMs, POs, and any other interested parties who want to be informed about updates to their product docs.
The plan is for PR owners to include the interested parties as reviewers, as appropriate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -383,3 +383,7 @@ Option 2: on GitHub
 
 1. Edit a file on GitHub.
 2. Submit changes as a PR on a new branch.
+
+### Style Guide for EDB contributors
+
+See [EDB documentation style guide](https://enterprisedb.atlassian.net/wiki/spaces/DCBC/pages/2387870239/Documentation+Style+Guide).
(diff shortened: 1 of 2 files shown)


drothery-edb:
@kelpoole , can you check that I added the correct PMs and POs? Thanks.

drothery-edb:
@jericson , this PR has been lingering for a while. Does this address the feedback you received?

jericson-edb (approved):
Looks good to me!

drothery-edb:
@jericson-edb , do you mind reviewing again? Do you think it is appropriate to add a link to the style guide from the readme?

jericson-edb:
>  Do you think it is appropriate to add a link to the style guide from the readme?

It would be better if we had a public style guide. For the moment we don't have a lot of outside contributors, but it would be annoying to people who want to help , but can't be sure they are submitting changes we will approve. Probably better to have the link than not, however.

PR #1947, opened 2021-10-18
fixing redirect

drothery-edb:
Changed redirect to fix broken link for EDB Cloud portal


jericson-edb (approved):
Tested locally and this does seem to work: http://localhost:8000/edbcloud/latest/using_cluster/03_modify_and_scale_cluster/

(Helpful instructions for redirects if you haven't seen them already: https://github.com/EnterpriseDB/docs/blob/develop/docs/how-tos/avoid-breaking-links.md)

PR #1987, opened 2021-11-02
Hotfix/jon/add to build

jericson-edb:
## What Changed?

Missed one place the name changed and changed one too many times.


josh-heyer (approved):
👍

PR #1999, opened 2021-11-03
Fix broken internal link in BigAnimal docs

ringerc:
## What Changed?

Fixup for broken link in commit aa1117b0e

## Checklist

**Content**


ringerc:
@drothery-edb I'm not sure if there's a better, more format-agnostic way to do internal links, but hope this is right.

Is it worth adding a basic style guide to the top level README, with hints about things like heading capitalisation, US-english spelling, internal links, etc?

drothery-edb (approved):
Thanks for fixing the link.

drothery-edb:
@ringerc , good suggestion about linking to the style guide from the readme. I'll do that.

drothery-edb (changes requested):
@ringerc , I just noticed that this PR has the change being merged into main. Please merge into develop instead.

josh-heyer (approved):

PR #2002, opened 2021-11-03
Move Migration Portal redirects out of individual files and into _redirects

josh-heyer:
## What Changed?

I don't think I've seen redirects updated correctly more than once here; the idea was that they'd be updated *while the new version was being prepped* but since it is consistently an after-thought, let's replace them with a wildcard in the Netlify file static/_redirects.

## Checklist

**Content**


josh-heyer:
@drothery-edb @Rasika03 this is just informational: next MP release should follow this pattern and add a line to the section in _redirects. I'll merge this once I've smoke-tested.

PR #2017, opened 2021-11-08
fix KB link and BigAnimal CLI typo

drothery-edb:
## What Changed?

Fixed broken link to the KB from the nav panel
Fixed incorrect command name in the BigAnimal CLI topic (animal > biganimal)


josh-heyer (approved):
Added one change to hit the *other* place the KB URL should be updated. Should be good to go!

PR #2050, opened 2021-11-18
Updated guides to feature finalized guide template and got rid of numbered headers and prerequisites in Thales Guide

sethijs:


drothery-edb (approved):
LGTM

PR #2058, opened 2021-11-23
Feedback from Support re support topic and KB link

drothery-edb:
## What Changed?

Changed support topic title and KB link label

### DIFF
--- a/src/components/index-sub-nav.js
+++ b/src/components/index-sub-nav.js
@@ -14,7 +14,7 @@ const IndexSubNav = () => (
     <ul className="list-inline mr-3">
       <IndexSubLink url="https://www.enterprisedb.com/">EDB Home</IndexSubLink>
       <IndexSubLink url="https://support.enterprisedb.com/support/s/">
-        Knowledge Base/Technical Alerts
+        Knowledge Base / Technical Alerts
       </IndexSubLink>
       <IndexSubLink url="https://www.enterprisedb.com/contact">
         Contact Us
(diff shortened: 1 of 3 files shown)


josh-heyer:
I've changed the other place this string appears as well, @drothery-edb 

But... I have a bone to pick with the use of `/` here - its meaning is that of the inclusive "and" (knowledge base AND technical alerts) but that meaning is somewhat ambiguous (could also mean "or") and we don't seem to be using this anywhere else in the navigation; I would recommend either just SAYING "and" or using an ampersand here.

josh-heyer (approved):

src/components/index-sub-nav.js
```diff
@@ -14,7 +14,7 @@ const IndexSubNav = () => (
     <ul className="list-inline mr-3">
       <IndexSubLink url="https://www.enterprisedb.com/">EDB Home</IndexSubLink>
       <IndexSubLink url="https://support.enterprisedb.com/support/s/">
-        Knowledge Base
+        Knowledge Base / Technical Alerts
```
drothery-edb, line 17:
```suggestion
        Knowledge Base and Technical Alerts
```
```suggestion
        Knowledge Base / Technical Alerts
```

src/components/side-navigation.js
```diff
@@ -23,7 +23,7 @@ const SideNavigationFooter = () => (
   <ul className="list-unstyled mt-0">
     <hr />
     <FooterItem url="https://support.enterprisedb.com/support/s/">
-      Knowledge Base
+      Knowledge Base / Technical Alerts
```
drothery-edb, line 26:
```suggestion
      Knowledge Base and Technical Alerts
```
```suggestion
      Knowledge Base / Technical Alerts
```

josh-heyer (approved):
👍👍👍

PR #2063, opened 2021-11-23
Moved Guide Template to docs repo, Updated format of existing guides

sethijs:


josh-heyer (approved):
Fixed a couple of issues with formatting and non-LFS images, but looks good otherwise!

PR #2064, opened 2021-11-23
virtual.ip.interface not a unique propertie

EFM-Bobby:
The virtual.ip.interface property shouldn't be in the list of props that have to be unique if using more than one agent on a node. Something similar used to be a requirement, which is probably how this got here. Will take a look at the v3 docs also.

## Checklist

**Content**


dbgreen (approved):

EFM-Bobby:
Wow, how do I change PR title? "propertie" left over from shortening a longer title.

drothery-edb (approved):
LGTM

PR #2065, opened 2021-11-23
virtual.ip.interface not a unique property

EFM-Bobby:
This can be removed from the list of unique properties. In older versions of EFM a sub-interface was required and this had to be unique. We got rid of the subinterface but not the documentation that went with it. Have filed a PR already for the v4 version of this page.

**Content**


drothery-edb (approved):
LGTM

PR #2066, opened 2021-11-24
feature: UX Improvements to header

robert-stringer:
## What Changed?

following feedback from usability testing, we're implementing a method for users to return to the EDB marketing page by clicking the EDB logo. The adjacent `/docs` will return users to the docs homepage.

### Homepage
image

### Content Page & Product Search
image


drothery-edb (approved):
Nice!

PR #2102, opened 2021-12-06
change references to epas 13 to epas 14

drothery-edb:
## What Changed?

Updated EPAS versions in examples in the supported Postgres versions table


nataliawojcik27 (approved):
lgtm

PR #2119, opened 2021-12-13
fixed link to script

drothery-edb:
## What Changed?

Fixed link to script (removed .sh)


drothery-edb:
Confirmed that the drop of .sh in the script name was intentional

josh-heyer (approved):
👍

PR #2129, opened 2021-12-15
Replacing Azure Service Provider with Azure resource provider as the …

moiznalwalla:
Replacing service provider term with resource provider as mentioned in https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-providers-and-types.


drothery-edb (approved):
LGTM

PR #2138, opened 2021-12-15
Use netlify deploy action instead of netlify-cli

josh-heyer:
## What Changed?

Second try (follows #2137): replace netlify-cli with deploy action, since netlify-cli is now suddenly failing to install on the runner (see: [1](https://github.com/EnterpriseDB/docs/runs/4537965585?check_suite_focus=true), [2](https://github.com/EnterpriseDB/docs/runs/4540134421?check_suite_focus=true))

Based on the workflow we've used for draft deploys for a while, but removes PR-specific values (and sets production flag for netlify).

### DIFF
--- a/.github/workflows/deploy-develop.yml
+++ b/.github/workflows/deploy-develop.yml
@@ -6,7 +6,7 @@ on:
   workflow_dispatch:
 
 concurrency:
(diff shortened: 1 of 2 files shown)


gsong:
Replaced by #2140.

PR #2148, opened 2021-12-16
Release: 2021-12-16

drothery-edb:
## What Changed?

- BDR 3.7.14 release
- BigAnimal: cross region restore updates
- Failover Manager: Corrected a property name 
- EPAS: added v12 rel notes


jericson-edb (approved):
👍

PR #2169, opened 2021-12-23
Final update to legacy redirects

josh-heyer:
## What Changed?

Docs 1.0 is still up, but not for long (ref: IT-34110, https://github.com/EnterpriseDB/enterprisedb.com/issues/789) - since it did stay up through the release of EPAS 10, we might as well add redirects for that product. And since EPAS 9.6 and 9.5 (STILL!) constitute a plurality (47%) of all traffic to the old docs site, we might as well try to capture that as well; after discussing this with @drothery-edb, we decided to just send it to the closest matching page on epas/latest.

...Also, I fixed a bunch of mappings for EDB*Plus and captured some legacy cruft pages like genindex that may or may not help.

This should be the last time I ever have to touch these, apart from removing them entirely at some point in the future. 🤞

## Checklist

**Content**


drothery-edb (approved):
LGTM

PR #2174, opened 2021-12-29
EPAS version 10 release notes 

dwicinas:
Point release information for EPAS version 11


dwicinas:
@drothery-edb, you might want to pay particular attention to the base release note for v10. 

Otherwise I think this is about ready to go.

drothery-edb (approved):

PR #2179, opened 2022-01-03
EFM restructuring

drothery-edb:
## What Changed?

Flattening structure and removing "guides" paradigm


drothery-edb:
@EFM-Bobby: here is a prototype of a new structure for the EFM docs. Let me know what you think.
https://61d343f8eb07cb7c0909641a--edb-docs-staging.netlify.app/docs/efm/latest/

EFM-Bobby:
Well, it's confusing to me since I'm used to the other format. I've never seen the "Choosing a deployment architecture" text before since it wasn't part of the user's guide. The information there is incomplete, but I see that the same info existed already in the other document. I think trying to sort all that out now when we're trying to EOL EFM isn't worth the time. These changes look ok to me; will just take me a while to be able to find things again.

josh-heyer (approved):
I like the new layout a lot; much more breadth is accessible right off the bat. The left nav is actually useful, and isn't duplicated in the landing page.

I've fixed a decent pile of broken links - see b5d212b279fcd8836f2d753d880ce8319aef57df 
Something to keep in mind when doing this sort of work: relative links will often need to be updated within files that are moved as well as in those that reference them. While the latter can (and should be!) addressed to some extent with redirects, the former must just be changed.

PR #2188, opened 2022-01-05
Release: 2022-01-05

drothery-edb:
## What Changed?

- Removed duplicated platform support info from EPAS doc.
- Fixed broken link in MP landing page.
- EFM v4.4 release doc updates.


jericson-edb (dismissed):
Looks good. I do wonder if it would make sense to use point release numbers on docs to avoid having to bump from 4.3 to 4.4 when there's a new release. Templates could help here.

josh-heyer (approved):

PR #2199, opened 2022-01-07
MySQL FDW updates for RHEL 8 Power

drothery-edb:
## What Changed?

Updated platform support
Restructured install topic and added RHEL 8 Power section


drothery-edb:
Made the CentOS 8 > Rocky Linux 8 changes in the MKT branch by mistake.

Issue #2206, opened 2022-01-09
Missing ticket number in EPRS 6.2.16 release notes


tureba:
EPRS version 6.2.16 was released with the fix to [XDB-976](https://enterprisedb.atlassian.net/browse/XDB-976).

That fix was prompted by support ticket 75608. But due to a missing field in Jira, the release notes didn't contain the ticket number when EPRS 6.2.16 got released.

If possible, please add the ticket number to the entry, which will help us get the ok from the customer to close the support ticket.

I believe the entry is ""Fixed an issue where an INSERT conflict in MMR Publication is subject to failure of Synchronize operation for a target SMR Subscription in a hybrid cluster.", but the XDB team might be able to double check through Jira now that we added the support ticket number there.

Regards, Arthur

drothery-edb:
@zahid75 , can you confirm that this description in the EPRS 6.2.16 release notes corresponds to support ticket 75608? Thanks.

zahid75:
That is correct, the release notes entry "Fixed an issue where an INSERT conflict in MMR Publication is subject to failure of Synchronize operation for a target SMR Subscription in a hybrid cluster." corresponds to ticket # 75608.

drothery-edb:
Added support ticket in https://github.com/EnterpriseDB/docs/pull/2215

See 
![image](https://user-images.githubusercontent.com/83650384/148932977-7a265aa0-69c5-422d-babf-90bb25cffa7e.png)

Let me know if this addresses the issue.

tureba:
Looks great! That's just what I needed. Thanks!

drothery-edb:
Update is now live

PR #2215, opened 2022-01-11
Replication Server 6.2.16: Missing rel note

drothery-edb:
## What Changed?

Added missing support ticket ID. Addresses https://github.com/EnterpriseDB/docs/issues/2206.


drothery-edb:
Received approval from Arthur on #2206

PR #2216, opened 2022-01-11
BigAnimal: adding regions

drothery-edb:
## What Changed?

Added supported regions


dwicinas:
@drothery-edb, I created a topic in Overview for regions and moved the list of regions out of the "Creating a cluster" section. You might want to take a look to see if that's what we need.

It looks like this PR already needs a review so I'm not doing anything further with the PR for now.

josh-heyer (dismissed):

PR #2219, opened 2022-01-11
Fix some broken & obsolete links in the EFM 4 docs

josh-heyer:
## What Changed?

- Relative links that no longer referenced sibling pages
- Postgresql.org links that referenced version 10 vs. current

## Checklist

**Content**


drothery-edb (approved):
Thanks for fixing all those links. It would be great if the doc team could get trained on using your recommended link checker. Or if you could automate link checking. In the meantime I really appreciate you checking the links for us.

PR #2229, opened 2022-01-15
BigAnimal: UPM-4047 restarts

drothery-edb:
## What Changed?

Added warning on db config page about certain parameter updates causing restarts


nataliawojcik27 (approved):
lgtm

PR #2234, opened 2022-01-18
BigAnimal: match docs to edb pre-flight script

dwicinas:
Made a number of changes to more closely align what appears in the docs with the verbiage in the script. Also tried to clarify that the command provided in the docs lists the possible options and arguments, which wasn't clear to me.


product_docs/docs/biganimal/release/getting_started/01_preparing_azure/index.mdx
```diff
@@ -25,29 +25,26 @@ EDB recommends using the `biganimal-preflight-azure` script to check whether all
...
 
     ```sh
-     curl -sL https://raw.githubusercontent.com/EnterpriseDB/cloud-utilities/main/azure/biganimal-preflight-azure | bash -s \
-      <subscription-ID> <azure-region> \
-      --instance-type <instance-type> \
-      --high-availability \
-      --endpoint <endpoint-type> \
-      --activate-region \
-      --onboard
+    curl -sL https://raw.githubusercontent.com/EnterpriseDB/cloud-utilities/main/azure/biganimal-preflight-azure | bash -s \
```
josh-heyer, line 31: there's no 2nd line, so don't need the continuation character (`\`) at the end

drothery-edb:
@ftouserkani-edb , we made a few more tweaks to this topic to better align the text in the doc to the command help. What do you think of these changes? Thanks for your efforts to improve this topic and bringing the issues to our attention.

drothery-edb:
@xugy99 I made you an optional reviewer on this additional changes to the Preparing your Azure account topic.

xugy99 (approved):
LGTM

ftouserkani-edb:
I still do not see 100% consistencies between the script output and what's included in the description of "Options".

Can we make the description of the options reflect what the script is putting out OR simply put the following screenshot in place instead of describing the "Options" individually in text?


Screen Shot 2022-01-21 at 11 23 36 AM

Screen Shot 2022-01-21 at 11 24 31 AM

drothery-edb:
@xugy99, @ftouserkani-edb, do we need the full curl command we are currently providing in the doc if users are running the script from the Azure command shell? Do we need to add a step to tell them to copy the script from the repo so the command will be found?

xugy99:
Hello @drothery-edb 

that `curl` command is to simplify the user experience of "download and copy the script", I think an experience script user supposes to understand what does that mean?

drothery-edb:
@ftouserkani-edb , made a few more changes for better alignment. How does it look now?

ftouserkani-edb:
Changes are aligned with the script output now. LGTM.

xugy99:
I updated the script, so there is some output that makes differently than before: https://github.com/EnterpriseDB/docs/pull/2254

drothery-edb (approved):

PR #2253, opened 2022-01-25
Use correct relative path format for EFM redirects

josh-heyer:
## What Changed?

This is almost certainly my fault. I used path.resolve() instead of URL path merging for these, which... Leads to situations where relative paths behave differently from what one would normally expect on the web.

Use ../ to walk up where necessary for EFM redirects. @drothery-edb I suspect you had most of these correct the first time... 

## Checklist

**Content**


josh-heyer:
Making this a hotfix to avoid disturbing develop

drothery-edb (approved):
LGTM - thanks

PR #2257, opened 2022-01-26
Add LiveCompare automation and use a dispatch script.

jericson-edb:
## What Changed?

* Added LiveCompare to the sync-and-process-files workflow
* Moved the logic for which process commands to use out of the workflow and into a dispatch script

## Checklist

**Content**

### DIFF
--- a/scripts/source/dispatch_product.py
+++ b/scripts/source/dispatch_product.py
@@ -1,0 +1,28 @@
+#!/usr/bin/env python3
+
+import argparse
+parser = argparse.ArgumentParser()
+parser.add_argument("repo",
+                    help="Product repository",
+                    type=str)
+parser.add_argument("workspace",
+                    help="GitHub workplace path",
+                    type=str)
+
+args = parser.parse_args()
+
+import os
+
+commands = {'EnterpriseDB/cloud-native-postgresql':
+            f'{args.workspace}/destination/scripts/source/process-cnp-docs.sh {args.workspace}/source {args.workspace }/destination',
+            'EnterpriseDB/fe':
+            f'mkdir -p {args.workspace}/destination/icons-pkg && \
+              cp -fr utils/icons-placeholder/output/* {args.workspace}/destination/icons-pkg/',
+            'EnterpriseDB/LiveCompare':
+            f'node {args.workspace}/destination/scripts/source/livecompare.js'
+            }
+            
+
+cmd = commands[args.repo]
+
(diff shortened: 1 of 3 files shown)


scripts/source/dispatch_product.py
```diff
@@ -0,0 +1,28 @@
...
+            f'{args.workspace}/destination/scripts/source/process-cnp-docs.sh {args.workspace}/source {args.workspace }/destination',
+            'EnterpriseDB/fe':
+            f'mkdir -p {args.workspace}/destination/icons-pkg && \
+              cp -fr utils/icons-placeholder/output/* {args.workspace}/destination/icons-pkg/',
+            'EnterpriseDB/LiveCompare':
+            f'node {args.workspace}/destination/scripts/source/livecompare.js'
+            }
+            
+
+cmd = commands[args.repo]
```
josh-heyer, line 26: Error handling could be a bit better here: if the source repo hasn't been added to this script yet, this'll fail with a `KeyError`. Should probably check to see if the command exists in the dictionary and fail with something descriptive (as in the old switch fallback) when it does not.

scripts/source/livecompare.js
```diff
@@ -19,7 +19,7 @@ const { exec, execSync } = require("child_process");
 const isAbsoluteUrl = require("is-absolute-url");
 
 const fileToMetadata = {};
-const basePath = path.resolve("temp_livecompare/docs/");
+const basePath = path.resolve("source/docs/");
```
josh-heyer, line 22: probably just make this a parameter for consistency with the CNP flow. Also the destination repo path, as right now this script assumes (line 50) that it'll be invoked from the root

josh-heyer (changes requested):
Haven't tested this, but two things jump out at me

scripts/source/dispatch_product.py
jericson-edb: Oh yeah. I thought of that earlier, but it slipped my mind. It might not hurt do do something with the return code at the end of the script too in case the command fails.

jericson-edb:
> Haven't tested this, but two things jump out at me

Straightened those out, I believe. I'm playing around with [act](https://github.com/nektos/act) to see if I can test locally, but I don't think it'll be enough to know for sure if it works. (It does tell me there isn't a syntax error in the workflow, which is nice.)

josh-heyer (approved):
Fixed a typo & tested - looks good to me!

PR #2259, opened 2022-01-26
style: format using prettier and black

gsong:
@jericson-edb and @josh-heyer, I'd like us to get into the habit of using code formatters:

* Python: black
* JS: prettier

### DIFF
--- a/scripts/source/livecompare.js
+++ b/scripts/source/livecompare.js
@@ -20,7 +20,7 @@ const isAbsoluteUrl = require("is-absolute-url");
 
 const fileToMetadata = {};
 const args = process.argv.slice(2);
-const basePath = path.resolve(args[0], 'docs');
+const basePath = path.resolve(args[0], "docs");
 const destination = path.resolve(args[1]);
 
 (async () => {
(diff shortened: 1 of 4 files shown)


scripts/source/livecompare.js
```diff
@@ -59,29 +65,40 @@ const destination = path.resolve(args[1]);
...
+  );
 })();
 
 // GPP leaves the files littered with these; they alter parsing by flipping sections to HTML context
 // remove them BEFORE parsing to avoid issues
-function stripEmptyComments(file)
-{
-  file.contents = file.contents.toString().replace(/<!--\s*-->/g, '');
+function stripEmptyComments(file) {
+  file.contents = file.contents.toString().replace(/<!--\s*-->/g, "");
```
gsong, line 98: [suggest] I'm not a fan of functions that mutate incoming params. If possible, I would refactor so there are no side-effects to calling a function.

josh-heyer: Reasonable; I've adjusted the other two BDR-ecosystem scripts in this regard as well. (at some point need to adapt Jake's scripts to a unifiedjs pipeline and remove all the redundancy here)

josh-heyer (approved):

PR #2261, opened 2022-01-27
BigAnimal: Additional restore info 

drothery-edb:
## What Changed?

Updated Backing up and restoring topic based on feedback from free trail feedback


drothery-edb:
Received approval from Natalia and Guang in Slack

PR #2263, opened 2022-01-27
Move node packages to dependancies

jericson-edb:
## What Changed?

 I just moved some node packages from `devDependencies` to `dependencies`. I think it fixes [the problem with scripts/source/livecompare.js](https://github.com/EnterpriseDB/docs/runs/4968354738?check_suite_focus=true).


josh-heyer (approved):

josh-heyer (changes requested):
This won't work; check out how it's done for CNP

jericson-edb:
> This won't work; check out how it's done for CNP

Got it. I reset the branch (and force pushed) and tried to duplicate CNP.

scripts/source/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "fileprocessor",
```
josh-heyer, line 2: probably give it a different name

josh-heyer (approved):
at some point soon, we should roll these into fileprocessor and avoid the issue, but for now probably call the package something else. Other than that, +1

josh-heyer (approved):

scripts/source/package.json
jericson-edb: Good call. And it really does make sense to unify them. I'll look at doing that next week when we tackle the other docs.

josh-heyer (approved):

PR #2270, opened 2022-01-28
BigAnimal Connecting your account topic: fixing hyphens in create-spn.sh curl command

jasonparton:
- Fixed typo in the syntax of the create-spn guidance.
- Original: "The syntax of the command is:

curl -sL https://raw.githubusercontent.com/EnterpriseDB/cloud-utilities/main/azure/create-spn.sh | bash -s \
  -- --display-name <display-name> \
  ----subscription <subscription> \
  -- -years <years>"

- Corrected: "The syntax of the command is:

curl -sL https://raw.githubusercontent.com/EnterpriseDB/cloud-utilities/main/azure/create-spn.sh | bash -s \
  -- -display-name <display-name> \
  -- -subscription <subscription> \
  -- -years <years>"

## Checklist

**Content**

### DIFF
--- a/product_docs/docs/biganimal/release/getting_started/02_connect_cloud_account.mdx
+++ b/product_docs/docs/biganimal/release/getting_started/02_connect_cloud_account.mdx
@@ -48,10 +48,10 @@ Alternatively, you can run the script in [Azure Cloud Shell](https://azure.micro
 The syntax of the command is:
 
 ```
-curl -sL https://raw.githubusercontent.com/EnterpriseDB/cloud-utilities/main/azure/create-spn.sh | bash -s \
-  -- -display-name <display-name> \
-  -- -subscription <subscription> \
-  -- -years <years>
+curl -sL https://raw.githubusercontent.com/EnterpriseDB/cloud-utilities/main/azure/create-spn.sh | bash -s -- \
+--display-name "hello-s" \
+--subscription c808xxxx-xxxx-xxxx-xxxx-xxxxxxxxb712 \
+--years 1
 ```
 
 Flag and option details:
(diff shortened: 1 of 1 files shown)


drothery-edb:
@Simonzhaohui , @jasonparton what is the purpose of the standalone double hyphens " -- " in the curl syntax? Also, are the options supposed to have one or two hyphens? I'm asking because in the curl syntax we show in https://www.enterprisedb.com/docs/biganimal/latest/getting_started/01_preparing_azure/#method-1-use-edbs-shell-script, the options have two hyphens (--help) and Jason, you changing these to one hyphen (-year).

jericson-edb:
> @Simonzhaohui , @jasonparton what is the purpose of the standalone double hyphens " -- " in the curl syntax? Also, are the options supposed to have one or two hyphens? I'm asking because in the curl syntax we show in https://www.enterprisedb.com/docs/biganimal/latest/getting_started/01_preparing_azure/#method-1-use-edbs-shell-script, the options have two hyphens (--help) and Jason, you changing these to one hyphen (-year).

The options should be two hyphens. One hyphen is only for the short version of the options:

```
Options:
  -d, --display-name: The name of Azure AD App.
  -s, --subscription: The Azure Subscription ID used by BigAnimal.
  -y, --years:        [Optional] The Number of years for which the credentials will be valid. Only accept positive integer value. Default: 1 year.
  -i, --id:           [Optional] The Application (client) ID of Service Principal will be updated.
  -h, --help:         Show this help.
```

The command is running `curl` to fetch the `create-spn.sh` script. Instead of saving the script, it runs it immediately using `bash -s`. The -s option tells `bash` to use the commands from stdin rather than a script on disk. A double hyphen tell `bash` to ignore the following options and pass them to the script it's running. We only need one double hyphen for that purpose, not one per option. As soon as `bash` sees the double hyphen, it ignores the rest of the options.

A slightly simpler script would be:

```
curl -sL https://raw.githubusercontent.com/EnterpriseDB/cloud-utilities/main/azure/create-spn.sh \
| bash -s -- \
--display-name "hello-s" \
--subscription c808xxxx-xxxx-xxxx-xxxx-xxxxxxxxb712 \
--years 1
```

(That still fails because I don't have the right subscription ID. ;-)

jericson-edb:
Another way to do it is:

```
curl -sL https://raw.githubusercontent.com/EnterpriseDB/cloud-utilities/main/azure/create-spn.sh \
| bash -s -- \
-d "hello-s" \
-s c808xxxx-xxxx-xxxx-xxxx-xxxxxxxxb712 \
-y 1
```

That uses the short options, which is shorter. For the sake of documentation, the longer options is better. `-d` could mean anything. ;-)

drothery-edb:
See related PR: https://github.com/EnterpriseDB/docs/pull/2273

product_docs/docs/biganimal/release/getting_started/02_connect_cloud_account.mdx
```diff
@@ -48,10 +48,10 @@ Alternatively, you can run the script in [Azure Cloud Shell](https://azure.micro
 The syntax of the command is:
 
 ```
-curl -sL https://raw.githubusercontent.com/EnterpriseDB/cloud-utilities/main/azure/create-spn.sh | bash -s \
-  -- -display-name <display-name> \
-  -- -subscription <subscription> \
-  -- -years <years>
+curl -sL https://raw.githubusercontent.com/EnterpriseDB/cloud-utilities/main/azure/create-spn.sh | bash -s -- \
+--display-name "hello-s" \
```
drothery-edb: I think we are mixing syntax and examples by using actual values for the parameters in this section. I think we should use the <user_supplied_info> convention in the syntax section. There is an actual example farther down in the page where we can use specific values.

drothery-edb (approved):

PR #2273, opened 2022-01-28
BigAnimal: using help output for usage info for scripts

drothery-edb:
## What Changed?

Trying new approach for script usage


jericson-edb:
I do like the idea of just using the help output, which is clear and likely to be accurate if we remember to update when the command is updated. (And this could be automated!) But... there is also value in just showing a command that works if you copy and paste it. This is especially useful if you need to run a script only occasionally. Particularly for installing something, the "pipe a curl command to bash" is a common idiom that we probably should use. (Examples I've seen just this week: [installing `act`](https://github.com/nektos/act#bash-script), [azure-cli](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli-linux?pivots=apt#option-1-install-with-one-command) and [Big Animal CLI](https://www.enterprisedb.com/docs/biganimal/latest/reference/cli/#download-the-binary-executable).) The nice thing about that approach is we don't have to worry about people having old copies of the script&mdash;they are just getting the most recent copy from the source.

It would be ideal if the command were less confusing. One option would be to see if we could use environment variables instead of passing options. It also helps, I think, to put the "magic" on one line so that people kinda accept it without needing to think too hard:

```
curl -sL https://raw.githubusercontent.com/EnterpriseDB/cloud-utilities/main/azure/create-spn.sh | bash -s -- \
--display-name "hello-s" \
--subscription c808xxxx-xxxx-xxxx-xxxx-xxxxxxxxb712 \
--years 1
```
The entire `curl` => `bash` line is just a way to let you run `create-spn.sh` without having to save it somewhere. It's such a common use case, I wish `curl` supported it directly. The other three lines have parameters you need to adjust.

drothery-edb:
@jericson-edb , do you think we are getting closer? Might make sense to have a call about this and have a live working session.

drothery-edb:
Closing. Not quite the right strategy.

Issue #1779, opened 2021-08-20
Stop forcing Netlify to process all of our files when not all of them have changed.


jericson-edb:
[Background](https://github.com/gatsbyjs/gatsby/issues/11961#issuecomment-466029938):

> The issue that Jed is describing is that if you are doing cache busting, and that cache busting is changing the contents of all of your assets, that means we will end up re-processing ALL of the files in every single deploy. This will slow down your deploy since it means you can't take advantage of the optimizations we have. If all you do is rename a file, this shouldn't break things since the SHA is only on file contents and not metadata. But if the cache busting makes even a minor change to the file contents then we will end up reprocessing it and re-uploading it.

[Gatsby is one culprit](https://github.com/gatsbyjs/gatsby/issues/11961#issuecomment-466192050):

> The root of the problem is webpack-runtime-[HASH].js, which changes in every build and is included in all pages. 

There is [a plugin](https://www.gatsbyjs.com/plugins/gatsby-plugin-remove-fingerprints/) that removes these build fingerprints. That might not be sufficient if we have other parts of the build process adding cache-breaking changes.

Note: it's fine to have changes such as the modification date in [the footer of every page](https://www.enterprisedb.com/docs/epas/latest/epas_compat_ora_dev_guide/):

![image](https://user-images.githubusercontent.com/83660216/130271864-fc713324-3ed5-48a1-8066-4dbda5402153.png)

But it's not helpful to have changes that happen when we build if the file content has not changed.

PR #2281, opened 2022-01-29
Netlify deploy optimization

josh-heyer:
## What Changed?

Netlify deploys take too long. A big part of this is that 6K+ files are hashing differently from build to build.

Turns out, a big part of this is the list of client-side redirects stored in app-<hash>.js. If the order changes (and it does) then the hash on that file changes, which in turn is propagated to every HTML file - and now those also hash differently.

There are a few possible solutions to this:

- ensure pages are added in the same order every time
- don't do client-side redirects
- sort the list of pages

Starting off with the last one, because it is easy. Reduces the no-change diff between builds from 6K+ to about 10 files. Which is still... 9 more than it should be (1 will always change as it contains the build timestamp). But, an improvement!

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -161,6 +161,11 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
 
   processFileNodes(result.data.allFile.nodes, actions);
 
+  // this is critical to avoiding excessive Netlify deploy times: it ensures the pages are ordered consistently from build to build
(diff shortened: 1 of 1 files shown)


josh-heyer:
Could use some more honing, but I think this is a significant improvement as-is; aiming it as a fix for #1779 @jericson

jericson-edb (approved):
Awesome! Should help.

Issue #2294, opened 2022-02-02
Feedback on /migration_portal/3.5.0/04_mp_migrating_database/01_mp_schema_extraction.mdx


mlewandowski-edb:
Step 5 under the "To extract all schemas in a database" heading directs the user to create a `par` file (named parameter2.par); however, the file does not appear to be used/referenced in subsequent steps.  Should the `impdp` command in Step 6 have included the command line option to use it?

mlewandowski-edb:
@mikram-edb @umangjtank @moazzumali
I discovered this issue earlier today when reviewing the Migration Portal 3.5.0 documentation.  I ran a test a little later in the day and confirmed that the `PARFILE` parameter set to the parameter file created in Step 5 to the `impdb` command in Step 6 will result in the system schemas being excluded in the result SQL file; whereas, if the `PARFILE` is not specified, they are included.  So, the documented `impdp` command in the "To extract all schemas in a database" should be updated to include the `PARFILE` parameter.

mikram-edb:
@mlewandowski-edb ,

I can see impdp command in step 6 is using the parfile created under step 5.

$ impdp <Username>@<ConnectIdentifier> DIRECTORY=<directory_name> TRANSFORM=OID:n SQLFILE=YourSchemas.sql DUMPFILE=db.dump **parfile=parameters2.par**

mlewandowski-edb:
Hmmm. Here is the command copied from Step 6:

` $ impdp <Username>@<ConnectIdentifier> DIRECTORY=<directory_name> TRANSFORM=OID:n SQLFILE=YourSchemas.sql DUMPFILE=db.dump
`
as shown in the document at this location:
https://www.enterprisedb.com/docs/migration_portal/latest/04_mp_migrating_database/01_mp_schema_extraction/

![image](https://user-images.githubusercontent.com/63003848/152285051-cfee0b72-e4df-451d-9f31-ec484ae5d59f.png)

mikram-edb:
Sorry Matt, I did not see your latest comment. Yes we got it and will be rectifying. Thanks for pointing out.

moiznalwalla:
I have created a Pull request to track these changes, [here](https://github.com/EnterpriseDB/docs/pull/2298).

Staging link:https://61fbb96e804d948463c3d612--edb-docs-staging.netlify.app/docs/migration_portal/latest/04_mp_migrating_database/01_mp_schema_extraction/

[...]

PR #2298, opened 2022-02-03
MP: Issue #2294 - Extracting schema changes

moiznalwalla:
Changes suggested after GA doc review.


mlewandowski-edb:
The changes look good.  However, I think the text boxes with the content that the users can copy should be updated to only include the information that the user would want to paste into their interface.  

For example, the text boxes for the `export.par` and `import.par` content include the `cat` commands.  I understand that this command is used to list the contents of a file (on linux or mac).  The command would not be applicable for Windows users using a DOS command prompt.  Also, we wouldn't want users to mistakingly copy that command into their parameter and not remove it.   I recommend removing the `cat` command from those boxes.

Similarly, the "`$ `" string in the `expdp` and `impdp` command text boxes represent the terminal prompt.  Leaving them in the copied command would result in an error.  Therefore, I recommend removing it from the text boxes.

mlewandowski-edb (approved):
The changes seem to align with my recommendations.  Will the version in staging be updated to include them for a final review before pushing to production?

drothery-edb:
@josh-heyer I added the redirect. Do you want to take a look before I merge and start the build?

josh-heyer (approved):

PR #2301, opened 2022-02-03
Release: 2022-02-03

drothery-edb:
## What Changed?

MP: Fixed extraction instructions
pgPool: fixed wording in rel notes


josh-heyer (approved):
FYI, I rebased this to include #2281 - that'll have to get built out on both develop and main before we'll see any improvement, so might as well get on it.

PR #2305, opened 2022-02-08
Use '-kvno 0' while generating the keytab files.

asheshv:
## What Changed?

When customer uses the same SPN (Domain User) for generating the keytab files, it ends up updating 'kvno' on every newly generated keytab files, and he won't be able to login to PEM Server ever.


drothery-edb (approved):
LGTM - do you want me to merge this change for you?

nidhibhammar:
I have confirmed with Ashesh on Slack to go ahead and merge this change. Thanks

PR #2311, opened 2022-02-08
Fixed a typo and add BDR

jericson-edb:
## What Changed?

* Fixed a typo in dispatch_product.py
* Added BDR automation

### DIFF
--- a/scripts/source/bdr.js
+++ b/scripts/source/bdr.js
@@ -20,7 +20,7 @@ const isAbsoluteUrl = require("is-absolute-url");
 
 const fileToMetadata = {};
 const args = process.argv.slice(2);
-const basePath = path.resolve(args[0], "docs/docs2/");
+const basePath = path.resolve(args[0], "docs/");
 const imgPath = path.resolve(args[0], "docs/img/");
 const destination = path.resolve(args[1]);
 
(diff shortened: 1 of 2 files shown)


scripts/source/bdr.js
```diff
@@ -52,12 +54,18 @@ const imgPath = path.resolve("temp_bdr/docs/img/");
   };
 
   const mdIndex = yaml.load(
-    await fs.readFile(path.resolve(basePath, "bdr-pub.yml"), "utf8"),
+      await fs.readFile(path.resolve(args[0], 'docs', "bdr-pub.yml"), "utf8"),
```
josh-heyer, line 57: Is this correct? No index in docs2?

jericson-edb: Looks like they removed docs2 from the repo. I've pulled out the docs2 in the rest of the script. (There's a problem parsing bdr-pub.yml at the moment too.)

scripts/source/bdr.js
```diff
@@ -19,8 +19,10 @@ const { exec } = require("child_process");
 const isAbsoluteUrl = require("is-absolute-url");
 
 const fileToMetadata = {};
-const basePath = path.resolve("temp_bdr/docs/docs2/");
-const imgPath = path.resolve("temp_bdr/docs/img/");
+const args = process.argv.slice(2);
+const basePath = path.resolve(args[0], "docs/");
+const imgPath = path.resolve(args[0], "docs/img/");
```
josh-heyer, line 24: minor nit, but I'd probably build this on basePath vs referencing args[0] twice; just makes it a wee bit easier to read backwards

josh-heyer (approved):

scripts/source/bdr.js
dougcalobrisi: So, we never had `docs2` in `master` or a STABLE branch. This is because we preprocess the markdown and don't want to put the processed md in a main branch. When we've handed off previously, it's been in a separate branch and does have a `docs/docs2/` directory. We'll likely need to do it this same way with the automation, as we can't give you the actual source docs directly or the doc we want to keep confidential will end up in the public docs repo.

josh-heyer: Ok. Jon's re-added docs2 in #2316, and will work with you to ensure that the automation triggers only run on the correct branches BDR-side. Having a different directory structure adds a bit of extra protection here, as the script will fail if the wrong branch gets triggered - so no protected content will end up here even if we have a glitch setting up triggers on the BDR side.

PR #2332, opened 2022-02-14
SLES 12 updates

drothery-edb:
## What Changed?

Added instructions for SLES 12 on x86 and ppc64le


drothery-edb:
Merged into SLES 15 branch

PR #2336, opened 2022-02-15
Test sync-and-process action

jericson-edb:
## What Changed?

* Added test for the sync-and-process action
* Add pglogical
* Add HARP (broken in testing because I don't have rsync installed in the image)
* Don't try creating a PR in testing
* Make sure Node scripts fail properly if they don't work.


josh-heyer (approved):

josh-heyer:
Will be nice to have tests for this!

PR #2348, opened 2022-02-17
Updated MP 3.5 rel notes

drothery-edb:
## What Changed?

Clarified in the rel notes that Migration Portal 3.5.0 no longer supports the assessment of files generated by the previously provided EDB DDL extractor


mlewandowski-edb (changes requested):
The Type corresponding to the Description is an Enhancement; however, the change to the migration portal is not really an enhancement.  @drothery-edb suggesting changing `Enhancement` to `Change`.  I agree with making that update.

PR #2350, opened 2022-02-17
MP rel notes prototype

drothery-edb:
Created two prototypes: one using our standard table format and one using the text-based format used in the MP wiki


drothery-edb:
No longer need the prototype. It has served its purpose.

PR #2361, opened 2022-02-21
BigAnimal: updates for UPM-4607 - support for GUC equations

drothery-edb:
## What Changed?

Added a section entitled "Using equations for parameter values" in the "Modifying database configuration parameters" topic

### DIFF
--- a/product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
+++ b/product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
@@ -36,7 +36,7 @@ If you enter a specific value for the `shared_buffers` parameter, unless you spe
 
 Alternatively, with GUC equations, you can specify that the value for `shared_buffers` should be 25% of RAM, but no higher than 8GB:
 
-((ram \* 0.25) > 8 GB) ? 8 GB : (ram \* 0.25)
+`((ram \* 0.25) > 8 GB) ? 8 GB : (ram \* 0.25)`
 
 
 ## Modify a parameter
--- a/product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
+++ b/product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
@@ -46,14 +46,45 @@ If a target GUC has a unit, a bare number in an equation is assumed to be in the
 ### Recursion
 We recommend no more than approximately 100 levels of recursion depth.
 
-### Example 
+### Examples 
 
-This example uses a formula for the `shared_buffers` parameter value.
-
-Instead of entering a specific value for the `shared_buffers` parameter, with GUC equations, you can specify that the value for `shared_buffers` should be 25% of RAM, but no higher than 8GB: 
+Instead of entering a specific value for the `shared_buffers` parameter, with GUC equations you can specify that the value for `shared_buffers` should be a certain percentage of RAM, but no higher than a certain size. For example: 
 
 `((ram * 0.25) > 8 GB) ? 8 GB : (ram * 0.25)`
 
+What if we want `wal_buffers` to be three percent of `shared_buffers`, up to a maximum of 16MB? The formula is:
+
+`((shared_buffers * 0.03) > 16MB) ? 16MB : (shared_buffers * 0.03)`
+
+The formula refers to the `shared_buffers`, which, in turn, has its own formula in this example. This second example shows how GUCs can refer to other GUCs.
+
+### Common errors and error handling
+
+alformed formulas result in an errors. Here are some examples:
+
+- Following a number with two units, for example: 
+
+  `work_mem = 10 GB kB`
+
+  produces the error:
+
+  `unexpected token "kB" at end of input (line 1, col 7)`
+
+- Creating an infinite evaluation loop by having a GUC referring to itself, for example:
+
+  `effective_cache_size = effective_cache_size`
+
+  gives the error:
+
+  `cycle in equation; GUC "effective_cache_size" already seen`
+
+- Mixing unit types, for example:
+
+  `work_mem = 10kB * 8s`
+
+  gives the error:
+
+  `both size and time units used in calculation; cannot proceed`
 
 ## Modify a parameter
 
--- a/product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
+++ b/product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
@@ -20,7 +20,7 @@ Not all database configuration parameters are supported by BigAnimal. Some param
 
 ## Using formulas for parameter values
 
-In addition to entering specific values for parameters, for some parameters you can specify formulas to automatically calculate a value. You can use formulas for parameters of type integer and real using the following operators: `+ - / * > >= < <= == != && || ! ? : ( )`. GUCs used in formulas must also be of type integer or real. All arithmetic is done on 64-bit floating point values rounded to an integer result. 
+In addition to entering specific values for parameters, for some parameters you can specify formulas to automatically calculate a value. You can use formulas for parameters of type integer and real using the following operators: `+ - / * > >= < <= == != && || ! ? : ( )`. Use `?` amd `:` in ternary formulas such as the [shared buffer example](#example). Use `( )` to specify [order of operations](order-of-operations), if needed. GUCs used in formulas must also be of type integer or real. All arithmetic is done on 64-bit floating point values rounded to an integer result.  
 
 BigAnimal has what we refer to as psuedo GUCs to facilitate creating equations. These read-only GUCs are:
 - `cpu_cores` &#8212 number of cores provided by the VM running the cluster nodes
@@ -29,10 +29,22 @@ BigAnimal has what we refer to as psuedo GUCs to facilitate creating equations. 
 - `ram` &#8212 amount of memory provided by the VM running the cluster nodes, in KB
 - `replica_count` &#8212 total number of nodes in the cluster
 
-You can't use GUCs with different types of units (such as size or duration) in formulas. For example, you can't use a formula that tries to multiply two seconds with three kilobytes (2s \* 3KB), but you can use a formula that multiplies two sizes even if they have different scales (2KB \* 8GB).
-
 Formulas can calculate a value relative to the specifications you selected for the cluster or relative to other non-pseudo GUCs. If you resize your cluster, BigAnimal recalculates formulas using the new values for the pseudo GUCs. 
 
+### Order of operations
+`*` and `/` are at the same operator precedence. `+` and `-` are at the same operator precedence. `*` and `/` are evaluated before `+` and `-`. Use parentheses to specify additional order of operations. 
+
+### Units
+GUCs can either be a bare number, such as 8, or have a unit, such as GB. Units are categorized into classes such as size or duration.
+
+You can't use GUCs with different classes of units in formulas. For example, you can't use a formula that tries to multiply two seconds with three kilobytes (2s \* 3KB), but you can use a formula that multiplies two sizes even if they have different scales (2KB \* 8GB). 
+
+The unit of the target GUC must match the unit used in the formula. For example, you can multiply two seconds by four days but you can't assign the result to a GUC of unit KB.
+
+If a target GUC has a unit, a bare number in an equation is assumed to be in the default unit for the target GUC.
+
+### Recursion
+We recommend no more than approximately 100 levels of recursion depth.
 
 ### Example 
 
--- a/product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
+++ b/product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
@@ -20,7 +20,7 @@ Not all database configuration parameters are supported by BigAnimal. Some param
 
 ## Using formulas for parameter values
 
-In addition to entering specific values for parameters, for some parameters you can specify formulas to automatically calculate a value. 
+In addition to entering specific values for parameters, for some parameters you can specify formulas to automatically calculate a value. You can use equations for parameters of type integer and real using the following operators: `+ - / * > >= < <= == != && || ! ? : ( )`. GUCs used in formulas must also be of type integer or real.
 
 BigAnimal has what we refer to as psuedo GUCs to facilitate creating equations. These read-only GUCs are:
 - `cpu_cores`
@@ -29,16 +29,16 @@ BigAnimal has what we refer to as psuedo GUCs to facilitate creating equations. 
 - `ram`
 - `replica_count` (node count)
 
-Formulas can calculate a value relative to the specifications you selected for the cluster or relative to two non-pseudo GUCs.
+`ram` and `disk_size` have the default unit of kB. 
 
-You can use equations for parameters of type integer and real. GUCs used in formulas must also be of type integer or real.
+Formulas can calculate a value relative to the specifications you selected for the cluster or relative to two non-pseudo GUCs. If you resize your cluster, BigAnimal recalculates formulas using the new values for the pseudo GUCs. 
 
 
 ### Example 
 
 This example uses a formula for the `shared_buffers` parameter value.
 
-Instead of entering a specific value for the shared_buffers parameter, with GUC equations, you can specify that the value for shared_buffers should be 25% of RAM, but no higher than 8GB: 
+Instead of entering a specific value for the `shared_buffers` parameter, with GUC equations, you can specify that the value for `shared_buffers` should be 25% of RAM, but no higher than 8GB: 
 
 `((ram * 0.25) > 8 GB) ? 8 GB : (ram * 0.25)`
 
--- a/product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
+++ b/product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
@@ -15,9 +15,9 @@ The list of parameters is populated based on the type of database you selected o
 Not all database configuration parameters are supported by BigAnimal. Some parameters, such as `wal_level` and `restore_command`, are reserved for EDB to provide the managed database features of BigAnimal. 
 !!!
 
-## Using equations for parameter values
+## Using formulas for parameter values
 
-In addition to entering specific values for parameters, for some parameters you can specify equations to automatically calculate a value relative to the machine specifications you selected for the cluster. 
+In addition to entering specific values for parameters, for some parameters you can specify formulas to automatically calculate a value relative to the machine specifications you selected for the cluster. 
 
 Parameters have types and you can use equations for parameters of type integer and real. See [Parameter Names and Values](https://www.postgresql.org/docs/current/config-setting.html#CONFIG-SETTING-NAMES-VALUES) for more information on the types of values parameters accept.
 
@@ -30,7 +30,7 @@ BigAnimal has what we refer to as psuedo GUCs to facilitate creating equations. 
 
 ### Example 
 
-This example uses an equation for the `shared_buffers` parameter value.
+This example uses a formula for the `shared_buffers` parameter value.
 
 If you enter a specific value for the `shared_buffers` parameter, unless you specify another unit, the number represents the number of eight kilobyte pages. For example, 1572864.
 
--- a/product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
+++ b/product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
@@ -15,6 +15,30 @@ The list of parameters is populated based on the type of database you selected o
 Not all database configuration parameters are supported by BigAnimal. Some parameters, such as `wal_level` and `restore_command`, are reserved for EDB to provide the managed database features of BigAnimal. 
 !!!
 
+## Using equations for parameter values
+
+In addition to entering specific values for parameters, for some parameters you can specify equations to automatically calculate a value relative to the machine specifications you selected for the cluster. 
+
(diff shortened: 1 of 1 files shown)


nataliawojcik27 (dismissed):
lgtm

product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
```diff
@@ -15,6 +15,30 @@ The list of parameters is populated based on the type of database you selected o
...
+- `disk_iops`
+- `disk_size`
+- `ram`
+- `replica_count`
+
+### Example 
+
+This example uses a formula for the `shared_buffers` parameter value.
+
+If you enter a specific value for the `shared_buffers` parameter, unless you specify another unit, the number represents the number of eight kilobyte pages. For example, 1572864.
```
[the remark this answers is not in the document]
drothery-edb: Addressed in commit: https://github.com/EnterpriseDB/docs/pull/2361/commits/46293599cdfd3d65edfb06bc53ba0fed3b17f27f

product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
```diff
@@ -15,6 +15,30 @@ The list of parameters is populated based on the type of database you selected o
...
+In addition to entering specific values for parameters, for some parameters you can specify formulas to automatically calculate a value relative to the machine specifications you selected for the cluster. 
+
+Parameters have types and you can use equations for parameters of type integer and real. See [Parameter Names and Values](https://www.postgresql.org/docs/current/config-setting.html#CONFIG-SETTING-NAMES-VALUES) for more information on the types of values parameters accept.
+
+BigAnimal has what we refer to as psuedo GUCs to facilitate creating equations. These read-only GUCs are:
+- `cpu_cores`
+- `disk_iops`
+- `disk_size`
+- `ram`
+- `replica_count`
```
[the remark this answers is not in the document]
ba-edb: > Ben, when you say PGD cluster do you mean EHA specifically?

Yes, I'm on a mission to eliminate the "EHA" initialism 😁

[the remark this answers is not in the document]
drothery-edb: @ba-edb do you know what the replica count is for those cluster types or is that a question directed to @edbmanniwood?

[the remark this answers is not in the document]
ba-edb: Definitely a Manni question 🌊

product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
```diff
@@ -15,6 +15,30 @@ The list of parameters is populated based on the type of database you selected o
 Not all database configuration parameters are supported by BigAnimal. Some parameters, such as `wal_level` and `restore_command`, are reserved for EDB to provide the managed database features of BigAnimal. 
 !!!
 
+## Using formulas for parameter values
+
+In addition to entering specific values for parameters, for some parameters you can specify formulas to automatically calculate a value relative to the machine specifications you selected for the cluster.
```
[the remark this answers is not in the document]
drothery-edb: Did some restructuring and editing to address this comment. See https://github.com/EnterpriseDB/docs/pull/2361/commits/8823345ab9e62cc5d4e64bc7af97fe919628138c

product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
```diff
@@ -15,6 +15,30 @@ The list of parameters is populated based on the type of database you selected o
...
+
+### Example 
+
+This example uses a formula for the `shared_buffers` parameter value.
+
+If you enter a specific value for the `shared_buffers` parameter, unless you specify another unit, the number represents the number of eight kilobyte pages. For example, 1572864.
+
+Alternatively, with GUC equations, you can specify that the value for `shared_buffers` should be 25% of RAM, but no higher than 8GB:
+
+`((ram \* 0.25) > 8 GB) ? 8 GB : (ram \* 0.25)`
```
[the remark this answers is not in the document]
drothery-edb: @edbmanniwood need your help here.

product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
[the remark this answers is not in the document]
drothery-edb: @ba-edb see https://github.com/EnterpriseDB/docs/pull/2361/commits/86d1887e2617c5faee926d4fec4767dbdbbbf047

product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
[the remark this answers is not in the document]
edbmanniwood: Operators available: 
`
+ 
- 
/ 
* 
> 
>=
< 
<=
==
!=
&&
||
! 
? 
: 
( 
) 
`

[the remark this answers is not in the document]
edbmanniwood: There is no `mod` available. There is integer arithmetic, but see UPM-8554 for why we might do all arithmetic in floating point and then just convert to int at the very end for GUCs that require type in (and not float, also known in PostgreSQL GUC documentation as "real").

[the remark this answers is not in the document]
edbmanniwood: Equations only work for target GUCs whose type is integer or real. Attempting to enter an equation for a string GUC will simply take the equation at face value. On the other hand, attempting to enter a string in an equation for an integer or real GUC will result in the equation parser looking for operators, numbers, or the names of other GUCs. So most likely, a random string in a GUC equation would yeild a parser error such as `unexpected token "foo" at line 1, col 1`.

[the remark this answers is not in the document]
edbmanniwood: If you divide one shorthand, such as `8 GB` by another shorthand, such as `1 MB`, the parser will look at the default unit for the target GUC. Let's say the default unit for the target GUC is "pages". (Each page is 8kB.) The evaluator will first convert `8GB` to the equivalent number of pages; the evaluator will then convert `1 MB` to the equivalent number of pages; the evaluator will then evaluate the equation and then assign that value to the target GUC.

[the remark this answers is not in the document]
edbmanniwood: You can divide a shorthand by an integer. The above-noted conversions will take place for the shorthand, whereas the bare integer will be assumed to be in the default unit for the target GUC.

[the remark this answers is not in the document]
edbmanniwood: If you re-size your cluster, all GUC equations will be re-run using the target machine's pseudo GUCs. For instance, if an equation is `ram / 2` then on an 8 GB machine expect the answer to be 4 GB and on a 16 GB machine, expect the answer to be 8 GB.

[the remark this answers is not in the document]
drothery-edb: Thanks, @edbmanniwood. @ba-edb I added a bit more detail based on Manni's answers to your questions. See https://github.com/EnterpriseDB/docs/pull/2361/commits/5a5ed7266a360f7d2c059fee3b7e6984a702bb1f

edbmanniwood (commented):
Two small comments. Nothing big.

edbmanniwood (approved):
This seems good to me.

product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
```diff
@@ -6,15 +6,45 @@ The database parameters listed on the DB Configuration tab are also referred to
...
+BigAnimal has what we refer to as psuedo GUCs to facilitate creating equations. These read-only GUCs are:
+- `cpu_cores`
+- `disk_iops`
+- `disk_size`
+- `ram`
+- `replica_count` (node count)
+
+`ram` and `disk_size` have the default unit of KB. 
+
+You can't use GUCs with different types of units (such as size or duration) in formulas. For example, you can't use a formula that tries to multiply two seconds with three kilobytes (2s \* 3KB), but you can use a formula that multiplies two sizes even if they have different scales (2KB \* 8GB).
```
edbmanniwood, line 34: If we could get rid of the backslashes here, that would be great.

drothery-edb: Those backspaces are escaping the asterisks so that Markdown doesn't output italics. They do not appear in the rendered content.

product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
```diff
@@ -6,15 +6,45 @@ The database parameters listed on the DB Configuration tab are also referred to
...
+-  For more information on the types of values parameters accept, see [Parameter Names and Values](https://www.postgresql.org/docs/current/config-setting.html#CONFIG-SETTING-NAMES-VALUES). The types apply to both PostgreSQL and EDB Postgres Advanced Server parameters.
+
 
 !!!note
 Not all database configuration parameters are supported by BigAnimal. Some parameters, such as `wal_level` and `restore_command`, are reserved for EDB to provide the managed database features of BigAnimal. 
 !!!
 
+## Using formulas for parameter values
+
+In addition to entering specific values for parameters, for some parameters you can specify formulas to automatically calculate a value. You can use formulas for parameters of type integer and real using the following operators: `+ - / * > >= < <= == != && || ! ? : ( )`. GUCs used in formulas must also be of type integer or real.
```
ba-edb, line 23: The wording around "integer or real" here confuses me a bit; many GUCs have types like seconds, milliseconds, or GBs, right? @edbmanniwood

product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
```diff
@@ -6,15 +6,45 @@ The database parameters listed on the DB Configuration tab are also referred to
...
+In addition to entering specific values for parameters, for some parameters you can specify formulas to automatically calculate a value. You can use formulas for parameters of type integer and real using the following operators: `+ - / * > >= < <= == != && || ! ? : ( )`. GUCs used in formulas must also be of type integer or real. 
+
+BigAnimal has what we refer to as psuedo GUCs to facilitate creating equations. These read-only GUCs are:
+- `cpu_cores`
+- `disk_iops`
+- `disk_size`
+- `ram`
+- `replica_count` (node count)
+
+`ram` and `disk_size` have the default unit of KB.
```
ba-edb, line 32: It seems like this section would read better if it was written something like:

BigAnimal has what we refer to as psuedo GUCs to facilitate creating equations. These read-only GUCs are:
- `cpu_cores`: the number of cores provided by the VM running the cluster nodes
- `disk_iops`: the number of IOPS allocated to the disks backing the primary data volume
- `disk_size`: the amount of memory backing the primary data volume, in KB
- `ram`: the amount of memory provided by the VM running the cluster nodes, in KB
- `replica_count`: the total number of nodes in the cluster

product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
ba-edb: > using the following operators: `+ - / * > >= < <= == != && || ! ? : ( )`

`?` and `:` are not operators in the same sense as the other operators in the system - furthermore I'm not sure that you could ever use the boolean operators as the value of a formula, since boolean GUCs do not accept formula. I think we need more detail here, and probably more examples.

product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
```diff
@@ -6,15 +6,45 @@ The database parameters listed on the DB Configuration tab are also referred to
...
+`ram` and `disk_size` have the default unit of KB. 
+
+You can't use GUCs with different types of units (such as size or duration) in formulas. For example, you can't use a formula that tries to multiply two seconds with three kilobytes (2s \* 3KB), but you can use a formula that multiplies two sizes even if they have different scales (2KB \* 8GB).
+
+Formulas can calculate a value relative to the specifications you selected for the cluster or relative to other non-pseudo GUCs. If you resize your cluster, BigAnimal recalculates formulas using the new values for the pseudo GUCs. 
+
+
+### Example 
+
+This example uses a formula for the `shared_buffers` parameter value.
```
ba-edb, line 41: The fact that this is for `shared_buffers` is somewhat irrelevant, right? It's useful for the description but meaningless elsewhere. How about we construct a somewhat more complex example, e.g., set shared buffers to the formula shown here and then set another GUC to a formula that includes `shared_buffers`?

product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
ba-edb: Two other potential concerns:

1. All arithmetic is on 64-bit floats - that should probably be noted
2. Do we need to talk about order of operations?

ba-edb: We probably also need to talk about the limits on recursion.

product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
ba-edb: We might also want to demonstrate and explain some common errors, e.g., type errors?

product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
drothery-edb: Is this the correct way to address potential concern 1? "All arithmetic is done on 64-bit floating point values rounded to an integer result."

drothery-edb: @edbmanniwood I need your help with Ben's latest round of comments.

edbmanniwood: Ben raises a good point about ? and :. Those are used in ternary expressions only. But they are useful; the sample equation uses a ternary expression! For operator precedence (a.k.a. order of operations) we are at teh mercy of https://github.com/PaesslerAG/gval, which at first glance doesn't seem to document operator precedence. Fortunately, it (and out library) support the all-important parenthesis, so the general advice would be: * and / are at the same operator precedence. + and - are at the same operator precedence. * and / are evaluated before + and - . _Use parenthesis for everything else._

edbmanniwood: Right now we have a maximum recursion depth of 1000, though IMHO that should be smaller; someone (probably RyanO) just picked that number and I preserved that number through every iteration/enhancement of the GUC lexer/parser/evaluator, assuming one day we would decide what the correct number should be. We might just wan to mention that we have a guard on recursion depth but reserve the right to determine the exact depth of recursion. We could promise to allow "around 100".

product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
drothery-edb: @edbmanniwood Ben has suggested we add more examples. Can you help with that?

product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
drothery-edb: Made some changes in https://github.com/EnterpriseDB/docs/pull/2361/commits/9ee2970224513d0887e17ff217378045aac5ab68, but I don't think we are quite there yet. I left the mentions of integer and real, but added more detail around operators and recursion levels.

product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
edbmanniwood: The parser will let users know about malformed equations. For instance, following a number with two units will result in an error.

work_mem = 10 GB kB

produces the error

unexpected token "kB" at end of input (line 1, col 7)

A GUC referring to itself would set up an infinite evaluation loop, so it is disallowed.

effective_cache_size = effective_cache_size

gives the error

cycle in equation;  GUC "effective_cache_size" already seen

Mixing unit types is always a bad idea. The GUC evaluator does not know what the result of 10 kB * 3 s would be.

work_mem = 10kB * 8s

gives the error

both size and time units used in calculation; cannot proceed.

drothery-edb: Thanks, @edbmanniwood. @ba-edb and Manni, how does this look? https://github.com/EnterpriseDB/docs/pull/2361/commits/00ed71215b078cd23944e3b6a48641699d848fc1

product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
```diff
@@ -6,15 +6,86 @@ The database parameters listed on the DB Configuration tab are also referred to
...
+Formulas can calculate a value relative to the specifications you selected for the cluster or relative to other non-pseudo GUCs. If you resize your cluster, BigAnimal recalculates formulas using the new values for the pseudo GUCs. 
+
+### Order of operations
+`*` and `/` are at the same operator precedence. `+` and `-` are at the same operator precedence. `*` and `/` are evaluated before `+` and `-`. Use parentheses to specify additional order of operations. 
+
+### Units
+GUCs can either be a bare number, such as 8, or have a unit, such as GB. Units are categorized into classes such as size or duration.
+
+You can't use GUCs with different classes of units in formulas. For example, you can't use a formula that tries to multiply two seconds with three kilobytes (2s \* 3KB), but you can use a formula that multiplies two sizes even if they have different scales (2KB \* 8GB). 
+
```
edbmanniwood, line 41: Let's get rid of those backslashes.

product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
```diff
@@ -6,15 +6,86 @@ The database parameters listed on the DB Configuration tab are also referred to
...
+The unit of the target GUC must match the unit used in the formula. For example, you can multiply two seconds by four days but you can't assign the result to a GUC of unit KB.
```
edbmanniwood, line 42: Let's make the first sentence. "The kind of unit of the target GUC must match the kind of unit used in the formula." This is so that the user will know it's OK to use kB and MB in the same equation (different units but the same kind) whereas s and kB will not be allowed (because they are different kinds of units.

product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
```diff
@@ -6,15 +6,86 @@ The database parameters listed on the DB Configuration tab are also referred to
...
+
+What if we want `wal_buffers` to be three percent of `shared_buffers`, up to a maximum of 16MB? The formula is:
+
+`((shared_buffers * 0.03) > 16MB) ? 16MB : (shared_buffers * 0.03)`
+
+The formula refers to the `shared_buffers`, which, in turn, has its own formula in this example. This second example shows how GUCs can refer to other GUCs.
+
+### Common errors and error handling
+
+alformed formulas result in an errors. Here are some examples:
```
edbmanniwood, line 63: Let's change "alformed" to "malformed".

product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
edbmanniwood: Hey, @drothery-edb . I left a couple of quick comments. Nothing big. Thanks for your persistence! :bow:

product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
drothery-edb: They don't show up in the output and if I don't escape the asterisks it causes italics. 

image

product_docs/docs/biganimal/release/using_cluster/03_modifying_your_cluster/05_db_configuration_parameters.mdx
drothery-edb: @edbmanniwood I'm struggling with this one a little. The sentence before we define what we mean by classes of units. Can we use "class" instead of "kind"? or maybe "type" or "category"? Or do we need to change the preceding paragraph?

I left the proceeding paragraph and used "category" instead of "kind". Let me know what you think.

ba-edb: My type theory brain suggests that "kind" is a better word to use in both places here than "category".

PR #2394, opened 2022-03-01
Release: 2022-03-01

drothery-edb:
## What Changed?

- Search improvements
- Added SLES 12 and 15 install topics for x86 and ppcle for all applicable products
- BigÅnimal:
   - Added a note listing the parameters that can't be decreased
   - Added information about new instance types in preflight script topic


jericson-edb (approved):
Looks good to me.

Issue #2377, opened 2022-02-24
Templates: Ideally code block would not require user modification


drothery-edb:
## Summary

Currently are install instructions require users to edit content in code blocks negating the ability to simply copy and paste into their terminal. We would like a solution that does not require a user to edit content in the code blocks.

## Where did you see the problem?

Here is an example:

```shell
# Install the repository configuration and enter your EDB repository 
# credentials when prompted
zypper addrepo https://zypp.enterprisedb.com/suse/edb-sles.repo

# Install SUSEConnect to register the host with SUSE, allowing access to 
# SUSE repositories
zypper install SUSEConnect

# Register the host with SUSE, allowing access to SUSE repositories
# Replace 'REGISTRATION_CODE' and 'EMAIL' with your SUSE registration 
# information
SUSEConnect -r 'REGISTRATION_CODE' -e 'EMAIL'

# Activate the required SUSE module
SUSEConnect -p PackageHub/15.3/x86_64

# Refresh the metadata
zypper refresh
```

Here is another example for MySQL SLES 12 x86
```
# Use one of the following commands to enable the appropriate 
# subrepository and install the MySQL Foreign data Wrapper:
#
# For MySQL8, enable the MySQL8 repository and disable the MySQL5 
# repository
zypper modifyrepo -d mysql57-community
zypper modifyrepo -e mysql80-community
#
# For MySQL5, enable the MySQL5 repository and disable the MySQL8 
# repository
zypper modifyrepo -d mysql80-community
zypper modifyrepo -e mysql57-community
```

drothery-edb:
Feedback from SMEs is that this in not necessary.

PR #2411, opened 2022-03-03
Demonstrate the use of prompts for variable information

josh-heyer:
Addresses #2377 

Modifies [sles-15.njk](https://github.com/EnterpriseDB/docs/compare/epic/templatized-install-guides...bugfix/josh/2377-prompt-for-registration?expand=1#diff-96adc25f3e09bcaaf5810ad97ee619457cb083bed1f819cea5c29f0dc76e69ef) and [sles-12.njk](https://github.com/EnterpriseDB/docs/compare/epic/templatized-install-guides...bugfix/josh/2377-prompt-for-registration?expand=1#diff-2e9112846f78e459530f2c62022709ba7f52019b42cea1ddffc0da00d6dc83a6) to prompt for SUSE registration info along with an explanatory comment. 

This approach won't work in every situation, but if we wish to allow for users to copy-paste these instructions verbatim then it is an approach that will allow for interactive prompting where necessary.


drothery-edb:
@josh-heyer, I think we can close this now,

PR #2415, opened 2022-03-05
Order PDF sections using the same metadata as the site

### LINKED ISSUE #2318: PDF topic ordering for BDR
## Summary

The order of topics in the BDR 4 PDF is alphabetical and does not reflect the logical order in the online content. The order of the topics in the PDF should match the order in the online content. 

image

This is basically the same issue as https://github.com/EnterpriseDB/docs/issues/1472.

josh-heyer:
## What Changed?

Fixes #2318 

Reads navigation section from frontmatter, orders pages accordingly when combining for PDFs.

Uses python-frontmatter for reliability. Potential future enhancement here for more robust parsing of titles, etc.

Updates tests, dependencies, container and CI builds; minimal code changes to avoid regressions. 

**For review,** before-after output (in text format for ease of comparison): https://drive.google.com/drive/folders/1zFk549CqZQsTz74VoIWR1VhtzT1bNMvE?usp=sharing

Ordering of sections, subsections should match what is on the website *exactly* - with the exception of empty pages, which will continue to be omitted.

Deployment with reordered PDFs: https://622303b6fec4f867e5c4d688--edb-docs-staging.netlify.app/


drothery-edb:
I've checked:

- BigAnimal
- EFM
- BDR

I'll move on to EPAS.

josh-heyer:
I'm comfortable if you're comfortable, @drothery-edb - I expect you've gotten a lot more grief from PDFs than I have 😁

PR #2420, opened 2022-03-07
BigAnimal: Fixed superset links

drothery-edb:
## What Changed?

Fixed the links to the Superset documentation


vishalsawale9 (approved):
I verified each of the links and it's correct now. Thanks!

PR #2427, opened 2022-03-08
BigAnimal: UPM-4772 - AWS Free Trial

drothery-edb:
## What Changed?

Updated the Free trial topic with AWS info


nataliawojcik27 (commented):
Recommend that we still call out they can use EPAS or PostgreSQL as their database type.

Change "AWS GP3 Volume, disk size P1-P4, IOPS and data throughput based on disk size" to something like "AWS GP3 Volume, disk size 1-4. IOPS and data throughput are static during the trial but can be configured with a paid account"

Maybe instead of "different provider" we should specify "different cloud provider"? Just to avoid any confusion

Also think we can remove the 'only' in "Backups are only retained for 30 days. "

nataliawojcik27 (commented):
lgtm! just recommend changing " AWS GP3 Volume, disk size 4 " to "GP3 Volume disk size 4"

Issue #2417, opened 2022-03-05
Feedback on /epas/14/epas_compat_sql/36_create_table.mdx (missing keyword)


CC-Hsu:
Hi, Team,

The last line of `UNLOGGED` parameter misses a keyword compared with [the corresponding PGSQL doc page](https://www.postgresql.org/docs/current/sql-createtable.html#SQL-CREATETABLE-UNLOGGED).

> The unlogged table is automatically truncated after a crash or _unclean_ shutdown.

Best Regards.

nidhibhammar:
Changes are done in PR: https://github.com/EnterpriseDB/docs/pull/2430

PR #2431, opened 2022-03-09
Release: 2022-03-09

drothery-edb:
## What Changed?

EPAS: 

- Addressed DB-1704 edb-modules replaced with edb_wait_states
- Feedback on /epas/14/epas_compat_sql/36_create_table.mdx (missing keyword) #2417


jericson-edb (approved):
Assuming this gets deployed, looks good to me.

PR #2437, opened 2022-03-10
BigAnimal: add default backup retention period

drothery-edb:
## What Changed?

added default info to backup topic


drothery-edb:
Aislinn approved through Slack

PR #2445, opened 2022-03-11
MTK 54: No longer supported - being archived

drothery-edb:
## What Changed?

Removed 54 folder
Added a redirect to 55

(Archived the PDF in the docs-archive repo)


drothery-edb:
@josh-heyer is this the correct redirect? 
`/docs/migration_toolkit/54/*      /docs/migration_toolkit/55/:splat  301`

PR #2447, opened 2022-03-11
Alter BDR, etc. source scripts to sync w/ deletion

josh-heyer:
## What Changed?

BDR, Harp, LiveCompare, pgLogical all need to behave in the same way: newly-added files get brought in, deleted files get removed. 

This was not happening consistently, mostly because these conversion scripts were written to be run manually and sorta expected to be dumping their output in a scratch directory. 

I've altered them to process files in place, and then copy over only select files (mdx, jpg, jpeg, png, svg) - deleting any files of these types that aren't found in the destination subtree. 

(for Harp, only this last sync step is performed)

One small hitch is the images for BDR, which 

1. live in the parent directory from the rest of the content, and
2. don't seem to exist at all in the last sync ref

...so I'm ignoring those for now (using the old sync logic) to avoid deleting them (they're still referenced in the docs). That can be fixed in the future.

### DIFF
--- a/scripts/source/bdr.js
+++ b/scripts/source/bdr.js
@@ -49,7 +49,7 @@ const destination = path.resolve(args[1]);
     file.path = destFilepath;
     try {
       await fs.mkdir(path.dirname(file.path));
-    } catch {}
+    } catch { }
     await write(file);
   };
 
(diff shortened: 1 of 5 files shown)


jericson-edb (approved):
👍

PR #2461, opened 2022-03-16
AWS Backup and Storage updates in BigAnimal

moiznalwalla:
https://enterprisedb.atlassian.net/browse/UPM-5350


moiznalwalla (commented):
@drothery-edb This is from the previous written content. is this accurate? We say above BA has a retention period so am not sure what we mean by this statement?

"Currently BigAnimal does not age out backups, so you can restore clusters to any time since cluster creation."

neel5481 (approved):
LGTM

drothery-edb:
After Neel approved we merged this into our main May release branch

PR #2463, opened 2022-03-16
Epic/adding hierarchy to install

dwicinas:
Using Migration Toolkit, I moved the different Linux platforms into two sub-folders. There is some redundancy of information, which I'm not crazy about, but it does accomplish what Kelly was requesting.


drothery-edb:
Decided on approach demonstrated in PR#2520

PR #2490, opened 2022-03-25
BigAnimal: AWS updates for CLI and API

drothery-edb:
## What Changed?

Added a note about examples working the same for AWS in the use case section

### DIFF
--- a/product_docs/docs/biganimal/release/reference/api.mdx
+++ b/product_docs/docs/biganimal/release/reference/api.mdx
@@ -27,11 +27,12 @@ This call returns the information that either:
 curl https://portal.biganimal.com/api/v1/auth/provider
 ```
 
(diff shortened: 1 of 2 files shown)


zozowell:
@drothery-edb only one thing maybe worth mentioning, 
!!!tip
!!!

drothery-edb:
@nataliawojcik27 , I added the following sentences to the CLI Sample use cases section: 

The examples in this section show Azure as the cloud provider. The functionality is the same when using AWS.

Do you think that is sufficient?

Also, since we call out what is and is not available in free trial topic, I don't think we need to add Zane's suggestion. Do you agree?

cc: @zozowell

drothery-edb:
Closing this PR as the changes have been merged in the main May release branch. Any additional changes should be made on the docs/biganimal/may-release branch.

PR #2497, opened 2022-03-28
Make postinstall optional

jericson-edb:
## What Changed?

- Made the post-install menu and header optional

## Checklist

**Content**


dwicinas (commented):
Most products don't have post-install instructions, so ideally the default behavior does not include a post-install section.  I think I understand from your changes how I would add a post-install section, but I'm less clear on what would be necessary to have no post-install section. Maybe a short discussion would be in order.

install_template/templates/products/postgres-enterprise-manager/base.njk
```diff
@@ -1,6 +1,9 @@
 {% extends "platformBase/" + platformBaseTemplate + '.njk' %}
 {% set packageName %}edb-pem{% endset %}
+{% block postinstallMenu %}-   [Performing the initial configuration](#initial-configuration){% endblock postinstallMenu %}
```
drothery-edb, line 3: Is there a way to pararametize this content so it doesn't need to be entered for the multiple products that have initial configurations? Same question for the heading (row 5). It may be a question of diminishing returns, but it would be nice not to have to enter the boilerplate each time.

jericson-edb:
> Most products don't have post-install instructions, so ideally the default behavior does not include a post-install section. I think I understand from your changes how I would add a post-install section, but I'm less clear on what would be necessary to have no post-install section. 

Makes sense. I updated the PR to show how to make a placeholder block for the post-install section. Since it's empty in the base template, it will be empty in the final render unless one of the specific templates adds content to the block.

The intra-page navigation will need to be copied with this setup, which isn't ideal. There are several possible solutions. I'll play around with it.

> Maybe a short discussion would be in order.

Agreed. @drothery-edb scheduled a meeting next week.

josh-heyer:
⚠Rebased! Fetch, then delete & recreate local branch (or `git pull --rebase`)

product_docs/docs/efm/4/03_installing_efm/failover-manager_4.4_sles-12_ppc64le.mdx
```diff
@@ -0,0 +1,63 @@
...
+Before you begin the installation process, log in as superuser.
+  ```shell
+  # To log in as a superuser:
+  sudo su -
+  ```
+
+
+## Set up the repository
+
+Setting up the repository is a one-time task. If you have already set up your repository, you do not need to perform this step. If you do need to set up the repository, you need to register with EDB. To receive credentials for the EDB repository, visit: [Repository Access Request](https://www.enterprisedb.com/repository-access-request).
```
drothery-edb, line 15:
```suggestion
Setting up the repository is a one-time task. If you have already set up your repository, you do not need to perform this step. If you do need to set up the repository, you first need to register with EDB. To receive credentials for the EDB repository, visit: [Repository Access Request](https://www.enterprisedb.com/repository-access-request).
```

josh-heyer:
⚠rebased! Make sure you update your local branch (delete, fetch, switch or `git pull --rebase`) before continuing work!

dwicinas:
@josh-heyer @jericson-edb We are satisfied with how files are generating for SLES topics for EFM and MTK. Can you review these changes and let us know if we're good to merge?

dwicinas:
We might have another commit to this because we're experimenting with one more re-use change. Hold off on merging until we give the go-ahead.

drothery-edb (approved):

PR #2512, opened 2022-03-31
Adding Migration Handbook to product search

drothery-edb:
## What Changed?

Updated main index file per Josh's instructions to add Migration Handbook to the Search by Product list


drothery-edb:
@josh-heyer I still don't see Migration Handbook showing up in the Search by Product list.

josh-heyer:
It won't show up until reindexed, which it won't do in a preview build @drothery-edb - I'll just merge this (I've verified the frontmatter is correct now)

josh-heyer (approved):

PR #2547, opened 2022-04-13
BigAnimal: clarifying when restores are possible

drothery-edb:
## What Changed?

Updated the first sentence of the Restores section


drothery-edb:
Aislinn approved in Slack

PR #2556, opened 2022-04-14
BigAnimal: AWS regions

drothery-edb:
## What Changed?

replaced AWS free trial regions (not needed because they are documented in the free trial topic) with the AWS regions


drothery-edb:
Merged the regions WIP branch into the May branch.

PR #2562, opened 2022-04-19
Install hierarchy for Hadoop based on architecture

dwicinas:
Restructured position of install files to create hierarchy based on architecture


dwicinas:
@drothery-edb I think this is ready for a review and potential merging.

drothery-edb (approved):

PR #2563, opened 2022-04-19
Revise hierarchy for Mongo install topics to include architecture

dwicinas:
Restructured installation topics so they are presented in a hierarchy that shows architecture.


dwicinas:
@drothery-edb This one should be ready for review as well

drothery-edb (approved):

PR #2569, opened 2022-04-20
Release: 2022-04-20

drothery-edb:
## What Changed?

JDBC Connector 42.3.3.1 release


josh-heyer (approved):
Added a redirect to avoid breaking any links to the old version of JDBC that might be floating around

PR #2584, opened 2022-04-26
Adding architecture hierarchy to OCL Connector

dwicinas:
Adding architecture hierarchy to TOC for installing OCL Connector


dwicinas:
@drothery-edb OCL Connector is also ready for review and merge. 

One thing to note. [This topic](https://www.enterprisedb.com/docs/ocl_connector/latest/04_open_client_library/01_installing_and_configuring_the_ocl_connector/install_on_linux_using_edb_repo/08_ocl_connector14_rhel7_ppcle/) is blank. There's just a heading and no content. I left it that way. Perhaps there was a reason? If it's a mistake, it's easy enough to delete the topic, but references in two index files should also be deleted.

drothery-edb (approved):

PR #2589, opened 2022-04-27
restructure install topics for PostGIS

dwicinas:
Create a hierarchy of install topics to accommodate generated topics.

Hierarchy should includes categories for architecture


dwicinas:
@drothery-edb the Windows images are displaying correctly. I made a couple tweaks to make titles more consistent. Otherwise, I think it's ready for review.

drothery-edb (commented):
I fixed a broken link and noticed that you inadvertently dropped a section from the old Installing landing page. Can you add that back in but let's move it to the top level under release notes and name it Supported platforms. Thanks.

dwicinas:
@drothery-edb I added that section back as I think you were requesting. Please check to confirm.

drothery-edb:
Please move the Supported platforms topics up a level in the hierarchy. See https://www.enterprisedb.com/docs/pem/latest/ for reference.

drothery-edb (approved):

PR #2592, opened 2022-04-28
initial setup for restructuring install topics for pgbouncer

dwicinas:
Create a hierarchy of install topics to accommodate generated topics.

Hierarchy should includes categories for architecture


drothery-edb:
I think this one has the same problem as the other recent restructures. You inadvertently deleted the supported platforms information.

drothery-edb (approved):

PR #2604, opened 2022-05-02
BigAnimal: updated EDB role names for Superset

drothery-edb:
## What Changed?

Updated Owner to Account Owner on Superset topic


drothery-edb:
Approved by Natalia

PR #2609, opened 2022-05-03
Restructure install topics for PgPool-II

dwicinas:
Restructure installation topics to provide a hierarchical TOC, including levels for architecture


dwicinas:
@drothery-edb Even though I was not able to view my changes on a local build, I did look over the Deploy build, and and I also previewed each file. I think this PR is ready for review and merge. However, I did not give it the same level of scrutiny I usually do, so there is some chance problems persist.

drothery-edb:
You inadvertently eliminated the platform support information. Please add as a new Supported platforms topic. Also, we should probably restructure https://www.enterprisedb.com/docs/pgpool/latest/02_extensions/ too.

dwicinas:
Actually, I moved the supported platforms section to a position under Release notes, like I did for PostGIS. I'll move it. Then I'll start work on restructuring the extensions section.

dwicinas:
@drothery-edb This one is ready for review. Lots of changes! I've looked it over but a careful look from another set of eyes would be good.

I removed the instructions for using the graphical installer for Linux installs and Linux extension installs. That was the recommendation for installs, but I extended that recommendation to include extension installations. I retained the instructions for creating extensions. You might disagree with where I positioned that topic.

drothery-edb (approved):

PR #2626, opened 2022-05-06
Release: 2022 05 06

drothery-edb:
## What Changed?

- BigAnimal: added topics for configuring specific IdPs
- PgPool: restructured install topics


jericson-edb (approved):
👍

PR #2632, opened 2022-05-09
BART: removing v2.5 - EOL'ed

drothery-edb:
## What Changed?

Removed version and added redirect


drothery-edb:
merging after testing the redirect

PR #2667, opened 2022-05-13
BigAnimal: private networking restrictions

drothery-edb:
## What Changed?

Added notes indicating that you can't switch to public networking after setting up private networking.


nstott (approved):
looks good to me!
thanks!

PR #2671, opened 2022-05-16
Fix break in list numbering

drothery-edb:
## What Changed?

Initial release of EDB Postgres Distributed


drothery-edb:
Augh. Deleting this branch and closing this PR starting the build over.

PR #2672, opened 2022-05-16
Release: 2022-05-16

drothery-edb:
## What Changed?

Initial release of EDB Postgres Distributed


josh-heyer (dismissed):
should be good once pgd images are in

PR #2674, opened 2022-05-16
PGD: High priority changes for May 17 build

drothery-edb:


product_docs/docs/pgd/4/choosing_durability.mdx
```diff
@@ -0,0 +1,14 @@
+---
+title: "Choosing durability"
+---
+
+EDB Postgres Distributed allows you to choose from several replication configurations based on your consistency, availability, and performance needs:
+
+* Asynchronous
+* Synchronous
```
PJMODOS, line 8: I think we should make it more clear that this is Postgres synchronous, by either calling it "Postgres synchronous" or "Syncrhonous (using `synchronous_standby_names`)", otherwise it will confuse reader in my opinion.

drothery-edb: can we add a link too? got any suggestions?

PR #2679, opened 2022-05-17
Edits to the conflicts topic

ebgitelman:


product_docs/docs/bdr/4/conflicts.mdx
```diff
@@ -651,191 +642,189 @@ node.
...
--   [Column-Level Conflict Detection](column-level-conflicts)
+-   [Origin conflict detection](#origin-conflict-detection) (default)
+-   [Row version conflict detection](#row-version-conflict-detection)
+-   [Column-level conflict detection](column-level-conflicts)
 
-### Origin Conflict Detection
+### Origin conflict detection
 
-(Previously known as Timestamp Conflict Detection, but this was confusing.)
+(Previously known as timestamp conflict detection.)
```
PJMODOS, line 772: I'd just remove this sentence completely, when we are at it (it has been known longer as origin conflict detection than it was known as timestamp conflict detection at this point)

product_docs/docs/bdr/4/conflicts.mdx
```diff
@@ -5,175 +5,172 @@ title: Conflicts
...
+nature of the application and the meaning of the data, so it's important that
 BDR provides the application a range of choices as to how to resolve conflicts.
 
-By default, conflicts are resolved at row level. That is, when changes from two
-nodes conflict, we pick either the local or remote tuple and discard the other
-one. For example, we may compare commit timestamps for the two conflicting
-changes, and keep the newer one. This ensures that all nodes converge to the
-same result, and establishes commit-order-like semantics on the whole cluster.
+By default, conflicts are resolved at row level. When changes from two
+nodes conflict, BDR picks either the local or remote tuple and discards the other
```
PJMODOS, line 17: As Kelly mentioned on slack, I don't think that adding "BDR" is a positive change here, given the BDR, PG Distributed duality. Maybe the passive tone of  "either the local or remote tuple is picked and the other
Is discarded" would be better in these kinds of sentences.

product_docs/docs/bdr/4/conflicts.mdx
```diff
@@ -186,13 +183,13 @@ Updating the Primary Key column is possible, so this SQL succeeds:
...
 
-...but if we have multiple rows in the table, e.g.:
+However, suppose there are multiple rows in the table:
 
 ```sql
 INSERT INTO pktest VALUES (3,3);
 ```
 
-...then some UPDATEs would succeed:
+Some UPDATE operations succeed:
```
PJMODOS, line 192: If we start replacing "UPDATEs" and "INSERTs" and "DELETEs" with "UPDATE operations", we'll need to be more context aware, I am afraid. For example, I "operation" makes sense when we are talking about the replicat (the apply side of the replication), but when we talk about SQL (user input) it's not "UPDATE operation" but "UPDATE statement".

So i.e. the 'INSERT operations that violate multiple UNIQUE constraints" above makes sense, "Where two concurrent `UPDATE` operations on different nodes change the same tuple" also above, I am not quite sure whether it should be "operation" or "statement" (which means likely both are fine) and the "Some UPDATE operations succeed" here should definitely be "statement".

PJMODOS: Then the question is whether we are making it more readable or more confusing, of course :)

product_docs/docs/bdr/4/conflicts.mdx
```diff
@@ -1013,86 +1000,86 @@ bdr.alter_node_set_conflict_resolver(node_name text,
...
+    re-created table before it sees the DDL to re-create the table. It then
+    incorrectly skips the remote data, assuming that the table is recently dropped,
+    and causes data loss. We hence recommend that you don't reuse the object namesq
     immediately after they are dropped along with this conflict resolver.
--   `skip_transaction` - skips the whole transaction that has generated the
+-   `skip_transaction` &mdash; Skips the whole transaction that generated the
     conflict. Can be used for `apply_error_ddl` conflict.
--   `update_if_newer` - update if the remote row was committed later (as
+-   `update_if_newer` &mdash; Update if the remote row was committed later (as
     determined by the wall clock of the originating server) than the conflicting
```
PJMODOS, line 1048: should we unify nomenclature between server/host/node to always be refered to as node?

product_docs/docs/bdr/4/conflicts.mdx
```diff
@@ -651,191 +642,189 @@ node.
...
 
-Users are advised to not leave down nodes for extended outages,
-as discussed in [Node Restart and Down Node Recovery](nodes).
+We suggest that you don't leave down nodes for extended outages,
+as discussed in [Node restart and down node recovery](nodes).
 
-On EDB Postgres Extended Server and EDB Postgres Advanced Server, BDR will
-automatically hold back the freezing of rows while a node is down to handle
-this situation gracefully without the need for changing parameter settings.
+On EDB Postgres Extended Server and EDB Postgres Advanced Server, BDR
```
PJMODOS, line 803: Obviously, here it would still stay as server as it's the product name, not a node reference.

product_docs/docs/bdr/4/conflicts.mdx
drothery-edb: I'll revert the changes. Better to avoid the possibility of confusion if the rules about when to use statement vs operation are too hard to figure out.

drothery-edb:
I changed host to node but wasn't so sure about the instances of server so I left them as is.
I changed PostgreSQL to Postgres in all instances except for URLs. Petr, you might want to check my changes.

product_docs/docs/bdr/4/conflicts.mdx
drothery-edb: reverted to Kelly's rewrite (with just a tiny bit of editing)

product_docs/docs/bdr/4/conflicts.mdx
drothery-edb: I think so. I saw your comment below and assume all instances of server can be changed to node except for in product names. I'll go ahead and make that change.

PJMODOS: yep thanks

PJMODOS (approved):
LGTM

PR #2683, opened 2022-05-18
PGD CLI - restructure prototype

drothery-edb:
## What Changed?

Moved content out of the main index file and created an Installing and Using topic


PJMODOS (approved):
Looks reasonable, I think we should add the CLI to the compatibility matrix btw :)

PR #2688, opened 2022-05-18
update to autostart instructions

EFM-Bobby:
The text made it sound like the user checks to see if files exist, rather than describing that the efm agent does it. Also removed a redundant sentence that I think isn't needed.

## Checklist

**Content**

### DIFF
--- a/product_docs/docs/efm/4/01_prerequisites.mdx
+++ b/product_docs/docs/efm/4/01_prerequisites.mdx
@@ -62,7 +62,7 @@ Where `x` specifies the Postgres version.
 
 If a primary node reboots, Failover Manager might detect the database is down on the primary node and promote a standby node to the role of primary. If this happens, the Failover Manager agent on the rebooted primary node doesn't get a chance to write the `recovery.conf` file, and the `recovery.conf` file prevents the database server from starting. If this happens, the rebooted primary node returns to the cluster as a second primary node.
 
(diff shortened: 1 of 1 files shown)


product_docs/docs/efm/4/01_prerequisites.mdx
```diff
@@ -62,7 +62,7 @@ Where `x` specifies the Postgres version.
 
 If a primary node reboots, Failover Manager might detect the database is down on the primary node and promote a standby node to the role of primary. If this happens, the Failover Manager agent on the rebooted primary node doesn't get a chance to write the `recovery.conf` file, and the `recovery.conf` file prevents the database server from starting. If this happens, the rebooted primary node returns to the cluster as a second primary node.
 
-To prevent this condition, ensure that the Failover Manager agent auto starts before the database server. The agent starts in idle mode. Check to see if there is already a primary in the cluster. If there is a primary node, the agent verifies that a `recovery.conf` or `standby.signal` file exists, and the database doesn't start as a second primary.
+To prevent this condition, ensure that the Failover Manager agent auto starts before the database server. The agent starts in idle mode and checks to see if there is already a primary in the cluster. If there is a primary node, the agent verifies that a `recovery.conf` or `standby.signal` file exists. If neither file exits, the agent will create the `recovery.conf` file.
```
drothery-edb, line 65:
```suggestion
To prevent this condition, ensure that the Failover Manager agent auto starts before the database server. The agent starts in idle mode and checks to see if there is already a primary in the cluster. If there is a primary node, the agent verifies that a `recovery.conf` or `standby.signal` file exists. If neither file exits, the agent creates the `recovery.conf` file.
```

drothery-edb (approved):

PR #2705, opened 2022-05-21
PGD: moving CLI content to main content prototype

drothery-edb:
## What Changed?

Trying out having the CLI as part of PGD


product_docs/docs/pgd/4/deployments/installing_cli.mdx
```diff
@@ -1,27 +1,17 @@
 ---
-title: "Installing the EDB Postgres Distributed Command Line Interface"
-navTitle: "Installing the PGD CLI"
+title: "Installing a standalone instance of PGD CLI"
+navTitle: "Installing PGD CLI"
 ---
 
 
-## Installing with TPAexec
```
smaher-edb, line 7: This is necessary. PGD CLI is totally different than PGD including everything packaging, configuration etc.

smaher-edb (commented):
Please note this will need thorough review if it is being merged. Just to summarize, PGD CLI is totally different thing than PGD (which is nothing but the another name given to PG flavors+BDR plugin). Installation, configuration and packages are all different for PGD CLI. All the original changes including installation guide (TPA + independent ), configuration guide, release notes has to be kept separate becuase they are different for PGD CLI.

product_docs/docs/pgd/4/deployments/installing_cli.mdx
```diff
@@ -1,27 +1,17 @@
...
-
-If you are using TPAexec (v22.14 or later) to deploy EDB Postgres Distributed, it installs the PGD CLI package (`edb-pgd-cli`) on each BDR node, by default. It also creates the configuration file (`/etc/edb/pgd-config.yml`) prepopulated with the endpoints of each node in the cluster so no additional configuration is necessary.
-
-If you wish to install the PGD CLI on any non-BDR instance in the cluster, you simply attach the pgdcli role to that instance in TPAexec's configuration file before deploying. See [TPAexec](/pgd/latest/deployments/tpaexec) for more information.
-
-
```
smaher-edb, line 13: All this information is must and should also be kept together.

smaher-edb:
I see this PR is still under development. But looking at folder structure, I think keeping PGD and PGD CLI installation together has no value. It will create more confusion. PGD CLI can certainly be in same PDF as PGD but it will make sense to keep all the CLI sections one below the another - Installation, Requirement, Configuration etc. All these are related section to one another and will only make sense when they are one below the another.

product_docs/docs/pgd_cli/1/requirements.mdx
```diff
@@ -1,11 +0,0 @@
----
-title: Requirements
```
smaher-edb, line 2: Information present here is necessary and should be near the Configuration steps.

smaher-edb:
New changes looks better. As you mentioned release notes are missing as they are still under discussion. 
I also assume these changes are not final and things will change as we have more clarity (like CLI still has lots of BDR references in auto-generated files)

smaher-edb (approved):

PR #2723, opened 2022-05-27
repmgr: add stub for 5.3.2 release

ibarwick:
Currently this page: https://www.enterprisedb.com/docs/repmgr/latest/ shows repmgr on version 5.2.1, which is outdated; I am guessing this is the correct way to update it?

### DIFF
--- a/product_docs/docs/repmgr/5.3.2/index.mdx
+++ b/product_docs/docs/repmgr/5.3.2/index.mdx
@@ -1,0 +1,22 @@
+---
+navTitle: repmgr
+title: "Replication Manager (repmgr)"
+directoryDefaults:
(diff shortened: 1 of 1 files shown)


ibarwick:
@drothery-edb per note in #2728

drothery-edb:
@ibarwick It depends if you want a 5.3.1 version of the doc and a 5.3.2. You can see in the test deployment that is what you did - added another version of the doc. 

You could just chance the version number by renaming the 5.3.1 folder to 5.3.2. 

Another option is to remove the versioning all together. See the stub topic for PostreSQL for an example of that approach: https://deploy-preview-2723--edb-docs-staging.netlify.app/docs/supported-open-source/postgresql/. That is what I would recommend considering the content isn't likely to change between versions so having versioned content doesn't add any value and requires maintenance. If you want to take that approach I can set that up for you.

drothery-edb:
See https://github.com/EnterpriseDB/docs/pull/3375

PR #2730, opened 2022-05-31
Remove prereqs form PEM

jericson-edb:
## What Changed?

* Reorganized the `prerequisites` block so I can remove it from the PEM documentation
* Add PEM to template processing so that I can see results
* Actually remove the `prerequisites` block from PEM
* 
The result of this change on postgres-enterprise-manager_8_sles-15_ppc64le.mdx is:

~~~
---
navTitle: SLES 15 on ppc64le
title: Installing Postgres Enterprise Manager on SLES 15 ppc64le
---

Before you begin the installation process, log in as superuser.

```shell
# To log in as a superuser:
sudo su -
```

<!-- Intentionally blank -->

# Install the following repository for PEM dependencies

```shell
zypper addrepo https://download.opensuse.org/repositories/Apache:/Modules/SLE_15_SP3/Apache:Modules.repo
```

## Install the package

```shell
zypper -n install edb-pem
```

## Initial configuration

```shell
# You can configure the PEM server using the following command:
/usr/edb/pem/bin/configure-pem-server.sh
```

For more details, see [Configuring the PEM Server on Linux](https://www.enterprisedb.com/docs/pem/latest/pem_inst_guide_linux/04_installing_postgres_enterprise_manager/05_configuring_the_pem_server_on_linux/).
~~~


josh-heyer:
Note to self: this is kind of a dangerous option, it should be explicit on product+platform combos that need it, not implicit as it is today!

PR #2737, opened 2022-06-02
Replace BDR refs from PGD CLI docs

smaher-edb:
## What Changed?

Changed
- "BDR cluster" to "EDB Postgres Distributed cluster" and 
- "EDB Postgres Extended" to "EDB Postgres Extended Server.”

## Checklist

**Content**

### DIFF
--- a/product_docs/docs/pgd_cli/1/pgd_show-nodes.mdx
+++ b/product_docs/docs/pgd_cli/1/pgd_show-nodes.mdx
@@ -6,8 +6,8 @@ Shows all nodes in the EDB Postgres Distributed cluster and their summary.
 
 ### Synopsis
 
-Shows all nodes in the EDB Postgres Distributed cluster and their summary such
-as name, node id, group, and current/target state etc.
+Shows all nodes in the EDB Postgres Distributed cluster and their summary
+including name, node id, group, and current/target state.
 
 Node States
 


smaher-edb:
It looks like at least in `pgd_show-nodes.mdx` files "Node States" section, we should use "PGD cluster" instead of "EDB Postgres Distributed cluster".

product_docs/docs/pgd_cli/1/pgd_show-nodes.mdx
```diff
@@ -2,21 +2,21 @@
...
 ---
 
-Shows all nodes in the BDR cluster and their summary.
+Shows all nodes in the EDB Postgres Distributed cluster and their summary.
 
 ### Synopsis
 
-Shows all nodes in the BDR cluster and their summary such as name, node id,
-group, and current/target state etc.
+Shows all nodes in the EDB Postgres Distributed cluster and their summary such
```
drothery-edb, line 9: We try to avoid Latin abbreviations. Can you change this in the source to: Shows all nodes in the EDB Postgres Distributed cluster and their summary including name, node id, group, and current/target state.

smaher-edb:
@PJMODOS @drothery-edb Just want to point out one important thing here - as asked in the email, I've made the changes only at places where "BDR cluster" words appeared. And not the individual word "BDR". 

For example, description for camo, clock skew and other commands still has word BDR in it. Please let me know if this needs to be changed as well. Although for sure we can't replace it with "EDB Postgres Distributed". It has to be `PGD CAMO` or `PGD Raft` like that though I think BDR looks more correct here. 

`"Shows BDR CAMO (Commit at Most Once) details."`
`"Shows the status of clock skew between each BDR node pair."`
`"Shows BDR Raft (consensus protocol) details."`
`"Shows BDR subscription (incoming replication) details."`
`"Shows the version of BDR and Postgres installed on each node."`
`* PROMOTING: Promotion from logical standby to full BDR node is in progress.`
`* ACTIVE: The node is a full BDR node and is currently ACTIVE. This is the
most common node status.`

smaher-edb:
Also, output of some of the commands will still have BDR word because underlying database views return that value e.g. message for Replslots

```
  $ pgd check-health

  Check      Status   Message
  -----      ------   -------
  CAMO       Critical At least 1 CAMO partner is not connected
  ClockSkew  Critical Clockskew cannot be determined for at least 1 BDR node pair
  Connection Critical The node bdr-b1 is not accessible
  Raft       Warning  There is no RAFT_LEADER, an election might be in progress
  Replslots  Critical There is at least 1 BDR replication slot which is inactive
  Version    Warning  There is at least 1 node that is not accessible
```

drothery-edb:
I'd lean to keeping BDR in the other places where BDR is used, but I think the definitive answer would come from @kelpoole and @PJMODOS.

drothery-edb:
I'm closing this PR because I have moved the changes to https://github.com/EnterpriseDB/docs/pull/2705, which moves the CLI content into the main PGD content per Kelly's request.

smaher-edb:
I have updated my review comments in - https://github.com/EnterpriseDB/docs/pull/2705/. Please note we should have all the information currently present in CLI docs to the new PR. And they must also be kept together. Except the name "PGD" there is nothing common in PGD and PGD CLI including the code, language, packages, installation steps, configuration, release notes etc.
This is very very important that we keep all the CLI info at one place. It can be under PGD but completely in separate section. Also, I don't think we should remove any content from existing docs now. All the information there is necessary in one way or another.

PR #2741, opened 2022-06-02
eprs user guide: description fixes

ccestes:
Made changes to instructions according to feedback from XDB-1097


anwar-sahib (approved):
All the required changes added

anwar-sahib (approved):

PR #2748, opened 2022-06-03
PGD: moved durability content from BDR

drothery-edb:
## What Changed?

Appended durability content from BDR to the Choosing durability topic


drothery-edb:
Moved change to #3065

PR #2755, opened 2022-06-07
Added info

ccestes:
Added additional information as per [XDB-961](https://enterprisedb.atlassian.net/browse/XDB-961)


anwar-sahib (approved):
Required text added.

PR #2764, opened 2022-06-09
Install templates: SLES topics for completed product set

drothery-edb:
Generated and deployed SLES topics for all products except for Replication Server, PEM, and EPAS


drothery-edb:
Experimenting with new script - works like a charm

PR #2765, opened 2022-06-09
Templatize install instructions for Replication Server

dwicinas:
Templatizing Linux install instructions for Replication Server

### DIFF
--- a/install_template/config.yaml
+++ b/install_template/config.yaml
@@ -1,568 +1,606 @@
 products:
-  - name: EDB JDBC Connector
+#   - name: EDB JDBC Connector
+#     platforms:
+#       - name: CentOS 7
+#         arch: x86_64
+#         supported versions: [42.x.x]
+#       - name: CentOS 7
+#         arch: ppc64le
+#         supported versions: [42.x.x]
+#       - name: AlmaLinux 8 or Rocky Linux 8
+#         arch: x86_64
+#         supported versions: [42.x.x]
+#       - name: RHEL 7 or OL 7
+#         arch: x86_64
+#         supported versions: [42.x.x]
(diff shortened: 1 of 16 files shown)


dwicinas:
@drothery-edb This work is ready to be merged.

install_template/config.yaml
```diff
@@ -575,3 +575,41 @@ products:
...
+#         arch: ppc64le
+#         supported versions: [7, 8]
+#       - name: RHEL 8 or OL 8
+#         arch: x86_64
+#         supported versions: [7, 8]
+#       - name: Ubuntu 18.04 or Debian 9
+#         arch: x86_64
+#         supported versions: [7, 8]
+#       - name: Ubuntu 20.04 or Debian 10
+#         arch: x86_64
```
josh-heyer, line 602: Let's try & leave everything uncommented in this file going forward - we can restrict what is deployed via rules in deploy.mjs

install_template/config.yaml
```diff
@@ -575,3 +575,41 @@ products:
...
+#         supported versions: [7, 8]
+      - name: SLES 12
+        arch: x86_64
+        supported versions: [8]
```
josh-heyer, line 606: No current v8 for eprs - looks like 6.2 and 7: https://www.enterprisedb.com/docs/eprs/latest/

josh-heyer (changes requested):

install_template/config.yaml
dwicinas: Fixed in latest commit

josh-heyer (dismissed):

drothery-edb (approved):

PR #2768, opened 2022-06-09
Replacing references to BDR with EDB Postgres Distributed (EPD) - index page

JanKarremans:


drothery-edb:
The Doc team will be applying global terminology changes once final decisions have been made.

PR #2769, opened 2022-06-09
Update epas_deb9_x86.mdx

BillSmith-EDB:
Needed to install gnupg2 before I could successfully install the EDB signing key.  Updated docs to include the command for installing gnupg2.

## What Changed?

- Modified docs to include step for installing gnupg2 which is needed to successfully import the EDB signing key.

## Checklist

**Content**


drothery-edb:
Closing as the install topics are generated so if we merged this PR it would get overwritten next time we generated the topics. You can create a GitHub issue or a Jira issue in the Documentation Feedback project if you feel this change is necessary.

PR #2772, opened 2022-06-10
Replacing references to BDR with EDB Postgres Distributed (EPD) - overview page

JanKarremans:


JanKarremans:
In this section, "Quorum" is also discussed. Should this be changed / adapted to reference "Group Commit" instead?

drothery-edb:
Closing since the doc team will be making a terminology pass on the docs before 5.0. Noted that instances of quorum may need to change.

PR #2773, opened 2022-06-10
Replacing references to BDR with EDB Postgres Distributed (EPD) - installation page

JanKarremans:


drothery-edb:
The doc team will be making a global pass with the BDR > EDB Postgres Distributed name changes.

PR #2779, opened 2022-06-10
Replacing references to BDR with EDB Postgres Distributed (EPD) - harp2.1.0_rel_notes page

JanKarremans:
2.1 is the first version paired to EPD


drothery-edb:
The doc team will be making a global pass with the BDR > EDB Postgres Distributed name changes.

PR #2784, opened 2022-06-13
Templatize Debian Ubuntu for EFM

dwicinas:
Working on template changes for debian and Ubuntu for EFM


dwicinas:
@drothery-edb, @josh-heyer Here are the changes for templatizing Debian/Ubuntu for EFM. Note that Debian and Ubuntu are broken into separate platforms, four total.

install_template/templates/platformBase/ubuntu-20.04.njk
```diff
@@ -3,13 +3,13 @@
 {% block prerequisites %}
```
josh-heyer, line 3: Let's go with platformBase/ubuntu.njk or platformBase/ubuntu-18-plus.njk or similar and then have both ubuntu templates derive from it - 20 will likely not be the last version of Ubuntu we support and deriving 18 from 20 is nonintuitive.

install_template/config.yaml
```diff
@@ -351,28 +351,34 @@ products:
...
         arch: x86_64
-        supported versions: [3.10, 4.0, 4.1, 4.2, 4.3, 4.4]
+        supported versions: [4.4]
       - name: RHEL 7 or OL 7
         arch: ppc64le
-        supported versions: [3.10, 4.0, 4.1, 4.2, 4.3, 4.4]
+        supported versions: [4.4]
       - name: RHEL 8 or OL 8
         arch: x86_64
-        supported versions: [3.10, 4.0, 4.1, 4.2, 4.3, 4.4]
```
josh-heyer, line 369: Let's try to keep all supported versions here - we can eliminate those we don't want via the deploy mappings.

josh-heyer (changes requested):
Couple of minor notes inline

install_template/config.yaml
dwicinas: okay, but let me say that it's somewhat painful dealing with the HUNDREDS of files generated when you run a build. I have to visually sift through the results to find the files I'm concerned with. More than once I've realized I'm looking at files that were generated for another version or sometimes even another product! What is the reasoning for keeping those older releases in the config file?

install_template/templates/platformBase/ubuntu-20.04.njk
dwicinas: Just to be clear, you're suggesting I create a file called platformBase/ubuntu.njk, and use that as the basis for different versions ofUbuntu, such as ubuntu-18.04.njk and ubuntu-20.04.njk. Correct? 

Should this approach be used across the board? Currently, for some SLES products, SLES 15 is based on SLES 12. That seems reasonable to me. I agree that basing Ubuntu 18 off of Ubuntu 20 is non-intuitive.

josh-heyer (dismissed):
Looks solid, with one quibble (I counted 5 occurrences of these absolute paths)

josh-heyer (dismissed):

drothery-edb (approved):

PR #2786, opened 2022-06-13
connectors: ubuntu and debian install restructure

ccestes:
Installation topics, Ubuntu and Debian, restructured as separate topics


dwicinas (approved):
I fixed a spacing issue in an index file TOC. This probably had nothing to do with Chris's work. After making that change, I approve. @ccestes or @drothery-edb can merge the changes.

PR #2799, opened 2022-06-14
Add livecompare, pglogical, edbplus; remove ark; alphabetize

josh-heyer:
## What Changed?

This is motivated by #2798 - need to get product names in some semblance of order for contextual search to work properly. 

- Removes "ark" definition - this product isn't built anymore (and hasn't been in a good while)
- Adds definitions for LiveCompare, pglogical, edb_plus

### DIFF
--- a/src/constants/products.js
+++ b/src/constants/products.js
@@ -1,26 +1,27 @@
 import IconNames from "../components/icon/iconNames";
 
 export const products = {
-  epas: { name: "EDB Postgres Advanced Server", iconName: IconNames.EDB_EPAS },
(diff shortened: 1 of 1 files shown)


drothery-edb (commented):
The last time I checked with Kelly on the correct product names, EPRS did not include "EDB" and most of the community-based products did. See https://enterprisedb.atlassian.net/wiki/spaces/DCBC/pages/2883584001/Product+names. I would check with her before making any changes though. Product name decisions are often quite fluid.

josh-heyer:
Punting on reworking existing names for now; we can take another pass on this later (maybe after we're importing product registry data)

PR #2800, opened 2022-06-15
remove billing:update permission from the BigAnimal protal_access doc

xugy99:
## What Changed?

- fix incorrect permission of "owner" role against billing in BigAnimal product

## Checklist

**Content**


xugy99:
@drothery-edb can you review this PR? 

I happen to find this doesn't match what we set in the BigAnimal service: BA doesn't have "update billing" action and permission at all.

PR #2802, opened 2022-06-15
Generated templatized SLES topics

dwicinas:
Deployed all SLES topics for every product except PEM and EPAS


drothery-edb:
@dwicinas the package name for Hadoop should be edb-as14-hdfs_fdw not edb-as2-hdfs_fdw, I believe

drothery-edb (approved):

PR #2803, opened 2022-06-15
correct the template config file versions & deploy mappings for mongo, mysql, postgis

josh-heyer:
More incorrect versions recorded in this config file (and I missed this when creating deploy mappings!)


drothery-edb (approved):
We made the same changes to the deploy.mjs file in our branch but missed the changes needed in the config.yaml file.

PR #2809, opened 2022-06-16
chore: adding chapter formatting to the opening page for EDB Postgres for Kubernetes

JanKarremans:
After discussing with @josh-heyer, we figured out why the tiles on the opening page were missing and how to add those.
This PR is to fix that.


josh-heyer:
FYI, @drothery-edb - deploy won't work on this PR, as it depends on repository secrets that won't be passed when run on a forked branch.

drothery-edb (approved):

PR #2826, opened 2022-06-20
EPAS 14 - Install SLES topics

nidhibhammar:
EPAS 14 SLES topics are successfully deployed after changing the names of the folders to match the mapping file naming convention. The names of the deployed files are starting with integer (05_ , 06, ...) hence the old files are also present in this PR.  We can discuss how to handle this in this PR and make the future strategy for other topics


install_template/templates/products/edb-postgres-advanced-server/sles-12.njk
```diff
@@ -1,6 +1,11 @@
...
-{{ super() }}
-{% endblock postinstall %}
+{% block individual_component %}
+To install an individual component:
+
+```shell
+zypper install <package_name>
+```
+
+Where `package_name` can be any of the available package from the [available package list](https://www.enterprisedb.com/docs/epas/latest/epas_inst_linux/install_details/rpm_packages/).
```
josh-heyer, line 10: Try to use relative paths for links within docs, e.g. `/epas/latest/epas_inst_linux/install_details/rpm_packages/`

josh-heyer (approved):

PR #2831, opened 2022-06-21
PEM 8 - Install SLES topics

nidhibhammar:


drothery-edb:
Comments: 

- Move Using the EPD repo topic before the Local repo topic in the nav tree
- Use the same linking convention for the EDB repo topic as we use for all the other products, see https://www.enterprisedb.com/docs/epas/latest/epas_inst_linux/installing_epas_using_edb_repository/
- You have a Prerequisites section in the current published content. Did you intentionally drop it in the generated topics? If so, why? If not, you should add it to the template.
- SLES 15 for Power should be above SLES 12 in the nav tree

drothery-edb:
@nidhibhammar I wasn't suggesting changing the file names, but changing the content format or nav order. Changing file names causes havoc.

drothery-edb (approved):

PR #2856, opened 2022-06-27
PEM 8.5 - updates

nidhibhammar:
Added conformance report links and edited the Non-root user doc for the PEM agent.

### DIFF
--- a/product_docs/docs/pem/8/registering_agent.mdx
+++ b/product_docs/docs/pem/8/registering_agent.mdx
@@ -149,33 +149,33 @@ To use a non-root user account to register a PEM agent, you must first install t
 
     Your PEM agent is now registered and started with `edb` user. If your machine restarts then this agent does not restart automatically and you need to start it manually using the previous command.
 
-    Optionally, you can create the service for this PEM agent as root user to start this agent automatically at machine restart as follows:
+5.  Optionally, you can create the service for this PEM agent as root user to start this agent automatically at machine restart as follows:
 
-5.  Update the values for the configuration file path and the user in the `pemagent` service file as superuser:
+    a.  Update the values for the configuration file path and the user in the `pemagent` service file as superuser:
     
-    ```ini
-    $ sudo vi  /usr/lib/systemd/system/pemagent.service
-    User=edb
-    Group=edb
-    WorkingDirectory=/home/edb/pem/agent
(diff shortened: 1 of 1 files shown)


product_docs/docs/pem/8/registering_agent.mdx
```diff
@@ -97,43 +97,84 @@ For information about using the pemworker utility to register a server, see [Reg
...
+    $ chmod a+x /home/edb/pem/run_pemagent.sh
+    $ cd /home/edb/pem
+    $ ./run_pemagent.sh
+    ```
+
+    Your PEM agent is registered and started with `edb` user at this point. If the machine restarts then this agent do not restart automatically and need to start manually using the above command.
+
+    Optionally, you can create the service for this PEM agent as root user to start this agent automatically at machine restart as follows:
+
+5.  Update the values for the configuration file path and the user in the `pemagent` service file as superuser:
```
drothery-edb, line 153: Steps 5 and 6 are only necessary if you choose to set up a service to restart the agent automatically, correct? If this is true we should nest these steps making them step 4a and 4b.

product_docs/docs/pem/8/pem_rel_notes/index.mdx
```diff
@@ -6,11 +6,11 @@ The Postgres Enterprise Manager (PEM) documentation describes the latest version
...
-| [8.5.0](03_850_rel_notes) | 23 Jun 2022  | NA                                                                                                                                                                                                                                                                                                                                                     | NA                                                                                                  |
+| [8.5.0](03_850_rel_notes) | 23 Jun 2022  | NA                                                                                                                                                                                                                                                                                                                                                     | [Conformance Report](https://www.enterprisedb.com/accessibility)                                    |
```
drothery-edb, line 9: I suggest we link to the specific report except for the link for 8.1.1 that doesn't have a specific report (I'm not sure why we link at all in that case though). The column header can be a link to https://www.enterprisedb.com/accessibility.

nidhibhammar: I was linking it to a specific report previously but then changed it to this page, as this report gets published along with packages so we have to wait until it gets published. Still, if you think links to actual reports make more sense I can do that. Let me know

drothery-edb (approved):

PR #2857, opened 2022-06-27
Changes to implement Ubunut and Debian install instructions

dwicinas:
Changes should include:

Hadoop
MongoDB
MySQL


dwicinas:
@drothery-edb I think this PR could be merged. It will make the Debian/Ubuntu changes for Hadoop and Mongo. It also does some initial setup for MySQl. Implementing MySQL, however, will be complicated.

PR #2861, opened 2022-06-27
OCL: code block updates

ccestes:
Updated code blocks (mostly to C) and excluded install topics


josh-heyer (approved):
one small correction, looks good otherwise

josh-heyer (approved):

PR #2874, opened 2022-06-30
PGD organization scratchpad

josh-heyer:
## Introduction

PGD is a recent creation, an attempt to create an overarching product that exists on top of / as a superset of the existing components BDR, TPA, Harp, LiveCompare

As such, the PGD documentation should be considered the entrypoint for folks looking to learn about EDB's ecosystem of HA products and designs, identify their collective requirements and limitations, prepare for evaluation or plan a strategy for deployment. Individual product docs should be focused on reference material for their individual components or services, e.g. documenting BDR built-in roles, harpctl commands, LiveCompare configuration, etc.
   
Current docs are split in roughly this manner, but some artifacts remain (for historical reasons or through oversight). A few representative examples: 
   
- HARP 2.0 docs contain an installation topic that partially duplicates information found under PGD, largely for the benefit of those using versions of PGD/BDR prior to v4. 
- The PGD "Other considerations" topic and the BDR "Architectural Overview" topic both contain roughly the same note on clocks and timezones
- The PGD section on Durability largely defers to the BDR topic "Durability and performance options", presenting little more than a link and a bullet list (that does not readily map to the linked topic).
- The BDR "Application use" topic should probably be introduced as part of PGD, as it is less specific to the mechanics of the BDR extension than with the application changes and testing that may be needed in order for PGD to be used.
- etc.

This PR is currently sort of a scratchpad for investigating possible remediation for issues such as these. 
   
## Research (ongoing)

I've [started a list of PGD/BDR/HARP/LiveCompare topics currently in docs](https://docs.google.com/spreadsheets/d/15mX3glnCjMTai7HEX67GgdTkRNssqLKxnbxmYnIM5qM/edit#gid=0), with notes on what I observed while reading them:

- **type** is how the topic appears to present itself (in format, title, etc.). E.g. "interface doc" appers to describe a CLI or API, "reference" covers complete set (or subset) of commands, configuration options, etc., "procedure detail" describes the steps necessary to accomplish some task (or tasks), etc.

- **Purpose** is my impression of the topic's reason for existing (which is... not necessarily how it was intended or how the information *should be* presented, but gotta start somewhere)

- **Categories** are a collection of keywords or phrases that pertain to the topic (think: tags.) I'm mostly just using this to quickly find stuff again.

I've also perused the knowledge base articles pertaining to BDR, HARP and LiveCompare. This is mostly for gaining background info on the evolution of these products; still have a lot more reading to do there.

## Current recommendations reflected by this PR

Deployment options:

   - Rename to "Deploying" - make this into an action for symmetry with "Upgrading" 
   - Link to TPAexec immediately (remove sentance implying multiple options as only one is currently available)

TPAexec:
   
   - Rename to "Deploying PGD with TPAexec" (nav: "Deploying with TPAexec")
   - Link to "Choosing your architecture" topic from the text "Trusted Postgres Architecture (TPA)" - TODO: are we still using this terminology going forward? Would be beneficial to define it somewhere
   - Link to "Installing TPAexec" topic immediately - leave explanatory sections for later by moving entire "Getting started" section to top of topic after intro

Installing TPAexec: 
   
   - Move the repo configuration from "Installing TPAexec" section to "Prerequesites" - it is a prerequesite, and a good introduction to how these repos are configured
   - Include note for Debian/Ubuntu repository
   - Describe how to find this script (with already-embedded token) via the customer portal
   - Rename subsection "Installing TPAexec" to "Installing TPAexec using a package manager"
      
Using TPAexec:
   
   - Move to last in the section
   
   (This is a reference topic)
   
Example: Deploying EDB Postgres Distributed
   
   - Add `--harp-consensus-protocol bdr` option to the configure call, and make "Silver" lowercase for layout
   - TODO: consider switching to docker for the example 
   - TODO: include detailed instructions / example for connecting via psql vs just "You can"
   
TODO: CLI install - rename, put last, crosslink with TPA
   
TODO: HARP install - immediately link to TPA, remove "can't access" note, make clear that rest of instructions are for legacy / unusual situations
   
TODO: software requirements need to be pushed up

### DIFF
--- a/product_docs/docs/pgd/4/deployments/index.mdx
+++ b/product_docs/docs/pgd/4/deployments/index.mdx
@@ -1,12 +1,14 @@
 ---
-title: "Deployment options"
+title: "Deploying EDB Postgres Distributed"
+navTitle: "Deploying"
 indexCards: simple
 
 ---
 
-You can deploy and install EDB Postgres Distributed products using the following methods:
+Current:
 
-- TPAexec is an orchestration tool that uses Ansible to build Postgres clusters as specified by TPA (Trusted Postgres Architecture), a set of reference architectures that document how to set up and operate Postgres in various scenarios. TPA represents the best practices followed by EDB, and its recommendations are as applicable to quick testbed setups as to production environments.
+- [Deploy PGD using TPAexec](tpaexec).
+  TPAexec is an orchestration tool that uses Ansible to build Postgres clusters as specified by TPA (Trusted Postgres Architecture), a set of reference architectures that document how to set up and operate Postgres in various scenarios. TPA represents the best practices followed by EDB, and its recommendations are as applicable to quick testbed setups as to production environments.
 
 Coming soon:
 
--- a/product_docs/docs/pgd/4/deployments/tpaexec/index.mdx
+++ b/product_docs/docs/pgd/4/deployments/tpaexec/index.mdx
@@ -1,19 +1,25 @@
 ---
-title: TPAexec
+title: "Deploying EDB Postgres Distributed with TPAexec"
+navTitle: "Deploying with TPAexec"
 navigation:
 - installing_tpaexec
-- using_tpaexec
 - quick_start
+- using_tpaexec
 ---
 
 The standard way of deploying EDB Postgres Distributed in a self managed setting,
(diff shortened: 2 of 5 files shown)


djw-m:
It's work is done, it will now walk into the sunset.

PR #2875, opened 2022-06-30
PEM agent - SLES topics

nidhibhammar:


drothery-edb (dismissed):
- Agent should be lowercase in the title of the install topics
- Looks like we are moving to the convention of using PEM as the modifier when talking about the server or agent. Use PEM in the install topic titles to adhere to that convention.

PR #2892, opened 2022-07-05
BigAnimal: GUC formulas - terminology improvements

drothery-edb:
## What Changed?

Using "kind" instead of "category" when talking about classes of units


drothery-edb:
I'm going to merge this. I think you both would agree with this change.

PR #2893, opened 2022-07-05
Templates: code block language

drothery-edb:
## What Changed?

Trying to use the correct language for these code blocks


nidhibhammar:
This content is removed from edb*plus base file hence these changes are not needed. Closing this PR.

PR #2909, opened 2022-07-07
PGD: consolidated release notes prototype

drothery-edb:
## What Changed?

Trying out consolidated release notes for PGD


drothery-edb:
@PJMODOS I DM'ed you about the new Supported BDR upgrade paths topic that needs your review about a week ago.  I'm trying to clean up all the outstanding BDR PRs so that we can can remove the source files from the old location. In this PR in addition to consolidating all the HARP, BDR, and pglogical release notes under PGD, I moved the upgrade information out of the release notes and into this topic. See https://deploy-preview-2909--edb-docs-staging.netlify.app/docs/pgd/latest/upgrades/upgrade_paths/. Is the information correct? Any recommendations for places we might want to link to this topic from?

@kelpoole , I believe this is the last change we wanted to make before merging this PR, correct?

PJMODOS (approved):
I think this is fine, some worry about the upgrade path version table maintainability but let's see

PR #2916, opened 2022-07-11
Remove the vague hint for props value

Jagdish-Kewat-EDB:
The hint for a 4 node BDR cluster mentions some vague values for `max_replication_slots` and `max_worker_processes`. These values don't seem to be rational and hence dropping the hint.

**Content**

- [ x] This PR removes existing content


ashutosh-bapat (approved):
The hint goes against all the advice in that section. Also setting max_replication_slots and max_worker_processes seems unintuitive when max_walsenders is set to a lower value.

PR #2920, opened 2022-07-12
navigational ordering for Hadoop

dwicinas:
Inserts files in navigation item in front matter of index files

Assorted name changes and link changes too


dwicinas:
Not too hard. I think this is ready for merging.

drothery-edb (approved):

PR #2931, opened 2022-07-14
MySQL FDW - install topics

nidhibhammar:
The install topics are generated but not deployed

This is the Debian/Ubuntu template work for the currently supported versions for MySQL.


dwicinas (dismissed):
Looks good

dwicinas:
@nidhibhammar @drothery-edb , I accidentally rebased this PR to develop. I thought I was working in a branch of my own. (Sorry!) The rebase encountered no issues, so I don't think I caused any problems. I just wanted you both to understand why you'll see some commits from me.

drothery-edb (changes requested):
The links on the Installing on Linux and the x86 landing pages are broken in the deployment. Those need to be fixed before we merge this.

drothery-edb (approved):

drothery-edb:
Just noticed that we are generating a Debian 11 topic. We shouldn't be doing that in this branch.

PR #2937, opened 2022-07-15
EDB*Plus: removing install prefixes

ccestes:
## What Changed?
- removed prefixes from install topics and index files
- added navigation to frontmatter
- removed prefix parameter from deploy.mjs


dwicinas (approved):
The spacing of the TOC on the PowerPC index page is single spaced. But the markup looks correct, so maybe it defaults to single spacing when there are only two items? Doesn't seem like a big deal to me, so I am approving without requesting changes.

PR #2948, opened 2022-07-19
Ark: archiving source files

drothery-edb:
## What Changed?

Created a PDF and added it to docs/archive
Removed Ark folder


drothery-edb:
@josh-heyer found one more instance of "ark" in the legacy_redirects_metadata.json file that I didn't remove. It seems like it was ok to keep.

josh-heyer (dismissed):

PR #2965, opened 2022-07-21
max_replication_slot  and parallel apply

ashutosh-bapat:
Fix the paragraph explaining max_replication_slot setting for parallel apply

**Content**

### DIFF
--- a/product_docs/docs/pgd/4/overview/bdr/configuration.mdx
+++ b/product_docs/docs/pgd/4/overview/bdr/configuration.mdx
@@ -47,10 +47,10 @@ N slots and WAL senders. During synchronization, BDR temporarily uses another
 N - 1 slots and WAL senders, so be careful to set the parameters high enough
 for this occasional peak demand.
 
-With parallel apply turned on, the number of slots must be increased to
-N slots from the formula \* writers. This is because the `max_replication_slots`
-also sets the maximum number of replication origins, and some of the functionality
-of parallel apply uses extra origin per writer.
+`max_replication_slots` sets the maximum number of replication origins.
+Some of the functionality of parallel apply uses an extra origin per writer.
+Hence, when parallel apply is enabled, you must set the `max_replication_slots` to
+`N \* (number of writers)` plus the number of slots needed for peak demand as described in the previous paragraph. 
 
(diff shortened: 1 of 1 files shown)


drothery-edb (dismissed):

drothery-edb:
@ashutosh-bapat , does this need a technical review from @pavanvd before merging? Should the change be applied to 3.7 also?

ashutosh-bapat:
> @ashutosh-bapat , does this need a technical review from @pavanvd before merging? Should the change be applied to 3.7 also?

Yes to both.

pavanvd:
I think we should make couple of things clear, so that users can make the right choice:
- changing max_replication_slots requires server restart
- the desired value depends on the size of the cluster, including any future node additions and future increase in the number of parallel writers
- cost associated with having a higher value than strictly needed (low, medium, high etc)

Since the parameter change requires a server restart, if the user starts with a strict value (say 4 for a 4 node cluster) and later wants to add more nodes to the cluster, it will require them to change the value of existing nodes first, restart those nodes and then only be able to add more nodes. So they might prefer to configure a slightly higher value to begin with, assuming the cost is low (which I think is the case)

Also, we should add a line about how Postgres derives max origins from max_replication_slots (there is no separate GUC) and hence it's necessary to increase max_replication_slots with parallel writers even though writers themselves don't consume replication slots (they consume origins)

drothery-edb:
@ashutosh-bapat once you incorporate @pavanvd 's comments, we can copy the final content to 3.7. In the meantime I copied the draft to the new source location of the BDR docs. The old location will be removed soon.

djw-m:
@ashutosh-bapat are these changes still relevant and can we get an update on the content to match @pavanvd 's comments. Thanks.

ashutosh-bapat:
> @ashutosh-bapat are these changes still relevant and can we get an update on the content to match @pavanvd 's comments. Thanks.

@djw-m I believe the changes are still relevant. I never got time to look at Pavan's comments. But I am not working on those anymore.

djw-m:
Cherry picked in https://github.com/EnterpriseDB/docs/pull/4662 for changes to live on. This PR is now closed.

PR #2970, opened 2022-07-21
templatize redhat install topics for MTK

dwicinas:
New approach for:

Redhat 8/ OL 8
Redhat 7 / OL 7
Rocky Linux 8 / Almalinux 8
Centos 7
Redhat 8 (ppc)


dwicinas:
@drothery-edb @nidhibhammar This PR is ready for your review. 

Let me try to explain what I did. 

In my last PR, I restructured templates and successfully generated the Red Hat family topics for EFM. When I tried to extend that approach to MTK, I realized I needed more flexibility. Almost every topic has the same kind of commands (or a subset of those commands), but often the values used in the commands vary between products. 

After much experimenting, I decided to reintroduce the use of a macro. Previously we had a macro in platformBase/_shared.njk. It was called something like "centosRepositoryConfigure". I had stopped using that macro in my last PR, but now I've decided to reintroduce it. I renamed the macro "redhatFamilyRepositoryConfiguration". 

This macro provides text for the first three commands that all products use (I think). I think of those commands as "set up the EDB repository," "enter credentials", and "install the EPEL repository". In those commands there are two variables you must define, packageManager and epelRepo. That's why the macro is useful.

In this PR, for all the Red Hat related templates, I had to add instructions to import the macro. For some of OS templates, I also set the values of packageManager  and epelRepo. For other OSs, I set the value of packageManager  and epelRepo at the product level rather than at the OS level. 

This approach gave me the flexibility I needed for the first two products, EFM and MTK. Hopefully these changes don't conflict with Nidhi's work on EPAS. I don't think they do, but we would need to merge and then run a build to be certain.

**One problem with this approach** is that install-docs:build does not work for all products! It encounters errors for some products. Resolving those errors might take substantial troubleshooting. For the purposes of this PR, I used config.yaml and commented out every product except EFM and MTK. Then, after generating the appropriate files, I discarded the changes to config.yaml. This technique may need to be part of our regular work flow until we have the kinks worked out of this process. 

Hopefully these lengthy notes are understandable! Sorry I won't be around for a week to take follow-up questions. That's why I tried to put all my thinking in writing.

drothery-edb:
Used macro approach in https://github.com/EnterpriseDB/docs/pull/2998 instead

PR #2976, opened 2022-07-26
PEM - Redhat install topics

nidhibhammar:


drothery-edb:
@nidhibhammar please resolve the conflicts so that we can see a deployment with the new topics.

drothery-edb:
Comments for server topics:

- Link to Prerequisites topic is broken
- Link to Configuring topic is broken
- Are the changes to the Initial configuration topic all by design? There is a lot of information in the published topics that doesn't display (at list for RHEL 8).
- I think there are formatting issues with the note that is resulting in html displaying at the bottom.
- Is there an easy way to note have the note be bulleted when there is only one issue in the note? I'm guessing not.
- Remove link to Debian 9 topic from https://deploy-preview-2976--edb-docs-staging.netlify.app/docs/pem/latest/installing_pem_server/pem_server_inst_linux/installing_pem_server_using_edb_repository/x86_amd64/

drothery-edb (approved):

PR #2985, opened 2022-07-28
Postgres for Kubernetes 1.16

josh-heyer:
## What Changed?

This brings the latest docs for EDB Postgres for Kubernetes - the FULL set of docs - into EDB Docs. 

I've done my best to preserve edits that were made previously to the subset of files already brought to EDB Docs, but I may have missed some - would appreciate a 2nd set of eyes to spot  omissions.

In the interest of automating most of this for the next release, I've created a PR with the same reconciliation edits here: https://github.com/EnterpriseDB/cloud-native-postgres/pull/97

The automated PR that this was based on can be found here: https://github.com/EnterpriseDB/docs/pull/2880

## Checklist

**Content**

- [x] This PR is for a release, please add this tag: product/postgresforkubernetes/1.16


josh-heyer:
Added special support for merging the upstream and in-repo index file to allow us fine-grained control over navigation and other EDB Docs options. Here's how it works:

- Upon import, the frontmatter yaml from the existing index.mdx file will be kept as-is. The rest of the content will come from the upstream index.md file, *provided* the `originalFilePath` entry in the frontmatter remains intact (in general, any files with this key in their frontmatter will be replaced when pulling from the upstream repo).
- Any entries in the frontmatter `navigation` list beginning with `!` will cause matching upstream files to be ignored. This can be used to prevent importing content we don't want here (ex: prevent the inclusion of release_notes and commercial_support) 

See [index.mdx](https://github.com/EnterpriseDB/docs/blob/docs/josh/cloud-native-postgres-1.16-sync/product_docs/docs/postgres_for_kubernetes/1/index.mdx) for examples

drothery-edb:
@josh-heyer I wiped out some of your changes resolving a merge conflict with index.mdx. I think we need to exclude the n2n and supported releases topics in addition to the release notes topics.

josh-heyer:
What's the reason for removing e2e @drothery-edb ?

drothery-edb:
> What's the reason for removing e2e @drothery-edb ?

Kelly didn't think we needed it. Will it cause an issue since it is still listed in the nav option in the front matter? (see line 48 of the index file.)

josh-heyer:
> Kelly didn't think we needed it. Will it cause an issue since it is still listed in the nav option in the front matter? (see line 48 of the index file.)

Won't cause problems if the file doesn't exist. We can remove it from the nav though if we're sure we don't want it.

PR #2991, opened 2022-07-31
Kubernetes: reworked diff content

drothery-edb:
## What Changed?

Incorporated content from the previous Differences topic into the landing page


drothery-edb:
Starting over

PR #2998, opened 2022-08-03
Implement red hat family install templates for MTK

dwicinas:
Install templates for MTK for:

- RHEL 8 x86
- RHEL 8 ppc
- RHEL 7
- Alma/Rocky Linux 8
- Centos 7


dwicinas:
@nidhibhammar @drothery-edb This PR uses the non-macro approach to implementing RedHat. For Alma Linux I had to repeat some verbiage, but other than that it was easy to implement.

drothery-edb (approved):

PR #3008, opened 2022-08-04
Outline strategy for forking product docs without losing history

josh-heyer:
## What Changed?

Wrote up a short process for doing long-lived "next product version docs" branches without quite so many headaches: https://github.com/EnterpriseDB/docs/blob/process/josh/product-fork-process-proposal/docs/agreements/product-version-fork.md

Would appreciate your review, @drothery-edb and @nidhibhammar - with your approval, I'd like to try this first with the EPAS 15 branch (work on which is already underway) and also the PGD 5 branch (yet to be created).

### DIFF
--- a/docs/agreements/product-version-fork.md
+++ b/docs/agreements/product-version-fork.md
@@ -1,0 +1,108 @@
+# Docs repository product version forking strategy
+
+## Version forking in brief
+
+"Forking" is currently (and traditionally) our strategy for handling product versions in the documentation for EDB products. In short, "forking" means that when a new (major) version of a product is released, a new set of documentation is released along with it, based on the documentation for the previous version but with updates to reflect the changes present in the new version of the product.
+
+For example: when EDB Postgres Advanced Server 14 was released, a new version of the documentation was created in this repo at product_docs/docs/epas/14/ - this was based on the documentation at product_docs/docs/epas/13/, but contains several key differences.
+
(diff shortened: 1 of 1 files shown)


nidhibhammar:
@josh-heyer, I worked on EPAS 15 branch with this approach. Everything was smooth except for one thing I struggled with which was rebasing the branch. Do you think staging the changes in VS code will help to handle this in a better way?

josh-heyer:
@nidhibhammar I do tend to think VSCode / GitLens is a better UI for rebasing than Github Desktop; IIRC, the other factor with EPAS15 was that we were doing some pretty substantial reorganization in the current version (14) while at the same time working on EPAS15 (moving out EDB*Plus, re-arranging installer topics, etc.) - that ended up making it harder to keep both in sync...

...Which I think is @djw-m 's concern: this still isn't particularly friendly to major reorgs. That said, I think *some* of the issues last time came down to a lack of discipline with reorgs:

- keeping reorg branches focused and scoped helps a lot: some of the issues with installers were that they ended up touching files across multiple products in multiple branches at once, which meant a lot of conflicts even *without* the version split. We can and are doing better with that sort of thing now.
- Ensuring that file moves / renames are detected by git helps a great deal, as many conflicts can be resolved automatically when that happens. Once again, we saw in EPAS15 multiple instances where files got moved around but not detected properly - so files were "added" by multiple branches and we had to sort out which ones were correct. We've discussed this somewhat, and this is an area where I particularly like GitLens for making it easier to see when a file move/rename is not detected.

PR #3011, opened 2022-08-05
PostGIS fixes

drothery-edb:
## What Changed?

Fixing install topics and adding 3.1 info
Added CentOS 7 topic that was missing
Misc other fixes


drothery-edb:
any new updates made to the install topics should be made in the template files, not the generated topics

theadamwright (dismissed):

install_template/templates/products/postgis/almalinux-8-or-rocky-linux-8.njk
```diff
@@ -1,2 +1,16 @@
...
+{% block installCommand %}
+
+```shell
+# To install PostGIS 3.2:
+dnf install edb-as<xx>-postgis32
+
+# To install PostGIS 3.1 using EDB Postgres 13-14:
+dnf install edb-as<xx>-postgis3
+
+# To install PostGIS 3.1 using EDB Postgres 11:
```
nidhibhammar, line 12: I think this is true for EDB Postgres 11 and 12 both

drothery-edb: per the repos, 3.1 for epas 12 was not released

PR #3021, opened 2022-08-09
MIG-3883

moiznalwalla:
https://enterprisedb.atlassian.net/browse/MIG-3883

### DIFF
--- a/product_docs/docs/migration_portal/4/images/mp_overview_home.png
+++ b/product_docs/docs/migration_portal/4/images/mp_overview_home.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:5cc9e9cc62dfa430f832c309a86d5d2ced9988cf6def20b70f9e1a2210f06e7a
-size 380127
+oid sha256:83ce76d6da7e6af7fdb5e479ae974e28731e32270a0d54ef9407730c6e769280
+size 375443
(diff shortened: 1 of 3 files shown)


moiznalwalla:
> It's not easy to review and comment changes to the png files in github and it's difficult to find the pages in which they appear in the deployment version.
> 
> Is this commit only replacing the mp_overview_home.png and mp_using_portal.png files with new versions?
> 
> Is the mp_using_portal_expdp_impdp.png simply there in case we want to use it instead (BTW, I can't view it to see what it looks like)?
> 
> The mp_overview_home.png screenshot includes a red box around the quick help links. This should not be included in the screenshot.
> 
> We only need to show a flow that includes the DDL extractor. I don't think we need a separate flow that shows the data pump utility approach as this is not the preferred approach.

Yes, mp_using_portal_expdp_impdp.png is simply there in case we want to use it in future. 

I am pasting the links of the pages with the updated screenshots, for easy reference:

DRAFTs: 

https://deploy-preview-3021--edb-docs-staging.netlify.app/docs/migration_portal/latest/03_mp_using_portal/ 

https://deploy-preview-3021--edb-docs-staging.netlify.app/docs/migration_portal/latest/03_mp_using_portal/02_mp_overview_project/ 

https://deploy-preview-3021--edb-docs-staging.netlify.app/docs/migration_portal/latest/03_mp_using_portal/01_mp_overview_home/

mlewandowski-edb (approved):

Issue #3022, opened 2022-08-09
Code Snippets should copy only the command, not the output


serdardalgic:
When I copy a code from any of the code snippets in https://www.enterprisedb.com/docs/biganimal/latest/reference/cli/ page, It copies the whole block, including the output, such as:
```
$ biganimal reset-credential ba-user1 
Visit this URL to login: https://auth.biganimal.com/activate?user_code=****-****
or press [Enter] to continue in the web browser 

Credential "ba-user1" reset operation succeeded
```
It should copy only the command without the output, e.g. 
```
biganimal reset-credential ba-user1 
```

A working example for this behaviour can be found in the terraform docs, e.g. https://learn.hashicorp.com/tutorials/terraform/init?in=terraform/cli Copy any of the code snippets, although it shows the output, it doesn’t copy it.

serdardalgic:
This change looks like it has been merged. The ticket can be closed as success, @josh-heyer @drothery-edb Thanks again!

PR #3025, opened 2022-08-10
Separate shell commands from output listings in BigAnimal docs

josh-heyer:
## What Changed?

Addresses issue #3022 - inability to copy certain commands in shell listings due to the presence of output listings.

Three conventions should be used in these docs for shell code blocks:

1. Do not lead with a "prompt" (such as `$` or `#`) - if copy-pasted, these will lead to either "command not found" or a no-op, both which are confusing.

2. Separate command output from the command(s) themselves by a line containing only the text `__OUTPUT__`. When built, this will result in a separate (but visually-connected) listing that will not be copied.

3. Break long lines in a consistent and visually-appealing fashion to aid in reading long commands. 70 characters give or take is a good rule of thumb - if more than two lines, might as well go with one parameter per line.

## Checklist

**Content**

### DIFF
--- a/product_docs/docs/biganimal/release/getting_started/preparing_cloud_account/01_preparing_azure/index.mdx
+++ b/product_docs/docs/biganimal/release/getting_started/preparing_cloud_account/01_preparing_azure/index.mdx
@@ -126,7 +126,7 @@ To check if an Azure resource provider is registered, use the following command.
 
 ```shell
 az provider show -n Microsoft.ContainerService
-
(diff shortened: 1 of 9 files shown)


drothery-edb (approved):
I reviewed this with Serdar and he is happy with the improvements.

serdardalgic:
Thank you very much for the implementation @josh-heyer , with @drothery-edb we checked the local build together and the result is really good 👍

PR #3002, opened 2022-08-04
Template changes and generated changes for Hadoop

dwicinas:
Changes necessary to templatize install topics for Hadoop, MongoDB, and MySQL FDW

PR #3029, opened 2022-08-10
MySQL: changes to config and deploy for RHEL

drothery-edb:
## What Changed?

Copied Dave's MySQL changes to config.yaml and deploy.mjs before reverting his changes to #3002.


drothery-edb:
Not needed

PR #3041, opened 2022-08-12
MP: code block formatting experimentation

drothery-edb:
## What Changed?
@josh-heyer There are some code blocks in the Migration Portal docs that could use some love similar to what you did for BigAnimal in this PR:  https://github.com/EnterpriseDB/docs/pull/3025. 

Questions: 

- Is there a way to turn off OUTPUT formatting in a code block? The scenario is an example that shows prompts and then output and another prompt. In my experimenting I created another code block for the next prompt, but I wonder if there is a way to represent this in one block. See the examples in this section: https://deploy-preview-3041--edb-docs-staging.netlify.app/docs/migration_portal/latest/04_mp_migrating_database/01_mp_schema_extraction/#for-sqlplus

- I tried ini formatting for a couple code blocks that had really long values. Seems Toggle wrap doesn't work with ini. Should we stick with text or shell or try manually formatting with ini (that seems like a bad idea to me)? can ini be updated so it Toggle wrap works? See step 1 here: https://deploy-preview-3041--edb-docs-staging.netlify.app/docs/migration_portal/latest/04_mp_migrating_database/01_mp_schema_extraction/01_data_pump_utility/#extract-all-schemas-in-a-database.


josh-heyer:
Some quick but detailed answers to your questions:

1. **Output formatting for multiple inputs**: This is a bit of a judgment call, but I would try to avoid mixing in OUTPUT and multiple code blocks. When possible, divide a sequence of commands into discrete steps - the output of each can then be shown. When multiple commands are effectively "atomic" (e.g., two steps always performed together, the output of one command fed directly into the next, etc.) then describe their purpose before showing them together - as they would be typed / entered - and list the output of both together. All that said...

  **Be careful of things in code blocks that are not cleanly divisible into input and output:** this is the case for the examples you provided! This block:
   https://github.com/EnterpriseDB/docs/blob/08498538eab5f5542472bb90281d8d63b128a271/product_docs/docs/migration_portal/4/04_mp_migrating_database/01_mp_schema_extraction/index.mdx?plain=1#L61-L63
 ...is both output (the two prompts) and input (the CSV value `HR, SCOTT, FINANCE` and the path value `/home/oracle/extracted_ddls/` provided in response to the prompt).  The warning that follows (and which you've *marked* as output) is actually a prelude to the **next** prompt in the following block: 
   https://github.com/EnterpriseDB/docs/blob/08498538eab5f5542472bb90281d8d63b128a271/product_docs/docs/migration_portal/4/04_mp_migrating_database/01_mp_schema_extraction/index.mdx?plain=1#L73
   These situations can be difficult to evaluate without more detailed knowledge of the software itself (more on that below), but the context itself can give us clues: the process being shown in these blocks is the same process *described* just above it in the numbered list. With that in mind, we might restructure these examples to exactly mirror those steps, breaking each prompt + input out into its own block associated with a numbered step. I've illustrated this here: 

  https://github.com/EnterpriseDB/docs/blob/a83f42e61f00267134186b771f9b3aa20baa4cad/product_docs/docs/migration_portal/4/04_mp_migrating_database/01_mp_schema_extraction/index.mdx?plain=1#L58-L99

  ...In my commit, I've also collapsed the two examples into one for brevity (the ONLY difference was the format of the path between Linux and Windows, which hardly seems worth repeating the rest). I've also marked all of these as "text" as none of this is actually SQL syntax and the only SQL*Plus syntax is found in the invocation of the script that starts these prompts (thus, any highlighting will be more of a distraction than helpful). Finally, I've removed the `SQL>` prompt from the first block (in the instructions), thus allowing it to be copied and pasted directly into SQL*Plus; I've left the prompt in the example, where - like the rest of the blocks that follow - it may serve to illustrate both the prompt and *a* potential response to that prompt.

2. There's nothing special about `ini` formatting - the issue with that block is simply that there's no obvious place to break - no spaces separating words.  I can adjust the styling used to wrap to address this if need-be - `overflow-wrap: anywhere; word-break: normal;` gives good results:
  image

  But ideally we try to restructure these blocks to avoid the need for wrapping in the first place where possible. This is where a bit of knowledge about the tool being described becomes useful: I was able to find a couple of handy references for the the Oracle Data Pump utility's Include parameter:

  1. https://docs.oracle.com/database/121/SUTIL/GUID-A3C5241C-098C-42FF-B7D1-6018B2ABC282.htm#SUTIL853
  2. https://oracle-base.com/articles/10g/oracle-data-pump-10g#IncludeExclude

  In reviewing them, we quickly discover that a single, comma-separated list of object types may also be specified as multiple `INCLUDE=` statements, thus allowing that line to be rewritten: 
  https://github.com/EnterpriseDB/docs/blob/a94fceaa68f1eae4b60769f7009449c8b0d5fa9c/product_docs/docs/migration_portal/4/04_mp_migrating_database/01_mp_schema_extraction/01_data_pump_utility.mdx?plain=1#L44-L60

  Changes such as this must be undertaken with care; ideally, we would test this process to ensure that the command behaves identically with both formats. Discussion or review with topic experts are also a good idea in cases where we're unfamiliar with the topic or unable to test. But when possible, the results *can be* much better than what is possible with just styling changes!

  (note that if I'm reading the data pump syntax correctly, it also allows spaces between comma-separated items; this would allow word-wrap to at least be effective without splitting a word across lines and would thus still be preferable to the styling change I illustrated above)

drothery-edb:
@josh-heyer thank you for the master class in code blocks and writing good examples. @moiznalwalla, can you follow up on the code blocks on the Data Pump topic? Besides the one that Josh updated there are two others that have similar wrapping issues. I wonder if simply adding spaces, once confirmed a user can copy and paste the command without issue, is the way to go with all three examples.

moiznalwalla:
Its not clear to me whether INCLUDE allows spaces between the comma separated object types. So I have used Josh's suggestion of adding multiple Include= lines. Also, incorporated some other feedback from Josh in the pages.

moiznalwalla:
> Its not clear to me whether INCLUDE allows spaces between the comma separated object types. So I have used Josh's suggestion of adding multiple Include= lines. Also, incorporated some other feedback from Josh in the pages.

Team approved the change. This si good to merge.

moiznalwalla (approved):
Changes are approved by Team so merging this branch now.

PR #3057, opened 2022-08-16
Update efm4_rhel_8_x86.mdx

martinmarques:
The less and greater symbols may confuse the user, and they actually need to change the `USERNAME:PASSWORD` part.

## What Changed?

Remove the `<` and `>` which were incorrect and might lead to confusion.

## Checklist

**Content**

- [ x ] This PR changes existing content

### DIFF
--- a/product_docs/docs/efm/4/03_installing_efm/x86_amd64/efm4_centos7_x86.mdx
+++ b/product_docs/docs/efm/4/03_installing_efm/x86_amd64/efm4_centos7_x86.mdx
@@ -18,7 +18,7 @@ Setting up the repository is a one-time task. If you have already set up your re
 #  Set up the EDB repository:
 yum -y install https://yum.enterprisedb.com/edbrepos/edb-repo-latest.noarch.rpm
 
-#  Replace '<USERNAME>' and '<PASSWORD>' below with
+#  Replace 'USERNAME' and 'PASSWORD' below with
 #  your username and password for the EDB repository:
 sed -i "s@<username>:<password>@USERNAME:PASSWORD@" /etc/yum.repos.d/edb.repo
 
--- a/product_docs/docs/efm/4/03_installing_efm/x86_amd64/efm4_other_linux8_x86.mdx
+++ b/product_docs/docs/efm/4/03_installing_efm/x86_amd64/efm4_other_linux8_x86.mdx
@@ -18,7 +18,7 @@ Setting up the repository is a one-time task. If you have already set up your re
 #  Set up the EDB repository:
 dnf -y install https://yum.enterprisedb.com/edbrepos/edb-repo-latest.noarch.rpm
 
-#  Replace '<USERNAME>' and '<PASSWORD>' below with
+#  Replace 'USERNAME' and 'PASSWORD' below with
 #  your username and password for the EDB repository:
 sed -i "s@<username>:<password>@USERNAME:PASSWORD@" /etc/yum.repos.d/edb.repo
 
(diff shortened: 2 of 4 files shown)


martinmarques:
@drothery-edb I see this same pattern in other pages of the EFM docs (namely all the installations that branch from [this section](https://www.enterprisedb.com/docs/efm/latest/03_installing_efm/x86_amd64/). Any chance we can replicate the simple patch on the other sections if @EFM-Bobby agrees :-)

EFM-Bobby:
To me your change implies that they need to change <username>:<password> to <somename>:<somepassword> and keep the <> symbols.

martinmarques:
The code snippet is this:

```
#  Replace '<USERNAME>' and '<PASSWORD>' below with
#  your username and password for the EDB repository:
sed -i "s@<username>:<password>@USERNAME:PASSWORD@" /etc/yum.repos.d/edb.repo
```

That will replace occurrences of `<username>` with `USERNAME` and occurrences of `<password>` with `PASSWORD`. That's why we should remove the `<` and `<` from the comment in the code.

martinmarques:
One thing to note is that this change doesn't affect Ubuntu and Debian installation instructions as in those it's correctly commented with respect to the code that installs the repository

EFM-Bobby:
> The code snippet is this:
> 
> ```
> #  Replace '<USERNAME>' and '<PASSWORD>' below with
> #  your username and password for the EDB repository:
> sed -i "s@<username>:<password>@USERNAME:PASSWORD@" /etc/yum.repos.d/edb.repo
> ```
> 
> That will replace occurrences of `<username>` with `USERNAME` and occurrences of `<password>` with `PASSWORD`. That's why we should remove the `<` and `<` from the comment in the code.

But we want "<username>" replaced with the name and not just "username" replaced with the name don't we? Or do we want the <>s in the line after the replacement? I have not used these instructions and don't understand what the problem is.

EFM-Bobby:
> > The code snippet is this:
> > ```
> > #  Replace '<USERNAME>' and '<PASSWORD>' below with
> > #  your username and password for the EDB repository:
> > sed -i "s@<username>:<password>@USERNAME:PASSWORD@" /etc/yum.repos.d/edb.repo
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > That will replace occurrences of `<username>` with `USERNAME` and occurrences of `<password>` with `PASSWORD`. That's why we should remove the `<` and `<` from the comment in the code.
> 
> But we want "" replaced with the name and not just "username" replaced with the name don't we? Or do we want the <>s in the line after the replacement? I have not used these instructions and don't understand what the problem is.

certainly USERNAME->username ought to be fixed though

EFM-Bobby (approved):
Yes, I'm ok with this once I finally saw what was going on. Like written elsewhere, this change should probably be in a lot of places.

martinmarques:
Added changes to the RHEL7 and Centos7 files which required the same fix in the docs

martinmarques:
I think I got all changes in now.

EFM-Bobby (approved):
Looks good.

EFM-Bobby:
(Removed my other comments when I wasn't understanding which words were being replaced -- the latter instances of username/password -- so it doesn't cause even more confusion.)

drothery-edb:
These are very good comments but as discussed in Slack, we will be closing this PR because these topics are generated. We will add a comment to that effect in the generated topics going forward. Since we are very close to updating these topics to not include the repo instructions, we aren't planning to make this temporary improvement as the change would affect many topics (all RHEL family install topics for all the EDB heritage products) and the testing effort isn't quite justified.

PR #3060, opened 2022-08-17
EPRS: 6.2 SSL default path note

drothery-edb:
## What Changed?

Added note about default paths when using Linux interactive installer to 6.2 doc


drothery-edb:
@muhammadimtiaz can we merge this? do you have any comments?

PR #3062, opened 2022-08-17
LiveCompare: consolidating versions

ccestes:
## What Changed?

- added individual version topics 
-  organized topics as tables
-  deleted and renamed versions to consolidate
- made copy edits
- added redirect links


drothery-edb:
@ccestes do a search of the docs for each feature to see if there is a topic or section we could link to for more information. For example, the first feature listed on the 2.2 release notes could point here: https://deploy-preview-3062--edb-docs-staging.netlify.app/docs/livecompare/latest/command_line_usage/#dry-run-mode. Then in that section add a note using the same format we use in the PEM docs for indicating when a new feature was added in a minor release. Here is an example: https://www.enterprisedb.com/docs/pem/latest/pem_cli/.

You don't need to do this for new platform support and other new features also might not have doc impact.

drothery-edb:
@dougcalobrisi , want to take a look at the changes Chris made when consolidating the versions? Do you think there is any more linking we should do for the new features? Or are we good to merge and publish?

dougcalobrisi:
@drothery-edb @ccestes 

Thanks for this - I think the current links in the Release Notes look good for v2.0. Are we going to do the same for v1.0?

Additionally:
- we should remove the copyright notices on all versions, as @alwaite noted we don't have those for other product docs and they are outdated
- we need to remove two enhancements from the most recent release notes (2.3). We'll be adding them back for the next release.

>   - Support for Ubuntu 22.04 (Jammy) on x86_64.
>   - Support for Debian 11 (Bullseye) on x86_64.

@alwaite is going to review further next week.

drothery-edb:
@dougcalobrisi and @ccestes,  I removed the copyright notice and the old 1.0 release notes, which were in Appendix A. I renamed Appendix B and added redirects.

In terms of the links in the v1 rel notes, I'm guessing why Chris didn't add any is because there were no release notes of type Feature and typically we just call out new features. He can review the enhancements and see if any have a logical place to link to for more info.

ccestes:
@drothery-edb @dougcalobrisi Like Dee Dee said, I was looking for Features to add links to, which I only found in v2. But, I found a relevant section for two Enhancements in the v1 release notes, so I added links to it. 

If there's anything else that should be linked, let us know.

alwaite (approved):

PR #3079, opened 2022-08-23
BigAnimal: Control status remediation

moiznalwalla:


product_docs/docs/biganimal/release/administering_cluster/customizing_compliance/aws_controls.mdx
```diff
@@ -0,0 +1,19 @@
+---
+title: "Remediating AWS control findings"
+
+---
+
+This page describes the control findings in your AWS Security Hub and suggests potential remediation steps for your BigAnimal resources.
```
l0r3bo, line 6: I would remove the term "control" which is of very widespread use (we tend to use it a lot internally, I may have used it myself when discussing this), but difficult interpretation. Moreover it is not used at all in the security hub documentation, which only talks about "findings".

Suggested rewording: "This page describes findings in your AWS Security Hub...".

Whether we want to add also "that may relate to BigAnimal cloud resources" is an interesting point, I think we should if that's not clear/implied by other pieces of documentation. And especially we should be consistent with what is discussed on the Azure side.

product_docs/docs/biganimal/release/administering_cluster/customizing_compliance/aws_controls.mdx
```diff
@@ -0,0 +1,19 @@
...
+title: "Remediating AWS control findings"
+
+---
+
+This page describes the control findings in your AWS Security Hub and suggests potential remediation steps for your BigAnimal resources. 
+
+!!! Important
+    Remediation steps provided below might increase your cloud costs. Before proceeding, make sure that you evaluate and prioritize the suggested actions based on your organization policies. 
+
+The control findings are listed below based on the level of impact to your resources.
```
l0r3bo, line 11: Same remark as above on the usage of "control". Suggested rewording "The findings...".

moiznalwalla: incorporated your feedback

product_docs/docs/biganimal/release/administering_cluster/customizing_compliance/aws_controls.mdx
moiznalwalla: Incorporated your feedback.

drothery-edb:
obsolete

PR #3113, opened 2022-09-01
BigAnimal: Configuring Pgpool

ccestes:


drothery-edb:
We decided to not include this topic, but instead add a few sentences about the benefits of using PgPool with BigAnimal and link out to the PgPool docs. See https://github.com/EnterpriseDB/docs/issues/3812

PR #2884, opened 2022-07-01
Patroni support documentation

josh-heyer:
## What Changed?

This is based on #2812 but with a branch instead of a fork so that we can do a deployment. Also fixed up two images that should have been LFS-tracked but weren't.

This PR adds documentation and examples about using a supported Patroni topology.

Note: the Vagrantfile examples are not exposed to the users and are just there for our convenience to run the quick-start examples locally if needed

## Checklist

**Content**

- [X] This PR adds new content
- [ ] This PR changes existing content
- [ ] This PR removes existing content
- [ ] This PR is for a release, please add this tag:

PR #3142, opened 2022-09-08
Patroni: copy edit

drothery-edb:
## What Changed?

Applied style conventions and copy edits to content


drothery-edb:
Using https://github.com/EnterpriseDB/docs/pull/2884 for the copy edit instead of this branch.

PR #3157, opened 2022-09-14
55.3.0

moiznalwalla:
https://enterprisedb.atlassian.net/browse/MIG-3962


product_docs/docs/migration_toolkit/55/06_building_toolkit.properties_file.mdx
```diff
@@ -297,29 +297,53 @@ Migration Toolkit reads connection specifications for the source database from t
 -   The name of a privileged user
 -   The password associated with the specified user
 
-If you're connecting to a SQL Server database, `SRC_DB_URL` takes the form of a JTDS URL. For example:
+If you're connecting to a SQL Server database, `SRC_DB_URL` takes the form of a JTDS URL or Microsoft JDBC URL, as follows:
```
drothery-edb, line 300:
```suggestion
If you're connecting to a SQL Server database, `SRC_DB_URL` takes the form of a [JTDS URL](#jtds-url) or [Microsoft JDBC URL](microsoft-jdbc-url).
```

mlewandowski-edb: Is a # required before "microsoft-jdbc-url" in the hyperlink definition?

mlewandowski-edb (changes requested):
Can we move the Defining a SQL Server URL  section to before the MySQL URL and Sybase section.  Also, I suggest reversing the order of the JTDS URL and Microsoft JDBC URL sections within the SQL Server URL section as it will be the preferred/recommended driver for SQL Server going forwarded.  The note in the Sybase URL about the jTDS driver would be relevant to the jTDS URL in SQL Server URL section too.

product_docs/docs/migration_toolkit/55/06_building_toolkit.properties_file.mdx
moiznalwalla: Fixed.

moiznalwalla:
> Can we move the Defining a SQL Server URL section to before the MySQL URL and Sybase section. Also, I suggest reversing the order of the JTDS URL and Microsoft JDBC URL sections within the SQL Server URL section as it will be the preferred/recommended driver for SQL Server going forwarded. The note in the Sybase URL about the jTDS driver would be relevant to the jTDS URL in SQL Server URL section too.

Incorporated.

mlewandowski-edb (approved):
I updated the tips related to the jTDS driver for SQL Server and Sybase ASE.

mlewandowski-edb (dismissed):

mlewandowski-edb (commented):
All of the installation pages seem to include this same first note in the Prerequisites section:

> Install EDB Postgres Advanced Server on the same host.

Is it really required that EDB Postgres Advanced Server be installed on the same host as Migration Toolkit?  I don't think it is for a source or target EPAS database.  Or is this note here to ensure that certain libraries are installed for use by Migration Toolkit?  If so, we may want to include some clarifying note.

drothery-edb:
> All of the installation pages seem to include this same first note in the Prerequisites section:
> 
> > Install EDB Postgres Advanced Server on the same host.
> 
> Is it really required that EDB Postgres Advanced Server be installed on the same host as Migration Toolkit? I don't think it is for a source or target EPAS database. Or is this note here to ensure that certain libraries are installed for use by Migration Toolkit? If so, we may want to include some clarifying note.

@mlewandowski-edb We either have that as a prerequisite or this:

- Install EDB Postgres Advanced Server on a host that the product can connect to using a connection string. It doesn't need to be on the same host. See [Installing EDB Postgres Advanced Server](https://www.enterprisedb.com/docs/epas/latest/epas_inst_linux/).

We might have this all wrong. Would it make sense for MTK if we used the bullet that describes connecting using a connection string?

mlewandowski-edb:
@drothery-edb 
> We either have that as a prerequisite or this:
> 
> * Install EDB Postgres Advanced Server on a host that the product can connect to using a connection string. It doesn't need to be on the same host. See [Installing EDB Postgres Advanced Server](https://www.enterprisedb.com/docs/epas/latest/epas_inst_linux/).
> 
> We might have this all wrong. Would it make sense for MTK if we used the bullet that describes connecting using a connection string?

Technically speaking, EPAS is not specifically required.  MTK could be used for migrations to PostgreSQL, albeit in a more limited way.   That being said, I like your revised version of the bullet.  I would go with that for now.

@zahid75 Do you have any opinion on this?

drothery-edb:
> @drothery-edb
> 
> > We either have that as a prerequisite or this:
> > 
> > * Install EDB Postgres Advanced Server on a host that the product can connect to using a connection string. It doesn't need to be on the same host. See [Installing EDB Postgres Advanced Server](https://www.enterprisedb.com/docs/epas/latest/epas_inst_linux/).
> > 
> > We might have this all wrong. Would it make sense for MTK if we used the bullet that describes connecting using a connection string?
> 
> Technically speaking, EPAS is not specifically required. MTK could be used for migrations to PostgreSQL, albeit in a more limited way. That being said, I like your revised version of the bullet. I would go with that for now.
> 
> @zahid75 Do you have any opinion on this?

I changed the bullet in https://github.com/EnterpriseDB/docs/pull/3176 as we were making other change to the install topics in that PR. But, I wonder if the better solution is to just not include a prerequisite regarding epas.

PR #3166, opened 2022-09-16
Rewrite URLs in combined MDX before generating PDF

### LINKED ISSUE #2314: Problem with PDF for HARP - Text size is too large
## Summary

The font/text size in the HARP PDF is much larger than other generated PDFs.

## Where did you see the problem?

HARP PDF: https://www.enterprisedb.com/docs/static/0afb65e45f107afa6bc3de8877a8b485/harp_v2.0_documentation.pdf

## Expected behavior

Text should be the normal font weight used in other PDFs.

## Screenshots
image

### LINKED ISSUE #3167: Relative links don't work in PDFs
[From @kelpoole:](https://edb.slack.com/archives/CU5QEU5L7/p1662680570587429)

> I had a sales rep inquiry (on behalf of a customer) about the links included when a pdf of the doc is created - currently only url links outside the docs site resolve correctly.  Anything that links within the pdf or docs sites doesn’t have a valid url/link - I’m not sure if linking within the doc is valid or can be done but if not having the urls be links to the edb docs site would be better than the current experience.

Primary issue: relative links don't work at all in PDFs. That means MOST docs-internal links don't work, as we default to using relative URLS (of the form `/product/version/...`) for them. Full URLs *will* work, but will resolve to the public website even if the referenced content is also contained within the PDF - which is still an issue for offline reading.

Nice to have: relative links that refer *to the same product* should resolve within the PDF itself vs linking to the website.

Nicer to have: all links that refer to content included in the PDF should resolve within the PDF itself.

josh-heyer:
## What Changed?

Adds a preprocessing step to PDF generation, after the individual topics are combined but before the Markdown is rendered to HTML. Within this step,

1. All links are rewritten as follows:
	- All headings have unique (within the PDF) IDs generated for them and link targets added 
	- All other IDs (ex: `<div id='blah'>`) are rewritten to be unique within the entire PDF (ex: `<div id='blah'></div><div id='blah2'></div>...`)
	- Links that resolve to content that will be included within the PDF (other pages within the same product, sections within the same page, HTML IDs) are rewritten to point to the PDF-specific IDs. This includes both Markdown links and HTML links (`<a href=...`)
	- Links that resolve to content that will *not* be included within the PDF (other products, other pages on EDB's website) are rewritten to be full URLs (when necessary).
2.  Heading depths are increased by 1 (this was formerly done by a regex in the combiner script, but this would also corrupt code on occasion)
3.  Annotations are rendered as blockquotes with the title in bold
4.  Code blocks with `__OUTPUT__` demarcations are split into two code blocks.

Also, I've modified the stylesheet to force table cell contents to wrap such that tables are more likely to render in full, even when doing so results in unattractive breakpoints within cell text. I consider this to be acceptable within the limitations of the PDF format.

The goal with all of these changes is to retain enough fidelity within the rendered PDF to make it suitable as an offline reference.

Fixes #3167
Fixes #2314

## Unfixable links

Needless to say, I am unable to properly rewrite links that... Don't work. While link-checking is not the primary goal of this PR, the foundations of a robust link-checker can be found here; for now though, we should focus on fixing links that this process *knows* cannot be resolved - these can be found by examining the output of the "Build all PDFs" step [in the build output](https://github.com/EnterpriseDB/docs/actions/runs/3071978965/jobs/4963122489)

> ⚠️⚠️  cannot find topic for path: /biganimal/release/getting_started/administering_cluster/01_portal_access/
> ⚠️  cannot find slug for method-2-manually-check-requirements in /biganimal/release/getting_started/preparing_cloud_account/01_preparing_azure/

The first warning listed there notes that the path `/biganimal/release/getting_started/administering_cluster/01_portal_access/` doesn't exist - yet a link references it. We should update that link to point to a valid URL.

The second warning notes that while the path exists, the target hash does not - chances are, a section was renamed or moved, and the link will need to be updated to point to the new section name and/or page.

## Checklist

**Content**

### DIFF
--- a/.github/workflows/deploy-draft.yml
+++ b/.github/workflows/deploy-draft.yml
@@ -10,9 +10,6 @@ concurrency:
 
 jobs:
   build-deploy:
-    if: |
-      (github.event.action == 'labeled' && github.event.label.name == 'deploy') ||
-      (github.event.action != 'labeled' && contains(github.event.pull_request.labels.*.name, 'deploy'))
     runs-on: ubuntu-latest
     steps:
       - name: inject slug/short variables
@@ -57,6 +54,27 @@ jobs:
         env:
           NODE_ENV: ${{ secrets.NODE_ENV }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - uses: actions/setup-python@v2
+        with:
+          python-version: "3.x"
+      - uses: r-lib/actions/setup-pandoc@v1
+        with:
+          pandoc-version: "2.14.1"
+      - name: Install wkhtmltopdf
+        run: |
+          curl -L https://github.com/wkhtmltopdf/packaging/releases/download/0.12.6-1/wkhtmltox_0.12.6-1.focal_amd64.deb > wkhtmltopdf.deb
+          sudo apt-get install ./wkhtmltopdf.deb
+      - name: Install Python dependencies
+        run: pip install -r requirements-ci.txt
+
+      - name: Install PDF Node dependencies
+        working-directory: ./scripts/pdf
+        run: |
+          npm install
+
+      - name: Build all pdfs
+        run: npm run pdf:build-all-ci
 
       - name: Checking Gatsby cache
         id: gatsby-cache-build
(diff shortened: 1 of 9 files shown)


drothery-edb:
I did some spot checking. The tables look good. The annotations are great. The code blocks look great. The only thing that looked a little off to me was the body font size (appears quite small) and the vertical spacing (lots of vertical white space).

drothery-edb (approved):

PR #3218, opened 2022-10-05
Fixing broken links 4

ccestes:
## What Changed?

As per #3166, fixing broken links (second pass)

### DIFF
--- a/product_docs/docs/pgd/3.7/bdr/configuration.mdx
+++ b/product_docs/docs/pgd/3.7/bdr/configuration.mdx
@@ -37,7 +37,7 @@ which vary according to the size and scale of the cluster.
 -   `max_replication_slots` - Same as `max_wal_senders`.
 -   `wal_sender_timeout` and `wal_receiver_timeout` - Determine how
     quickly an origin considers its CAMO partner as disconnected or
-    reconnected; see [CAMO Failure Scenarios](../../latest/bdr/camo/#failure-scenarios) for
+    reconnected; see [CAMO Failure Scenarios](/pgd/latest/bdr/camo/#failure-scenarios) for
     details.
 
 Note that in normal running for a group with N peer nodes, BDR will require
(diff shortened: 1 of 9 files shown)


ccestes:
Hi @josh-heyer -- This is the last round of link fixes according to #3166. There are a few issues that may need your input:

- The [PGD v3.7 Commit At Most Once (CAMO)](https://www.enterprisedb.com/docs/pgd/3.7/bdr/camo/) topic links to the v4 topic, which creates broken links in v3.7. @drothery-edb said that we can now add the content to the v3.7 topic, allowing us to link within the version. Are you able to do that?
- Links from this [PEM topic](https://www.enterprisedb.com/docs/pem/latest/considerations/installing_pem_server_and_apache_web_server_preferences/) appear broken despite working in both my local build and on the site. Additionally, the output displays the paths differently than what they are. For example, the file path /pem/latest/installing_pem_server/pem_server_inst_linux/installing_pem_server_using_edb_repository/ appears in the warning as /pem/8/using_edb_repository. This is the same for all the links on the topic. Any ideas what's going on?
- There are a few other stray broken links that appear to work and don't switch between versions. I'm not sure what's wrong with them, but they are low priority as there are only a few spread out across the entire docs.

josh-heyer:
The CAMO thing is... Not ideal. Until / unless we can do a full import of the PGD 3.7 docs, pointing to the protected site is likely the best option here.

Regarding PEM: funny story... There are two PEM paths named "considerations" 

1. https://github.com/EnterpriseDB/docs/blob/97b12ad863139965909b110c20219b6a29d70d17/product_docs/docs/pem/8/considerations/index.mdx
2. https://github.com/EnterpriseDB/docs/blob/992df3736e68395f74a887e22ea71e2425ebd56f/product_docs/docs/pem/8/considerations.mdx

Due to how we construct URL paths from the filesystem, both of those map to the same URL: `/pem/8/considerations/` - one of them must "win", and that ends up being the first one. Which means your edits (and @ebgitelman 's edits from July, and probably some of @nidhibhammar 's past edits as well) are all for naught - that file simply isn't visible on the website.

We should reconcile the differences between these two files, and then remove `considerations.mdx`

drothery-edb:
@ccestes Nidhi is looking at the PEM issues.

PR #3220, opened 2022-10-05
Extensions: note mapping BA db names to product names

drothery-edb:
## What Changed?

Added BA note to extensions topic


drothery-edb:
Approved by Natalia through Slack

PR #3222, opened 2022-10-06
MTK: included info about moving file for EDB JDBC

drothery-edb:
## What Changed?

- Added headers to the Installing JDBC driver topic
- Added note about EDB JDBC if previously installed


mlewandowski-edb (approved):
The changes look good and can be published.

PR #3223, opened 2022-10-06
EPAS: Module name update

ccestes:
## What Changed?

Updated EPAS content as per [DF-311](https://enterprisedb.atlassian.net/browse/DF-311)


RonPrivett (approved):
Update looks good, thanks!

PR #3232, opened 2022-10-07
docs(UPM-9010): Remove the SSO note for superset

vishalsawale9:
## What Changed?

Removing the SSO auth mechanism note between BigAnimal and Superset as it is no longer valid after the new PKCE flow implementation. 

## Checklist

**Content**


moiznalwalla (approved):
LGTM. If SSO is no longer valid do we need to add steps for the new authentication method?

vishalsawale9:
> LGTM. If SSO is no longer valid do we need to add steps for the new authentication method?

The two portals, although, shares the same auth0 instance for authentication are now hosted on separate auth0 client apps, which means, it doesn't get logout if the user logouts from the other one. And we already have a line on the same page, which does tell us how to access the Superset from BA portal.

PR #3248, opened 2022-10-11
Docs/biganimal/emphasize private endpoints

drothery-edb:
## What Changed?

Prototyping a different structure to emphasize the best practice of using private endpoints


drothery-edb:
Reimplemented in https://github.com/EnterpriseDB/docs/pull/3250 sans the move of the connecting from Azure and AWS  topics to under creating

PR #3257, opened 2022-10-13
fix incorrect path for language pack

josh-heyer:
## What Changed?

Pretty sure this was a typo @ccestes - double-check?


ccestes (approved):

ccestes:
> ## What Changed?
> Pretty sure this was a typo @ccestes - double-check?

Yep! Good catch

PR #3259, opened 2022-10-14
Adjust the example layout

tureba:
The --layout parameter is case sensitive, and only accepts lower case values. So the "Silver" in the example causes it to fail, issuing a message to the user. 

## What Changed?

This change does s/Silver/silver/, which allows the example to be used as is intended.

## Checklist

**Content**


drothery-edb:
@tureba thanks for submitting this correction. I'll go ahead and merge it so we can get it into the next doc build.

PR #3260, opened 2022-10-17
Updated the Upgrade EPAS considerations

nidhibhammar:
Addressing feedback provided in the Slack thread: https://edb.slack.com/archives/CDV042H1D/p1665642577945449

### DIFF
--- a/product_docs/docs/epas/14/epas_upgrade_guide/03_limitations.mdx
+++ b/product_docs/docs/epas/14/epas_upgrade_guide/03_limitations.mdx
@@ -6,7 +6,7 @@ Consider the following when upgrading EDB Postgres Advanced Server:
 
 
 -   The `pg_upgrade` utility cannot upgrade a partitioned table if a foreign key refers to the partitioned table.
--   If you are upgrading from a lower version of EDB Postgres Advanced Server, and you use partitioned tables that include a `SUBPARTITION BY` clause, you must use `pg_dump` and `pg_restore` to upgrade an existing EDB Postgres Advanced Server installation to a later version of EDB Postgres Advanced Server. To upgrade, you must:
+-   If you are upgrading from the version 9.4 server or a lower version of EDB Postgres Advanced Server, and you use partitioned tables that include a `SUBPARTITION BY` clause, you must use `pg_dump` and `pg_restore` to upgrade an existing EDB Postgres Advanced Server installation to a later version of EDB Postgres Advanced Server. To upgrade, you must:
     1.  Use `pg_dump` to preserve the content of the subpartitioned table.
     2.  Drop the table from the EDB Postgres Advanced Server 9.4 database or a lower version of EDB Postgres Advanced Server database.
     3.  Use `pg_upgrade` to upgrade the rest of the EDB Postgres Advanced Server database to a more recent version.
--- a/product_docs/docs/epas/14/epas_upgrade_guide/03_limitations.mdx
+++ b/product_docs/docs/epas/14/epas_upgrade_guide/03_limitations.mdx
@@ -2,7 +2,7 @@
 title: "Upgrade considerations"
 ---
 
-The following apply to upgrading EDB Postgres Advanced Server:
+Consider the following when upgrading EDB Postgres Advanced Server:
 
 
 -   The `pg_upgrade` utility cannot upgrade a partitioned table if a foreign key refers to the partitioned table.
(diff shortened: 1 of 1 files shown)


product_docs/docs/epas/14/epas_upgrade_guide/03_limitations.mdx
```diff
@@ -2,12 +2,14 @@
 title: "Upgrade considerations"
 ---
 
-The following limitations apply to upgrading EDB Postgres Advanced Server:
+The following apply to upgrading EDB Postgres Advanced Server:
```
drothery-edb, line 5:
```suggestion
Consider the following when upgrading EDB Postgres Advanced Server:
```

product_docs/docs/epas/14/epas_upgrade_guide/03_limitations.mdx
```diff
@@ -2,12 +2,14 @@
...
+
 
 -   The `pg_upgrade` utility cannot upgrade a partitioned table if a foreign key refers to the partitioned table.
--   If you are upgrading from the version 9.4 server or a lower version of EDB Postgres Advanced Server, and you use partitioned tables that include a `SUBPARTITION BY` clause, you must use `pg_dump` and `pg_restore` to upgrade an existing EDB Postgres Advanced Server installation to a later version of EDB Postgres Advanced Server. To upgrade, you must:
+-   If you are upgrading from a lower version of EDB Postgres Advanced Server, and you use partitioned tables that include a `SUBPARTITION BY` clause, you must use `pg_dump` and `pg_restore` to upgrade an existing EDB Postgres Advanced Server installation to a later version of EDB Postgres Advanced Server. To upgrade, you must:
```
drothery-edb, line 9: I would think we would either remove the bullet entirely or leave it. I think the question is it likely someone is upgrading from 9.4 or earlier at this point. We should ask a customer facing SME like Adam that question.

product_docs/docs/epas/14/epas_upgrade_guide/03_limitations.mdx
```diff
@@ -2,12 +2,14 @@
...
     1.  Use `pg_dump` to preserve the content of the subpartitioned table.
     2.  Drop the table from the EDB Postgres Advanced Server 9.4 database or a lower version of EDB Postgres Advanced Server database.
     3.  Use `pg_upgrade` to upgrade the rest of the EDB Postgres Advanced Server database to a more recent version.
     4.  Use `pg_restore` to restore the subpartitioned table to the latest upgraded EDB Postgres Advanced Server database.
 -   If you perform an upgrade of the EDB Postgres Advanced Server installation, you must rebuild any hash-partitioned table on the upgraded server.
+-   If you are using an ODBC, JDBC, OCI, or .NET driver to connect to your database applications and upgrading to a new major version of EDB Postgres Advanced Server, upgrade your driver to the latest version after or before upgrading EDB Postgres Advanced Server.
```
drothery-edb, line 15:
```suggestion
-   If you are using an ODBC, JDBC, OCI, or .NET driver to connect to your database applications and upgrading to a new major version of EDB Postgres Advanced Server, upgrade your driver to the latest version when upgrading EDB Postgres Advanced Server.
```

product_docs/docs/epas/14/epas_upgrade_guide/03_limitations.mdx
nidhibhammar: @theadamwright, let us know your thoughts/suggestions here.

theadamwright: I'm not clear on the issue here, a dump/restore is required for partitioned table with SUBPARTITION BY, even 10-13?

drothery-edb: In this PR, Nidhi was adding an item to the Upgrade considerations list. When we looked at the existing topic, we noticed the bullet, that as written, was specific to customers upgrading from 9.4 or earlier. Our question is do we need to keep the bullet?  Is anyone likely to be upgrading from 9.4 or earlier?

drothery-edb: @theadamwright we think the text of the bullet we are questioning is probably still correct, but we were wondering if we needed to provide information like this that is specific to users upgrading from 9.4 or earlier.  Our question is, is it safe to assume that no one would be upgrading from 9.4 or earlier at this point? If it is safe to assume all our users are on epas 9.5 or later, we would remove the bullet. If it is not safe to assume that, we would retain the bullet and include that it only applies to 9.4 or earlier upgrades. We are planning to go with the latter option for now. We can revisit when we should remove notes specific to older versions of epas later.

drothery-edb (approved):

PR #3265, opened 2022-10-18
Update 01_understanding_qotas_in_azure.mdx

jingjingliu20:
## What Changed?

- Correct the error of not all the clusters require ESv3 CPU cores, this is depending on instance and cluster type.
- BigAnimal has more cluster type, e.g. BDR, HA with 1 replica, HA with 2 replicas. So we need to be careful about the doc only covering single node and HA which is not true now.
- Highlight why we need 14 Dv4 cores, 6 of them are only required during AKS upgrade. 
- About the recommended IP quota, usually customer doesn't need to create 50 clusters, so there is no need to set the IP limit as 50.

## Checklist

**Content**

- [X] This PR adds new content

### DIFF
--- a/product_docs/docs/biganimal/release/getting_started/preparing_cloud_account/01_preparing_azure/01_understanding_qotas_in_azure.mdx
+++ b/product_docs/docs/biganimal/release/getting_started/preparing_cloud_account/01_preparing_azure/01_understanding_qotas_in_azure.mdx
@@ -31,13 +31,9 @@ To prevent failures while creating your clusters, ensure that each of the follow
 
 Every BigAnimal cluster with public network access is assigned a single public IP address, and this IP address counts against the quota for both basic and standard IP address types in a region. BigAnimal can't create more clusters if the IP address limit is reached.
 
-### Default Azure limit
-
-The default public IP addresses limits for basic and standard type are set to 10. See [Public IP address limits](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/azure-subscription-service-limits#publicip-address) for more information.
-
 ### Recommended limit
 
-This is decided by the number of the clusters you will create. 
+The default public IP addresses limits for basic and standard type are set to 10. See [Public IP address limits](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/azure-subscription-service-limits#publicip-address) for more information. If you need more than 10 clusters, increase the limit to the number of clusters you plan to deploy plus current usage.
 
 ## vCPU limits
 
@@ -51,7 +47,7 @@ The number of cores required by the database cluster depends on the instance typ
 
 BigAnimal requires an additional eight Dv4 virtual machine cores per region for management resources. 
 
-As the AKS cluster involves performing periodic upgrades to the proper Kubernetes version, an additional six Dv4 virtual machine cores per region are needed for AKS upgrade.
+BigAnimal requires an additional six Dv4 virtual machine cores per region for periodic maintenance upgrades.
 
 ### Recommended limits
 
@@ -59,4 +55,4 @@ BigAnimal recommends the following per region when requesting virtual machine re
 
 -   Total Regional vCPUs: minimum of 50 per designated region
 -   Standard Dv4 Family vCPUs: minimum of 14 per designated region
--   Other Family vCPUs: depending on the instance type, cluster type and number of clusters.
+-   Other Family vCPUs: depending on the instance type, cluster type, and number of clusters.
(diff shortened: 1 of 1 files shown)


product_docs/docs/biganimal/release/getting_started/preparing_cloud_account/01_preparing_azure/01_understanding_qotas_in_azure.mdx
```diff
@@ -41,30 +37,26 @@ The default public IP addresses limits for basic and standard type are set to 10
 
 ### Recommended limit
 
-EDB recommends that you increase the public IP addresses limits for Basic and Standard type to 50. If you need more than 50 clusters, increase the limit to the number of clusters you plan to deploy.
+This is decided by the number of the clusters you will create.
```
drothery-edb, line 40: Instead of keeping this section, how about we add the following sentence after the end of the previous paragraph: 

>If you need more than 10 clusters, increase the limit to the number of clusters you plan to deploy.

If we do that I think we can drop the "Default Azure limit" heading.

product_docs/docs/biganimal/release/getting_started/preparing_cloud_account/01_preparing_azure/01_understanding_qotas_in_azure.mdx
```diff
@@ -41,30 +37,26 @@ The default public IP addresses limits for basic and standard type are set to 10
...
-For example, if you provision the largest virtual machine E64sv3 with high availability enabled, it requires (3 \* 64) = 192 ESv3 cores per region.
-
-BigAnimal requires an additional eight Dv4 virtual machine cores per region.
+-   A cluster running on an E{N}sv3 instance with HA enabled and 2 replicas uses 3 \* {N} ESv3 cores.
 
-### Default limit
+BigAnimal requires an additional eight Dv4 virtual machine cores per region for management resources. 
 
-The default number of total vCPU (cores) per subscription per region is 20. For more information, see [Virtual Machines limits - Azure Resource Manager](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/azure-subscription-service-limits#virtual-machines-limits---azure-resource-manager).
+As the AKS cluster involves performing periodic upgrades to the proper Kubernetes version, an additional six Dv4 virtual machine cores per region are needed for AKS upgrade.
```
drothery-edb, line 54: We haven't really introduced the concept of the AKS cluster in the docs. The only place we mention AKS is under pricing and in metrics labels so it might cause some confusion. Can we say instead:

> BigAnimal requires an additional six Dv4 virtual machine cores per region for periodic maintenance upgrades.

Or, would it be possible to include the number in the previous paragraph and say:

> BigAnimal requires an additional 14 Dv4 virtual machine cores per region for management resources. 

AKS is listed under management costs on the Pricing page.

product_docs/docs/biganimal/release/getting_started/preparing_cloud_account/01_preparing_azure/01_understanding_qotas_in_azure.mdx
```diff
@@ -41,30 +37,26 @@ The default public IP addresses limits for basic and standard type are set to 10
...
 
 ### Recommended limits
 
 BigAnimal recommends the following per region when requesting virtual machine resource limit increases:
 
--   Total Regional vCPUs: minimum of 60 per designated region
--   Standard ESv3 Family vCPUs: minimum of 50 per designated region
+-   Total Regional vCPUs: minimum of 50 per designated region
 -   Standard Dv4 Family vCPUs: minimum of 14 per designated region
+-   Other Family vCPUs: depending on the instance type, cluster type and number of clusters.
```
drothery-edb, line 62:
```suggestion
-   Other Family vCPUs: depending on the instance type, cluster type, and number of clusters.
```

product_docs/docs/biganimal/release/getting_started/preparing_cloud_account/01_preparing_azure/01_understanding_qotas_in_azure.mdx
jingjingliu20: Sure, I have updated it. The new limit needs to consider the current usage. So I changed it to:

> If you need more than 10 clusters, increase the limit to the number of clusters you plan to deploy plus current usage.

product_docs/docs/biganimal/release/getting_started/preparing_cloud_account/01_preparing_azure/01_understanding_qotas_in_azure.mdx
jingjingliu20: Updated, thanks!

product_docs/docs/biganimal/release/getting_started/preparing_cloud_account/01_preparing_azure/01_understanding_qotas_in_azure.mdx
jingjingliu20: I prefer the first one as these additional six Dv4 are not always required, they are only required during upgrade.

drothery-edb (approved):

PR #3281, opened 2022-10-24
Creating consistent URL structure for FDWs

dwicinas:
Reworked installing directory, including all install topics, to implement consistent approach to file naming to result in standardized URLs


dwicinas:
@drothery-edb Ready for review and possible merging

drothery-edb (approved):

PR #3282, opened 2022-10-25
we don't recommend the user to store data in postgres database

sszhao1020:
## What Changed?

Add the statement about not storing data in postgres database
Avoid storing data in the `postgres` database, which is used as an access point to configure the cluster.

## Checklist

**Content**


drothery-edb:
@sszhao1020 do you approve of my changes? https://deploy-preview-3282--edb-docs-staging.netlify.app/docs/biganimal/latest/using_cluster/01_postgres_access/

drothery-edb (approved):

PR #3287, opened 2022-10-26
Standardized URLs for connectors

dwicinas:
Reworked installing directory, including all install topics, to implement consistent approach to file naming to result in standardized URLs


dwicinas:
@drothery-edb, I've looked this one over and I think it's ready commit. It covers JDBC, OCL, and ODBC.

PR #3297, opened 2022-10-31
integral deployment (deployment paths in frontmatter)

josh-heyer:
Adds a new way to specify deployment paths via the templates themselves. This should reduce the amount of editing we have to be doing to deploy.mjs, but more importantly provides us with a clear path to migrating products to a consistent path layout while also adding redirects. 

See examples for hadoop, mysql, mongodb and documentation in README.md


josh-heyer:
Good eye, @dwicinas - fixed the replacement expression in d0ba39aed

dwicinas:
This PR seems ready to merge to me.

PR #3316, opened 2022-11-07
PEM online help - removed duplicate topics: PEM-4623

nidhibhammar:


drothery-edb (approved):

sonotley (changes requested):
The deleted version of the config params page still appears in search and then gives a 404 if I click it (perhaps this is just a cache issue or something?)

drothery-edb:
> The deleted version of the config params page still appears in search and then gives a 404 if I click it (perhaps this is just a cache issue or something?)

@josh-heyer a deleted topic shouldn't appear in search results once it is deployed to production, correct?

josh-heyer:
Correct, @drothery-edb. Draft deploys don't generate their own search index (would slow down deployment), so you get whatever results would currently appear in production when searching on these - that means any new content doesn't show up, and removed content does (but will produce a 404 when attempting to view).

PR #3317, opened 2022-11-07
Extensions: refreshing extensions table

drothery-edb:


josh-heyer:
@drothery-edb please add `edb-docs-sourcing@extensions-sheet-importer.iam.gserviceaccount.com` as a viewer on the source spreadsheet - the source action will then work again (remove + re-add the `source_extensions` label to trigger)

PR #3324, opened 2022-11-09
Adjust indentation of a "!!! Important" section

tureba:
Currently, that section shows up literally as `!!! Important Alternatively, you can list namespaces using a label selector, as explained in` in the documentation. I believe that might be caused by the block being 1 space further in, when compared to the rest of the 3. section, so this change tries to handle that.

## What Changed?

The indentation of the `!!! Important` section went from 4 spaces to 3 spaces, to match the rest of the contents of the 3. section


drothery-edb (approved):
Thank you for making this correction.

PR #3327, opened 2022-11-09
Add Microsoft.DataProtection and Microsoft.StoragePool as prerequesites for Azure

jasonparton:
Microsoft.DataProtection and Microsoft.StoragePool needed to be added to this list in the documentation as part of the Resource Providers (aka "namespaces") that customers need to ensure are Registered prior to onboarding BigAnimal. If @valeriodelsarto wouldn't mind reviewing this change, that would be appreciated.

### DIFF
--- a/product_docs/docs/biganimal/release/getting_started/preparing_cloud_account/01_preparing_azure/01_understanding_qotas_in_azure.mdx
+++ b/product_docs/docs/biganimal/release/getting_started/preparing_cloud_account/01_preparing_azure/01_understanding_qotas_in_azure.mdx
@@ -19,8 +19,10 @@ To prevent failures while creating your clusters, ensure that each of the follow
 | Microsoft.Capacity             | Checks the Azure resource quota                                                                        |
 | Microsoft.AlertsManagement     | Monitors failure anomalies                                                                             |
 | Microsoft.ContainerService     | Manages cluster workloads run on the Azure Kubernetes Service                                          |
+| Microsoft.DataProtection       | Manages the backup instance, backup policies and recovery points                                       |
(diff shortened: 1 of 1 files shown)


drothery-edb (approved):

valeriodelsarto:
I would wait for updating this, because both Resource Providers IIUC are not required by our code and was tried to get registered due to and `azurerm` terraform provider bug (that will be fixed with https://enterprisedb.atlassian.net/browse/UPM-12936) or weird behaviour (that we are going to fix with https://enterprisedb.atlassian.net/browse/UPM-12938). Full investigation in comments here https://enterprisedb.atlassian.net/browse/UPM-12879. Do you agree?
My suggestion would be don't do anything into our docs waiting for the results/outcomes of https://enterprisedb.atlassian.net/browse/UPM-12939.

drothery-edb:
> I would wait for updating this, because both Resource Providers IIUC are not required by our code and was tried to get registered due to and `azurerm` terraform provider bug (that will be fixed with https://enterprisedb.atlassian.net/browse/UPM-12936) or weird behaviour (that we are going to fix with https://enterprisedb.atlassian.net/browse/UPM-12938). Full investigation in comments here https://enterprisedb.atlassian.net/browse/UPM-12879. Do you agree? My suggestion would be don't do anything into our docs waiting for the results/outcomes of https://enterprisedb.atlassian.net/browse/UPM-12939.

Since [UPM-12939](https://enterprisedb.atlassian.net/browse/UPM-12939) is now closed to we have an answer to whether we should make this change or not?

valeriodelsarto:
I think that this PR can be closed without merge, as I was saying in the previous comment. Do you agree @jasonparton?

PR #3364, opened 2022-11-21
Update index.mdx

scottatchue:
The title for PgBouncer and Pgpool both had double "with" in it.

## What Changed?

deleted double "with" in heading in two places.


drothery-edb (approved):
Thanks for fixing this, Scott!

PR #3371, opened 2022-11-22
standarized URLs for PostGIS

dwicinas:
Created standardized "installing" directory, renaming all files as necessary and creating required redirects.


dwicinas:
@drothery-edb, this one should be ready to merge

drothery-edb (approved):

PR #3372, opened 2022-11-22
Warn about synchronous replication issues with multiple bdr databases

irionr:
BDR-2852

## What Changed?

BDR3.7 and BDR4 docs related to synchronous replication.

### DIFF
--- a/product_docs/docs/pgd/3.7/bdr/durability.mdx
+++ b/product_docs/docs/pgd/3.7/bdr/durability.mdx
@@ -44,6 +44,13 @@ Replication](https://www.postgresql.org/docs/11/warm-standby.html#SYNCHRONOUS-RE
 (PSR), which is uni-directional, but offers a synchronous variant that
 can used in combination with BDR.
 
+!!! WARNING
+	This only works when using a single database per node. When using multiple
+	BDR enabled databases per node, which is not generally recommended, the
+	LSN based confirmations may originate from any one of the databases from
+	a node specified in `synchronous_standby_names` and thus not assure the
+	data is really flushed to disk.
+
 This chapter covers the various forms of synchronous or eager
 replication and its timing aspects.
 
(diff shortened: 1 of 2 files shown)


product_docs/docs/pgd/3.7/bdr/durability.mdx
```diff
@@ -44,6 +44,12 @@ Replication](https://www.postgresql.org/docs/11/warm-standby.html#SYNCHRONOUS-RE
 (PSR), which is uni-directional, but offers a synchronous variant that
 can used in combination with BDR.
 
+!!! WARNING
+    While using multiple BDR enabled databases (we advise against it)
+    the lsn sync confirmation may come from a diferent database on a
+    node that is in our synchronous_standby_names and thus not assure
+    that the data is really synced on the database that we really want.
+
```
mwanner, line 52: This is not as clear as it should be, IMO. I also think we should avoid "we" in documentation, but use third person.

How about this:
This only works when using a single database per node. When using multiple BDR enabled databases per node, which is not generally recommended, the LSN based confirmations may originate from any one of the databases from a node specified in `synchronous_standby_names` and thus not assure the data is really flushed to disk.

mwanner (changes requested):
I proposed slightly different wording. However, writing that, I wonder whether it's actually correct.

Does an LSN confirmation not apply for all databases per node? I.e. a flush LSN confirms everything up to that LSN had been flushed - independent of the database or for all databases (on a node). Am I missing something? Where's the issue here?

product_docs/docs/pgd/3.7/bdr/durability.mdx
irionr: Done, thanks! 

@drothery-edb could you please check/merge it?

drothery-edb (approved):

PR #3373, opened 2022-11-22
copy edits to control structures section of SPL guide

ebgitelman:


product_docs/docs/epas/14/epas_compat_spl/05_control_structures/04_case_expression/02_searched_case_expression.mdx
```diff
@@ -18,9 +18,13 @@ CASE WHEN <boolean-expression> THEN
 END;
 ```
 
-`boolean-expression` is evaluated in the order in which it appears within the `CASE` expression. `result` is an expression that is type-compatible in the context where the `CASE` expression is used. When the first `boolean-expression` is encountered that evaluates to `TRUE`, `result` in the corresponding `THEN` clause is returned as the value of the `CASE` expression. If none of `boolean-expression` evaluates to true then `result` following `ELSE` is returned. If no `ELSE` is specified, the `CASE` expression returns null.
+-  `boolean-expression` is evaluated in the order in which it appears win the `CASE` expression.
```
nidhibhammar, line 21: win word in the phrase "appears win the `CASE` expression need to be changed

nidhibhammar (approved):

PR #3382, opened 2022-11-27
Subscriber-Only nodes are not considered in quorum

wind39:
Documentation only mentions Logical Standby nodes, but we also need to mention Subscriber-Only nodes.


drothery-edb (approved):
@PJMODOS do you want anyone on the dev team to review this PR?

mwanner (approved):
LGTM and is technically correct.

PR #3396, opened 2022-11-30
Remove suggestion about harpctl unmanage node

wind39:
`harpctl unmanage node <node_name>` doesn't exist. If you try that, the command will fail.

Per [documentation](https://www.enterprisedb.com/docs/pgd/latest/harp/08_harpctl/#harpctl-unmanage), `harpctl unmanage` works only for `cluster` at the moment.


EFM-Bobby:
> @dbgreen @EFM-Bobby Can you please confirm this is correct?

If the command is failing, then it sounds like it's not supported. I only see references to unmanaging the cluster in this:
https://github.com/EnterpriseDB/harp/blob/master/internal/cmd/unmanage.go

mwanner (approved):
Let's get this in. Better to remove it.

PR #3420, opened 2022-12-07
Update azure_ad.mdx

sszhao1020:
## What Changed?
The assignment required choose no means all the users in AAD can login BA and use BA. So I removed don't


jasonparton (approved):
Reviewed & approved.  Thank you!

PR #3442, opened 2022-12-13
Extensions: fix table formatting

drothery-edb:
## What Changed?

Trying to fix issues caused by yellow background color in the source doc. Removed the color in the Doc Preview, but the source_extension step isn't finding differences in the file and is erroring out.

https://docs.google.com/spreadsheets/d/1GXzzVYT6CULGgGcyp0VtBfOtbxuWxkOU2pRYW42W4pM/edit#gid=1884264748

### DIFF
--- a/advocacy_docs/pg_extensions/index.mdx
+++ b/advocacy_docs/pg_extensions/index.mdx
@@ -1484,4 +1484,3 @@ Categories of extensions:
 </table>
 
 
-temp change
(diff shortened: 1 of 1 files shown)


advocacy_docs/pg_extensions/index.mdx
```diff
@@ -1484,4 +1484,4 @@ Categories of extensions:
 </table>
 
 
-
+temp change
```
drothery-edb, line 1487:
```suggestion

```

PR #3452, opened 2022-12-15
modules-to-resources

nataliawojcik27:
Changed 'modules' to 'resources' - use of modules is incorrect terminology since that is what the customer makes not what we provide


drothery-edb:
Closed as a dup of [3453](https://github.com/EnterpriseDB/docs/pull/3453). @nataliawojcik27 Jason beat you to this change although in his PR and changed "modules" to "resources and data sources".  I assume you are find with his change. If not, let me know.

PR #3462, opened 2022-12-17
Modify Replica section

sandeeparoraedb:
I changed couple of lines however looks like the whole section needs a relook. Most of the points are referring to the Create Replica options. Like we should be saying - customer can modify these settings under the replica...then then provide a list like Modify Cluster Page - https://www.enterprisedb.com/docs/biganimal/latest/using_cluster/03_modifying_your_cluster/


drothery-edb:
@sandeep-edb @jasonparton and I decided not to implement this change. Reach out to Jason if you have concerns.

PR #3463, opened 2022-12-18
Minor corrections for CAMO and Group Commit chapters

cwcaro:
Correct grammatical errors

Correct when SET LOCAL bdr.commit_scope and be assigned.

Use boolean type spelling. Seems most consistent with BDR function signatures for Node Management and System Functions

Emphasize that Group Commit must be set before transaction has written data.

BDR-2805


cwcaro:
@mwanner some of the more egregious grammatical errors in CAMO chapter were corrected in earlier revisions.

PR #3466, opened 2022-12-19
Section - "Modify a faraway replica"

sandeeparoraedb:
## What Changed?
Removed text for the region related changes, Region is not modifiable.


drothery-edb (approved):

drothery-edb:
Thanks for pointing out the error, @sandeeparoraedb. We appreciate the help.

PR #3468, opened 2022-12-20
projects-to-Projects

nataliawojcik27:
capitlizing the P


nataliawojcik27:
@drothery-edb we should capitalize the P in Projects in the docs

drothery-edb:
I'd rather we don't make this change. Capitalizing terms like this gets messy. Even though it is a construct, it isn't really a proper noun and we use the term generically a lot. I think it is ok to have that there is an inconsistency between the docs and the UI, but I actually think we should change it (and Organization) to lowercase in the UI. Note that the [MongoDB docs](https://www.mongodb.com/docs/atlas/reference/user-roles/) don't capitalize organization or projects.

PR #3479, opened 2022-12-25
Loadbalancer example

sandeeparoraedb:
The example was showing K8 cluster, so changed it and updated to Load balancer example. Please review it with the dev/product team


jingjingliu20 (commented):
@sandeeparoraedb `service.k8s.aws/stack` is the tag of the load balancer. Actually I think we can remove this step and mention the load balancer name at the next step:

Identify the load balancer that is tagged with the ID of the cluster to which you want to connect(`<cluster-id>-rw-internal-lb`). For example, `p-96fh28m3cb-rw-internal-lb`. Note the name of that network load balancer.

Does it make sense?

sandeeparoraedb:
@jingjingliu20 -yes, makes sense to look at the name of the resource and figure out rather than tags. I completely ignored the "tagged" word and just looked at the name of the resource.

drothery-edb:
I incorporated your suggestion @jingjingliu20. Is this ready to merge now?

jingjingliu20:
@drothery-edb Yes, it's ready to merge, thanks!

PR #3485, opened 2022-12-29
Minor consistency for calls to the biganimal CLI in free trial docs

### LINKED ISSUE #3054: Feedback on /biganimal/release/free_trial/detail/experiment/backup_and_restore.mdx
Hiya,

Some of command line syntax is not consistent. I know these options relay on autocomplete option for command line terminal autocomplete option. However, I'd recommend they should be consistent. Either with ./biganimal or simply biganimal. I hope this helps.

./biganimal delete-cluster --id p-xxxxxxxxxxx
biganimal restore-cluster --id p-xxxxxxxxxxx --from-deleted --restore-point 2022-03-31T02:39:40+0000
./biganimal show-clusters
biganimal show-cluster-connection --id  p-yyyyyyyyyy

josh-heyer:
## What Changed?

Use unprefixed calls (assumes `biganimal` is on the path) everywhere for consistency. Also avoid old, camelCase flags.

Fixes #3054

### DIFF
--- a/product_docs/docs/biganimal/release/free_trial/detail/create_a_cluster/create_cluster_cli.mdx
+++ b/product_docs/docs/biganimal/release/free_trial/detail/create_a_cluster/create_cluster_cli.mdx
@@ -90,7 +90,7 @@ highAvailability: false
 Use the config file to create a new cluster:
 
 ```shell
-biganimal create-cluster --clusterConfigFile create_cluster.yaml
+biganimal create-cluster --cluster-config-file create_cluster.yaml
 __OUTPUT__
 Are you sure you want to create cluster "test_cluster"? [y|n]: y
 Create Cluster operation is started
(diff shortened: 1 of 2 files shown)


drothery-edb:
@jasonparton we use the prefixed BA calls in the Connecting your cloud topics but not in the Using the CLI topic. Josh is standardizing on not using in the Free trial section. Are you ok with us standardizing on not using it in the rest of the docs? Do you think it is save to assume that users have it in their path? Should we add a note about adding it to your path in the Using the CLI topic?

josh-heyer:
A note to that effect already exists in that topic; see also #3484

drothery-edb:
> A note to that effect already exists in that topic; see also #3484

I'm wondering if we should have info on adding biganimal to your path in the topics in the other part of the content (outside the free trial section).

drothery-edb (approved):

PR #3492, opened 2022-12-30
ASP edits

ccestes:
## What Changed?

Making suggested edits to ASP docs from Simon Riggs. Addresses the straight-forward comments. The more complex comments will be addressed in a later PR.

This work is being tracked in https://enterprisedb.atlassian.net/browse/ASP-17.

### DIFF
--- a/advocacy_docs/pg_extensions/advanced_storage_pack/using.mdx
+++ b/advocacy_docs/pg_extensions/advanced_storage_pack/using.mdx
@@ -25,8 +25,8 @@ This example is for an IoT thermostat that reports house temperatures and temper
 
 ```sql
 CREATE TABLE iot (
-    thermostat_id        bigint,
-    recordtime           timestamp,
+    thermostat_id         NOT NULL,
+    recordtime            NOT NULL,
     measured_temperature  float 4,
     temperature_setting   float 4,
 ) USING autocluster;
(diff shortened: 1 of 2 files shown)


advocacy_docs/pg_extensions/advanced_storage_pack/using.mdx
```diff
@@ -25,8 +25,8 @@ This example is for an IoT thermostat that reports house temperatures and temper
 
 ```sql
 CREATE TABLE iot (
-    thermostat_id        bigint,
-    recordtime           timestamp,
+    thermostat_id         NOT NULL,
```
markdilger, line 28: Should be
  thermostat_id bigint NOT NULL,

advocacy_docs/pg_extensions/advanced_storage_pack/using.mdx
```diff
@@ -61,7 +61,7 @@ INSERT INTO iot (thermostat_id, recordtime) VALUES (5678, 03:44);
 When you select the data from the IoT table, you can see from the ctid location that the data with the same `thermostat_id` was clustered together:
 
 ```sql
-SELECT ctid, thermostat_id, recordtime FROM iot;
+SELECT ctid, thermostat_id, recordtime FROM iot ORDER BY CTID;
```
markdilger, line 64: The example output below needs to  actually be ordered by the CTID column, but don't just change the ctid column, actually move the full rows up and down until they are ordered by that column.  (This is because, if you don't move the full rows, then rows where thermostat_id = 456 won't show next to each other.)

Issue #3457, opened 2022-12-15
Feedback on /epas/14/epas_guide/03_database_administration/01_configuration_parameters/02_summary_of_configuration_parameters.mdx


drothery-edb:
From @lucianobotti:
Customer shared a link from EPAS documentation about SSL, which from my tests and the PG documentation seems wrong:https://www.enterprisedb.com/docs/epas/latest/epas_guide/03_database_administration/01_configuration_parameters/02_summary_of_configuration_parameters/#:~:text=ssl,server%20certificate%20file.

EPAS Docs note that a restart is needed while PGDOCs mentioned only a reload :
The server reads these files at server start and whenever the server configuration is reloaded. On Windows systems, they are also re-read whenever a new backend process is spawned for a new client connection.
https://www.postgresql.org/docs/current/ssl-tcp.html#SSL-CERTIFICATE-CREATION:~:text=The%20server%20reads,the%20server%20log.
Also pg_settings shows sighup :
edb=# select name,context from pg_settings where name ~ 'ssl';
         name         | context  
----------------------------------------+----------
 ssl                  | sighup
 ssl_ca_file              | sighup
 ssl_cert_file             | sighup
 ssl_ciphers              | sighup
 ssl_crl_dir              | sighup
 ssl_crl_file              | sighup
 ssl_dh_params_file           | sighup
 ssl_ecdh_curve             | sighup
 ssl_key_file              | sighup
 ssl_library              | internal
 ssl_max_protocol_version        | sighup
 ssl_min_protocol_version        | sighup
 ssl_passphrase_command         | sighup
 ssl_passphrase_command_supports_reload | sighup
 ssl_prefer_server_ciphers       | sighup
(15 rows)
 
Could we please address this on the EPAS Docs?

PR #3496, opened 2022-12-30
PG 10+: SSL settings are updated by reload

tureba:
Starting with PG 10, the ssl settings are updated by a reload (and the same is true for EPAS 10):

* https://www.postgresql.org/docs/10/release-10.html
* https://www.postgresql.org/docs/10/ssl-tcp.html#SSL-SERVER-FILES

## What Changed?

This changes the context of the SSL settings to reflect the current versions


tureba:
I'm not entirely sure how to associate this PR with a past issue, but it's intended to cover issue 3457: https://github.com/EnterpriseDB/docs/issues/3457

PR #3507, opened 2023-01-04
Delete Replica section

sandeeparoraedb:
<would be good to have information on what happens to backups and if customer does not have any other replicas/clusters in this region then they should be advised on deleting the region so there is no additional cost of the management plane>


drothery-edb:
Thanks for the feedback @sandeeparoraedb. I created a GitHub issue with your suggestion. See https://github.com/EnterpriseDB/docs/issues/3517. I'm going to close the PR. PRs are typically used to submit specific changes versus suggestions for new content.

PR #3509, opened 2023-01-05
Fix typo in `allow_incremental_backups`

tureba:
The configuration name is plural (...backupS), but this one occurrence in the docs was written as singular, by accident.

I didn't adjust the table to have an extra space in that column, or it would make the diff difficult to read. But let me know if that's desirable.

## What Changed?

`allow_incremental_backup` (singular) -> `allow_incremental_backups`  (plural)


drothery-edb (approved):

drothery-edb:
Thanks for the fix!

PR #3510, opened 2023-01-05
Update Password statement for replica creation & Modification

sandeeparoraedb:
As per new workflow, the password is not required to be entered during the Faraway Replica creation, I modified the statement however please get it reviewed for sanity check and change the verbiage so it is more appropriate. Same thing applies to Modify replica section, that should not let users modify the password. password change should be done on the primary.


drothery-edb:
@sandeeparoraedb do you agree with my edits?

sandeeparoraedb:
@drothery-edb Sorry I am not able to see what you have updated, on the live page, I still see the password point.

drothery-edb:
> @drothery-edb Sorry I am not able to see what you have updated, on the live page, I still see the password point.

I haven't committed or published my changes yet. You can see them here: https://github.com/EnterpriseDB/docs/pull/3515/commits/82ac5cad8d379ab0572ca6a2cb079403b08d4db8

PR #3515, opened 2023-01-06
Update replication.mdx

jasonparton:
I agree that the FAQ provides more than the BA Docs, so I took a stab at moving some of the Faraway Replicas FAQ content over. There is still FAQ content that "could" be moved over, but truthfully, the Sales/PreSales team should be speaking to most of the remaining points during their discussions with the Customer.

### DIFF
--- a/product_docs/docs/biganimal/release/overview/replication.mdx
+++ b/product_docs/docs/biganimal/release/overview/replication.mdx
@@ -2,21 +2,22 @@
 title: "Faraway replicas"
 ---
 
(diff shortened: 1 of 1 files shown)


drothery-edb:
@jasonparton I made some copy and formatting edits. See https://github.com/EnterpriseDB/docs/pull/3515/commits/82ac5cad8d379ab0572ca6a2cb079403b08d4db8. Let me know what you think.

drothery-edb (dismissed):

PR #3532, opened 2023-01-12
BigAnimal: more detail for CLI commands for faraway replicas

drothery-edb:
## What Changed?
Included info on update, delete, and restore


edbamolvyavahare (approved):
LGTM

PR #3536, opened 2023-01-16
efm agents omit own location in .nodes file

EFM-Bobby:
Added two clarifications:
1. At startup, members need to have "at least" the membership coordinator's address in the file. The file doesn't need to have only that one address.
2. Added a node that agents don't writ their own info to the .nodes file -- the file is used for discovering other nodes and an agent doesn't have to discover itself.


product_docs/docs/efm/4/04_configuring_efm/03_cluster_members.mdx
```diff
@@ -26,9 +26,9 @@ chown efm:efm efm.nodes
...
 
-The agents update the contents of the `efm.nodes` file to match the current members of the cluster. As agents join or leave the cluster, the `efm.nodes` files on other agents are updated to reflect the current cluster membership. If you invoke the [`efm stop-cluster`](../07_using_efm_utility/#efm_stop_cluster) command, Failover Manager doesn't modify the file.
+The agents update the contents of the `efm.nodes` file to match the current members of the cluster. As agents join or leave the cluster, the `efm.nodes` files on other agents are updated to reflect the current cluster membership. If you invoke the [`efm stop-cluster`](../07_using_efm_utility/#efm_stop_cluster) command, Failover Manager doesn't modify the file. Note: an agent will not write its own information to the file as it does not need its own location to discover other members at startup.
```
drothery-edb, line 31:
```suggestion
The agents update the contents of the `efm.nodes` file to match the current members of the cluster. As agents join or leave the cluster, the `efm.nodes` files on other agents are updated to reflect the current cluster membership. If you invoke the [`efm stop-cluster`](../07_using_efm_utility/#efm_stop_cluster) command, Failover Manager doesn't modify the file. Note: an agent doesn't write its own information to the file as it does not need its own location to discover other members at startup.
```

drothery-edb (approved):

PR #3499, opened 2023-01-03
EPAS and PostGIS - codeready builder setup changes

nidhibhammar:

PR #3538, opened 2023-01-16
Install instructions: fixed for EPRS and pgPool

drothery-edb:
## What Changed?

- Removed codeready-builder instructions for EPRS - updating the instructions for codeready-builder was done in another PR (https://github.com/EnterpriseDB/docs/pull/3499), but the EPRS changes didn't get in at that time - adding them now
- Fixed package names for pgPool extensions - there is no hyphen between "as" and the as version number in the package names (see https://edb.slack.com/archives/CU5QEU5L7/p1673878714011569). Note that I verified that the change applies to all the OSs by checking Repos 2.0. I found that the pgPool Debian 11 packages weren't in the repo. I notified Hayee about that.


josh-heyer (approved):
👍

PR #3539, opened 2023-01-16
BigAnimal: Terraform updates

drothery-edb:
## What Changed?

Changed API version to v3
Moved errant See also section to the correct topic - see https://deploy-preview-3539--edb-docs-staging.netlify.app/docs/biganimal/latest/using_cluster/01_postgres_access/#see-also


jasonparton (commented):
They look good.  Thank you!

PR #3541, opened 2023-01-17
BigAnimal: PCI compliance

drothery-edb:
## What Changed?

Addresses: https://enterprisedb.atlassian.net/browse/UPM-12439


drothery-edb:
Natalia approved on the ticket.

PR #3547, opened 2023-01-18
Minor fixes to install process for Windows and MacOS

dwicinas:
Created a minimal topic for installing on Mac. That topic references and EDB blog for step-by-step details.

Incorporated command line tool information from Josh's postgresql_journey efforts.

Other small fixes.


advocacy_docs/supported-open-source/postgresql/installer/02_installing_postgresql_with_the_graphical_installation_wizard/windows.mdx
```diff
@@ -97,3 +95,21 @@ Before the setup wizard completes the PostgreSQL installation, it offers to laun
...
 Fig. 10: The installation wizard offers to Launch Stack Builder at exit
+
+## Command line tools
+
+By default, the installer does not modify the system path. If you want to use command line tools (such as `psql`), you can add PostgreSQL to the path after installation.
+
+1.  Open the System Properties control panel and select the **Advanced** tab (or run `SystemPropertiesAdvanced.exe`)
+2.  Click **Environment Variables** to open the environment variables editor.
+3.  Find the `Path` variable under the **System variables** heading and click **Edit**.
+4.  Add the path that you specified as the installation directory during installation, with `\bin` appended (by default: `c:\program files\postgresql\15\bin`, where *15* is the version of PostgreSQL that you installed).
```
drothery-edb, line 106:
```suggestion
4.  Add the path that you specified as the installation directory during installation, with `\bin` appended (by default: `c:\program files\postgresql\15\bin`, where `15` is the version of PostgreSQL that you installed).
```

drothery-edb:
@dwicinas once we get the issues on this branch straightened out, you need to add a redirect because you renamed a file.

dwicinas:
@drothery-edb I added redirects to the file you identified, as well as two other files that I renamed.

I used the link checker tool and found no broken links. I also searched for potential broken links across all products. The only borken links I found are in Josh's postgresql_journey directory. I did not fix those.

drothery-edb:
I don't think the absolute path for the redirect is correct. See my question to Josh in the docs-tech-help channel.

Also, why are you renaming the other two files? I don't think it is necessary and probably not worth the redirect overhead.

drothery-edb:
I agree that if you renamed the other two files, you should add redirects, but my question was why you renamed them in the first place? We try to limit renaming files unless there is a strong case for doing so because of the overhead that comes along with renaming files. I don't think the other two needed to be renamed to retain the proper nav order, but now we don't have to depend on the numbered prefix to set the ordering anyway. We could just use the navigation option in the front matter. We could possibly justify removing the prefix and using the navigation option instead to have cleaner URLs, but I don't think the redirect overhead is justified for just changing the prefix numbers.

drothery-edb:
Something seems amiss with your branch again. You shouldn't be getting merge conflicts.

drothery-edb:
You may have missed one of my comments. You don't have the full path needed for the redirects. You need to add `/supported-open-source` before `/postgres`

PR #3558, opened 2023-01-21
Update connecting_azure.mdx

jingjingliu20:
Prerequisites should mention Owner role requirement on Azure subscription.

## What Changed?

Add Owner role requirement on Azure subscription on the Prerequisites section.


jasonparton:
I might suggest changing the verbiage to read as follows:
"Before connecting to your cloud, ensure that you are assigned either the Global Administrator role or the Privileged Role Administrator role in Azure AD, and you have the role of Owner for the subscription being used."
We once onboarded a customer that had the role of Owner for an Azure subscription, but not for the one being onboarded.  Just a thought...not stuck on it.

jasonparton (changes requested):
I might suggest changing the verbiage to read as follows:
"Before connecting to your cloud, ensure that you are assigned either the Global Administrator role or the Privileged Role Administrator role in Azure AD, and you have the role of Owner for the subscription being used."
We once onboarded a customer that had the role of Owner for an Azure subscription, but not for the one being onboarded. Just a thought...not stuck on it.

jasonparton (approved):

PR #3562, opened 2023-01-23
PGD 5: free trial

drothery-edb:
## What Changed?

Added the section and copied the BigAnimal topic over


drothery-edb:
We will be creating a new PR for this work.

PR #3570, opened 2023-01-25
PEM: fix PDF links

drothery-edb:
## What Changed?

Links worked in online doc and PDF but link checker was throwing errors
Checking to see if changing from absolute paths to relative paths fixes it


josh-heyer:
@drothery-edb pretty sure these paths aren't the issue. We're in another situation where two files both map to the same path: 

1. product_docs/docs/pem/9/considerations/index.mdx 
2. product_docs/docs/pem/9/considerations.mdx

The errors are in `considerations.mdx` - probably want that file gone. Don't see any errors in considerations/index.mdx - no reason to modify that.

PR #3587, opened 2023-01-31
Create standardized file structure for PEM

dwicinas:


drothery-edb:
@dwicinas do we still need the epas lines in deploy.mjs? They might have creeped back in when resolving a merge conflict, but I'm guessing we don't need them.

product_docs/docs/pem/9/installing/creating_pem_repository_in_isolated_network.mdx
```diff
@@ -1,12 +1,13 @@
 ---
-title: "Creating a EDB repository on an isolated network"
+title: "Creating an EDB repository on an isolated network"
 navTitle: "Setting up local repository"
```
drothery-edb, line 3:
```suggestion
navTitle: "Setting up a local repository"
```

drothery-edb (approved):

dwicinas:
@drothery-edb I'm uncertain how to completely remove the last set of deployment instructions from deploy.mjs. Currently that last set is for EPAS 14. When I remove that block of code, I get an error during deployment. I could start just deleting additional instructions, but the safest thing might be to ask Josh, presumably in the tech docs channel?

drothery-edb (approved):

PR #3596, opened 2023-02-01
Command line parameters for Windows and Mac

dwicinas:


dwicinas:
@drothery-edb This is ready for your review. I'm not sure it needs a technical review. I used the information for PostgreSQL 15 that Nidhi provided and combined it with some of the definitions for PostgreSQL 10 found in the Wayback Machine. Nothing seems controversial. There are a couple parameter values that I don't define, but I would need some technical help to obtain those definitions. So I just left them undefined.

drothery-edb:
Would it be possible to combine the two tables and just note the differences between platforms? Or does the scope of the differences between the two make that tricky?

dwicinas:
There's two or three for windows that don't exist for Mac. And there's another 2-3 with different defaults. 
So it could be done in a single table, I think

dwicinas:
@drothery-edb I consolidated the Mac and Windows descriptions. It's ready for your review, again.

drothery-edb (approved):

PR #3618, opened 2023-02-07
Fix broken links in PEM

dwicinas:
Fixed many broken links caused by restructuring of PEM

Also moved "Prerequisites" topic to appear under "Installing" and renamed it "Linux prerequisites.


dwicinas:
@drothery-edb This should be ready for your review and merging if you think it's ready.

drothery-edb (approved):

PR #3650, opened 2023-02-13
Add release notes for PGD 3.7.20

irionr:
BDR-3181


mwanner (changes requested):
Requires the same change as requested in https://github.com/EnterpriseDB/pglogical/pull/375.

PR #3653, opened 2023-02-13
Added End of support notice for EPAS 11

mlewandowski-edb:
Added a section to notify users that Migration Portal support for EPAS 11 as a target database is planned to be removed in May;


mlewandowski-edb (commented):
Looks good

drothery-edb (approved):

PR #3674, opened 2023-02-15
BDR-3235: Update durability and consistency chapters and pull them up from BDR

PJMODOS:
## What Changed?

- moved group commit,camo,lag tracking to PGD under durability
- added separate commit scopes chapter to PGD under durability with the updated syntax
- moved coflicts/eager/clcd and crdt to PGD under consistency.

This is for BDR-3235


mwanner (approved):
LGTM

PR #3675, opened 2023-02-15
Clarify use of legacy synchronous replication with BDR

mwanner:
It seems synchronous_standby_names is not compatible with CAMO at all, not just with the CAMO partner node. Mark it as deprecated more globally and visibly.


product_docs/docs/pgd/5/choosing_durability.mdx
```diff
@@ -5,7 +5,7 @@ title: "Choosing durability"
 EDB Postgres Distributed allows you to choose from several replication configurations based on your consistency, availability, and performance needs:
 
 * Asynchronous
-* Synchronous (using `synchronous_standby_names`)
+* Synchronous (using `synchronous_standby_names`, deprecated)
```
PJMODOS, line 8: We should not deprecate this, we don't have anything that's comparable performance wise, if user only cares about durability from APP POV, and not the consistency properties, they should be free to use this no?

PJMODOS: Especially given we have no proper group commit benchmarks :)

mwanner: Yes, unless in combination with CAMO, Eager, Group Commit. So maybe "deprecated" is too strong, then.

The durability chapter does recommend Group Commit instead, though.

mwanner: Dropped this change.

Issue #3637, opened 2023-02-10
Feedback on /pem/9/pem_online_help/01_toc_pem_getting_started/04_pem_roles.mdx


FreddySo-EDB:
"The examples shown above creates a login role named acctg_clerk that will have access to the Audit Manager; "

Either this statement is incorrect, or the example in the picture above this statement is incorrect.

PR #3697, opened 2023-02-21
PEM - roles fix

nidhibhammar:
Fixes https://github.com/EnterpriseDB/docs/issues/3637


drothery-edb:
Let's remove the topic from the OLH section and add a redirect to it in the remaining topic. I think it makes more sense to remove duplicate topics when we encounter them then to continue to maintain them both. Do you agree?

drothery-edb:
Let's merge this but consider moving the Getting Started section to the main content.

drothery-edb (approved):

PR #3711, opened 2023-02-22
Use alias of the private link service

sszhao1020:
## What Changed?
1. Connect to an Azure resource by resource ID or alias. -> allows the private link service in a different account
2. update the last IP to make it consistent.


sszhao1020:
As discussed here -> https://edb-starlight.slack.com/archives/C02JYUCEJV7/p1677131956918599?thread_ts=1676307554.638649&cid=C02JYUCEJV7
We will leave the doc with the current status and close the PR hee.

PR #3742, opened 2023-02-28
add v15 async link to efm sync replication section

EFM-Bobby:
Starting with v15, a database property is available that means users may not need the efm reconfigure.sync.primary property. Am creating the PR but don't think I have the link correct (don't know the syntax) yet so will have to ask about it.


EFM-Bobby:
Closing this. Realized this applies to Postgres Distributed (which is the new name for ????) and I don't think this applies to physical streaming replication.

PR #3747, opened 2023-02-28
Example edits of epas and pem rel notes

ebgitelman:


drothery-edb:
We are developing guidelines for writing up release notes. You can see the [draft here](https://docs.google.com/document/d/1igBb4eSxJR_1eW04owVQ0rRCY8wOLyysHJIi7P6-59k/edit?usp=sharing). While working through defining the guidelines, Betsy applied them to the latest release notes for PEM and EPAS. Any concerns about the rewrites Betsy did using the guidelines? You would be giving us our initial feedback from stakeholders. Once we incorporate any feedback you may have, we will roll out the guidelines to a wider audience and make them generally available.

product_docs/docs/pem/9/pem_rel_notes/910_rel_notes.mdx
```diff
@@ -6,8 +6,8 @@ New features, enhancements, bug fixes, and other changes in PEM 9.1.0 include:
...
 | Enhancement | Added the ability to copy probes and alerts to all servers in a group without having to select them individually.                                                                                         |
-| Bug Fix     | Fixed an issue whereby alert notifications were sent for Low or Medium alerts when the user had selected not to receive them. [Support Ticket #87664]                                                     |
-| Bug Fix     | Fixed an issue whereby webhooks would not be called if there was an inactive endpoint in the spool. [Support Ticket #87365]                                                                               |
-| Other       | The PGD Worker Errors probe and associated alert template and dashboard chart have been removed as they were found to produce incorrect data. They will be replaced with a new probe in a future release. |
\ No newline at end of file
+| Bug Fix     | Fixed an issue whereby alert notifications were sent for Low or Medium alerts when the option to send them was disabled. [Support Ticket #87664]                                                     |
```
sonotley, line 11: I'm happy with this specific change, but it's a handy example for one bit of feedback I had on the guideline. Sometimes trying to avoid 'the user' might result in us making things less clear. For example 'you' would sound odd here so you've switched to the passive voice instead and I'm not sure it's an improvement. In the past when writing technical docs I tried hard to avoid the passive voice as it is nearly always clearer if you state who is doing a given action and with PEM I think we'll find that 'the user' is sometimes a different person from 'you'.

sonotley:
Happy with overall approach and guidelines. They fit with how I tend to write anyway so they seem pretty natural. I have left one comment inline.

product_docs/docs/pem/9/pem_rel_notes/910_rel_notes.mdx
ebgitelman: Agreed. Thanks for the input!

drothery-edb (approved):

PR #3751, opened 2023-03-01
Fixed confusing location name

nick-ivanov-edb:
Perhaps it's better not to use an AWS region name for a location in a Docker deployment

This might cause confusion. 

## What Changed?

Changed the TPA deployment location from "eu-west-1" (AWS region name) to a neutral "dc1" (for data centre 1).

Also fixed a typo.


josh-heyer (approved):
👍

PR #3754, opened 2023-03-01
Advanced Storage pack: changing to generic Postgres terminology

drothery-edb:
## What Changed?

We use Postgres when referring to our set of Postgres distributions: PG, EPAS, and PGE


drothery-edb:
@theadamwright we want to use the generic name for Postgres here, correct?

theadamwright:
Yes, the generic Postgres terminology should be used since ASP supports community PostgreSQL, EDB Postgres Extended Server, and EDB Postgres Advanced Server.

theadamwright (approved):

PR #3758, opened 2023-03-02
Remove account registration step from EPRS

dwicinas:
Removed the step asking for account registration information


dwicinas:
@drothery-edb Based on information from Nidhi, I removed a step from this procedure.

nidhibhammar (approved):

PR #3773, opened 2023-03-08
Update note on troubleshooting repo access to reflect reality

josh-heyer:
## What Changed?

Rewrite the (still temporary) note in TPA Installation topic to stand a chance of being useful.

## Background

Big problem: the curl command as-written will never produce an explicit error (we tell it to squelch errors!). A token with insufficient access or a mistyped token will fail silently. ...However, a good token will *succeed* noisily!

Other problem: old tokens are not the only issue here. Depending on subscription level, current customers may also encounter (silent) failures. 

Changes attempt to address both of these issues and provide some direction for those encountering them.

### DIFF
--- a/product_docs/docs/tpa/23/INSTALL.mdx
+++ b/product_docs/docs/tpa/23/INSTALL.mdx
@@ -76,9 +76,12 @@ $ curl -1sLf 'https://downloads.enterprisedb.com/<your-token>/postgres_distribut
 $ curl -1sLf 'https://downloads.enterprisedb.com/<your-token>/postgres_distributed/setup.rpm.sh' | sudo -E bash
 ```
 
-!!! Warning "Troubleshooting older tokens"
-    If the command above fails with "Access denied", you may have an access token that was generated before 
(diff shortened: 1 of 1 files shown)


josh-heyer:
Just FYI: I'm still considering this a temporary note, as I believe (hope!) this will be better addressed in the repo UI at some point. If it ends up being permanent, it should be pushed up to the TPA repo.

(current location ensures that it is temporary by default, as I must go out of my way to preserve it each time the TPA docs are synchronized)

kelpoole:
Only Extreme HA subscription customers can get TPA from legacy 2Q repos  - trial users can't do this

drothery-edb (approved):

josh-heyer:
> Only Extreme HA subscription customers can get TPA from legacy 2Q repos - trial users can't do this

This isn't touching that area, but probably a good thing to review the TPA docs for; there are still places that assume 2Q repos are the default, legacy portal access is a thing, etc. Some of this is still necessary, but should be contextualized.

PR #3781, opened 2023-03-13
Moved pgbouncer refs assuming builtin is default

djw-m:
## What Changed?

On what information we have, have moved builtin proxy to default, removed experimental warnings, removed requirement for pgbouncer binaries from docs.

BUT: Have kept references to pgbouncer as it does appear this is only a default change, so this will need PM/Tech review for correctness.

### DIFF
--- a/product_docs/docs/pgd/4/harp/04_configuration.mdx
+++ b/product_docs/docs/pgd/4/harp/04_configuration.mdx
@@ -198,7 +198,7 @@ Specify Proxy-based settings under a `proxy` heading, and include:
   state are available in status checks and other interactive events.
 
 - **`type`**: Defaults to `builtin` to use the built-in passthrough proxy. Can also be set to `pgbouncer`
-  to use the deprecated (since HARP 2.1) embedded PgBouncer support. . 
+  to use the deprecated (since HARP 2.1) embedded PgBouncer support. 
   
   * Default: `builtin`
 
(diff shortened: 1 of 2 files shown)


PJMODOS (approved):
I am not too keen about those long lines because it makes the source markdown hard to read, and personally think we should limit to somewhere around 80-100 chars per line, but otherwise it looks fine.

product_docs/docs/pgd/4/harp/04_configuration.mdx
```diff
@@ -203,16 +197,16 @@ Specify Proxy-based settings under a `proxy` heading, and include:
   Each proxy node is named to ensure any associated statistics or operating
   state are available in status checks and other interactive events.
 
-- **`type`**: Specifies whether to use pgbouncer or the experimental built-in passthrough proxy. All proxies must use the same proxy type. We recommend to experimenting with only the simple proxy in combination with the experimental BDR DCS.
-  Can be `pgbouncer` or `builtin`.
+- **`type`**: Defaults to `builtin` to use the built-in passthrough proxy. Can also be set to `pgbouncer`
+  to use the deprecated (since HARP 2.1) embedded PgBouncer support. .
```
drothery-edb, line 201:
```suggestion
  to use the deprecated (since HARP 2.1) embedded PgBouncer support. 
```

drothery-edb (approved):

PR #3792, opened 2023-03-15
Quickstart paths fixed

djw-m:
## What Changed?

Fixed paths for renamed repo

Deploying preview to check image rendering (as not working in local build)

### DIFF
--- a/product_docs/docs/postgres_for_kubernetes/1/installation_upgrade.mdx
+++ b/product_docs/docs/postgres_for_kubernetes/1/installation_upgrade.mdx
@@ -53,7 +53,7 @@ for a more comprehensive example.
     ports, as explained in the official
     [docs](https://cloud.google.com/kubernetes-engine/docs/how-to/private-clusters#add_firewall_rules)
     and by this
-    [issue](https://github.com/EnterpriseDB/cloud-native-postgres/issues/1360).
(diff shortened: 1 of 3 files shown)


josh-heyer:
For these to work, they need to be `https://raw.githubusercontent.com/cloudnative-pg/cloudnative-pg/main/docs/src/samples/`... the public repo is cloudnative-pg in the cloudnative-pg org. The internal repo *is* `cloud-native-postgres`, but raw links to these won't work even if you have access to the repo (which most readers will not).

These changes will also need to be made here: https://github.com/EnterpriseDB/cloud-native-postgres/blob/main/docs/src/quickstart.md (next release this file will be imported, similar to how we do TPA docs)

The twist here is that *most* of those files are also based on the cloudnative-pg ones... With some intermediary search-and-replace for repo name. That's gonna cause a few other broken links, as seen on these pages:

- https://github.com/EnterpriseDB/cloud-native-postgres/blob/36573bb7b1d41d236bdc9270c990046cbb2299ed/docs/src/monitoring.md
- https://github.com/EnterpriseDB/cloud-native-postgres/blob/729b53496347960b8e0758ba8bcc88170d502b3f/docs/src/security.md
- https://github.com/EnterpriseDB/cloud-native-postgres/blob/968c3f0f83c79102a673b7d35835e2bf1bc29109/docs/src/installation_upgrade.md

drothery-edb (approved):

PR #3799, opened 2023-03-16
Updates to profiling workloads content in PEM and EPAS

drothery-edb:
## What Changed?

Updating SQL Profiler and Index Adivisor content in both PEM and EPAS to reflect that these are both extensions now

https://enterprisedb.atlassian.net/browse/DF-324


tusharahuja (approved):

tusharahuja (approved):
looks good to me. Thanks Niddhi

PR #3803, opened 2023-03-17
Fix broken links

smaher-edb:
## What Changed?

- Fix broken links

### DIFF
--- a/product_docs/docs/pgd/5/upgrades/index.mdx
+++ b/product_docs/docs/pgd/5/upgrades/index.mdx
@@ -189,7 +189,7 @@ that the upgraded node is working as expected.
 ## Moving from HARP to PGD-Proxy
 
 HARP can for a time coexist with the new
-[Connection management](/product_docs/docs/pgd/5/routing) configuration.
+[Connection management](../routing) configuration.
 
 This means you can upgrade whole pre-5 cluster to PGD 5 cluster, setup
 the connection routing, replace the HARP Proxy with PGD-Proxy, move application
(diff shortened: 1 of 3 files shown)


smaher-edb:
Looks like the last time's fix didn't work. Updating again. Not sure whether there is some way to test these links before merging this PR.

product_docs/docs/pgd/5/upgrades/index.mdx
```diff
@@ -189,7 +189,7 @@ that the upgraded node is working as expected.
 ## Moving from HARP to PGD-Proxy
 
 HARP can for a time coexist with the new
-[Connection management](/product_docs/docs/pgd/5/routing) configuration.
+[Connection management](../routing) configuration.
```
smaher-edb, line 192: @drothery-edb just wanted to confirm, is there any guidelines around how these link should be

For example, if not wrong, in this case, I guess both of these will work fine but I guess first one will create problem if we move this as it is in v6 doc?

- `[Connection management](/pgd/5/routing)`
- `[Connection management](../routing)`

drothery-edb: The form we can use if we want the link to go to the latest version is: [Connection management](/pgd/latest/routing)

drothery-edb: I personally find that format the easiest.

smaher-edb: Yes, I mean in this case, we don't want it to go to the latest...but with whatever version user has selected currently in browser (because they are mostly navigating from 1 page to another and mostly they want to stay on same version).

drothery-edb: I added the deploy label. It kicks off a draft deployment. It takes a while though. I kind test the links in my local build if you like.

smaher-edb: ok, thanks.

drothery-edb (approved):

smaher-edb:
Great, thanks. I will merge this PR then.

PR #3815, opened 2023-03-17
Update 08_harpctl.mdx

nick-ivanov-edb:
Descriptions for `harpctl get location` and `harpctl get leader` erroneously state that the location name is optional and will be determined automatically if omitted. In fact both commands fail if the location is omitted.

## What Changed?

Descriptions for `harpctl get location` and `harpctl get leader`.

### DIFF
--- a/product_docs/docs/pgd/4/harp/08_harpctl.mdx
+++ b/product_docs/docs/pgd/4/harp/08_harpctl.mdx
@@ -177,8 +177,7 @@ mycluster true
 ### `harpctl get leader`
 
 Fetches node information for the current lead master stored in the DCS for the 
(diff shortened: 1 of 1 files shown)


djw-m (approved):
Changes confirmed (checked source) and no automatic location detection done. Long descrioptions:

get leader - Returns details of a locations leader node. The location needs to be specified on the command line.
	For example: 
	harpctl get leader fenway

get location: Returns information about a location. The location may be specified either in the config file or on the command line.
	For example: 
	harpctl get location northamerica
	harpctl get location

Also no need to notify devs as this us bringing the content into line with actuality.

PR #3821, opened 2023-03-20
Fix up admonition icons at CSS level

djw-m:
## What Changed?

Brings the icons back to admonitions. 

EVERYTHING WILL NEED REVIEWING BEFORE THIS MERGES....


josh-heyer (approved):
Looks much better, more distinct

PR #3830, opened 2023-03-21
Small link fix to correctly nav to the SQL guide

djw-m:
## What Changed?

Link fix.


ebgitelman (commented):
Okay, but this would work as well, and I think Dee Dee prefers this style:
/epas/latest/epas_compat_sql/

and it would probably be more useful to go right to the page 

/epas/latest/epas_compat_sql/39_create_type/

Which I should have done in the first place :-)

PR #3840, opened 2023-03-23
Support admonition icons, all types in dark mode

josh-heyer:
## What Changed?

We were missing a couple of admonition types in dark mode, causing severe retina burns.
Also, we should allow our diversity of icons to... shine through


djw-m (approved):
Looks good, rolled up dark mode here and.... well anyway, all the icons look good.

Did note we don't have a page which has one of all types of admonition (disguised as a Collophon which could act as a reference to how we present things) which could be used to test all these elements in future. Ping seperately if you think thats a good idea.

PR #3849, opened 2023-03-24
Update 04_backup_and_restore.mdx

nataliawojcik27:
@ccestes do we need to say 'sign into the BigAnimal portal' for the cluster restore for PGD clusters? we don't for the regular clusters? i think this can be streamlined


ccestes:
@nataliawojcik27 I merged your changes from this branch and updated the steps so that they're the same as the regular cluster. This is how it looks currently: https://deploy-preview-3764--edb-docs-staging.netlify.app/docs/biganimal/latest/using_cluster/04_backup_and_restore/#perform-an-extreme-high-availability-cluster-restore.

PR #3851, opened 2023-03-27
BDR-3342 - Add rel_notes for HARP 2.2.2

Jagdish-Kewat-EDB:
Update release notes to add record for HARP release 2.2.2.

Have updated the release notes however I am not sure and have few queries regarding this as mentioned below:

1. The release date for PGD 4.3.1 is a TBD however we are releasing HARP 2.2.2 sometime this week. So will there be a new release like PGD 4.3.2 for BDR side of changes ?
2. The fix that is being pushed out for HARP 2.2.2 also needs to be applied for `PGD CLI 1` also. The release for CLI is planned for sometime in May 2023. Will that we a separate release for this too ?
3. Will all these components be released as part of a single PGD release `4.3.1` with `BDR 4.3.1`, `HARP 2.2.2` and `CLI 1.1.1` ? In that case do we have a single release date for all of them ?
4. Can I just add the `HARP 2.2.2` stuff under `PGD 4.3.0` itself ?


Jagdish-Kewat-EDB:
Creating a PR for these changes under PGD 4.3.0

PR #3862, opened 2023-03-29
fixes for local image and layout

djw-m:
## What Changed?

Path names for images to enable local dev viewing 

Image layout to center smaller image


djw-m:
Set to deploy to make it easier to see.

djw-m:
Something strange is afoot. The current deploy doesn't show the first image, apparently skipping the responsive image options for it. This works locally though. More investigation is needed.

drothery-edb (approved):

PR #3878, opened 2023-03-31
Release: 2023-03-31

drothery-edb:
## What Changed?

- PEM 9.1.1 and 8.6.1 release notes
- Ubuntu 22 updates
- pgbBackrest 2.45
- BigAnimal: Softened maintenance window note
- Extensions table updates


install_template/config.yaml
```diff
@@ -91,46 +91,46 @@ products:
     platforms:
       - name: CentOS 7
         arch: x86_64
-        supported versions: [14.1.0.1]
+        supported versions: [15.2.0.1]
```
josh-heyer, line 94: Probably want both versions?

josh-heyer (approved):

PR #3879, opened 2023-04-03
BDR-3342 - Add release note entry for HARP dependencies upgrade

Jagdish-Kewat-EDB:
* Add entry for HARP 2.2.2 release notes for PGD 3.7
* Add the release version HARP 2.2.2 to index page for PGD 4 along with a footnote to upgrade to this version for all installations using HARP 2.2.1

TBH the entry for PGD 4 still looks odd to me. We are mentioning HARP 2.2.2 to be released as part of PGD 4.3.0 which in fact is not the case. Ideally I would have preferred to have any entry for this under PGD 4.3.1 with release date as `30 March 2023` and then the footnote would have made more sense. The detailed rel_notes for PGD 4.3.1 would contain entry only for HARP 2.2.2.
The upcoming release for PGD in May 2023 would then become 4.3.2. I am not sure about the challenges involved with this change in PGD release schedule so my suggestion may not be a viable option here.

### DIFF
--- a/product_docs/docs/pgd/3.7/harp/01_release_notes/harp2.2.2_rel_notes.mdx
+++ b/product_docs/docs/pgd/3.7/harp/01_release_notes/harp2.2.2_rel_notes.mdx
@@ -1,0 +1,10 @@
+---
+title: "Version 2.2.2"
+---
+
+This is a patch release of HARP 2 that includes fixes for issues identified
+in previous versions.
+
+| Type | Description |
+| ---- |------------ |
+| Enhancement     | Upgrade dependencies to fix security vulnerabilities |
--- a/product_docs/docs/pgd/3.7/harp/01_release_notes/harp2.2.2_rel_notes.mdx
+++ b/product_docs/docs/pgd/3.7/harp/01_release_notes/harp2.2.2_rel_notes.mdx
@@ -7,4 +7,4 @@ in previous versions.
 
 | Type | Description |
 | ---- |------------ |
-| Enhancement     | Upgrade dependencies to fix security vulnerabilities |
+| Change | Upgrade 3rd party dependencies to fix Github dependabot alerts |
--- a/product_docs/docs/pgd/4/rel_notes/index.mdx
+++ b/product_docs/docs/pgd/4/rel_notes/index.mdx
@@ -30,7 +30,7 @@ The EDB Postgres Distributed documentation describes the latest version of EDB P
 
 | Release Date | EDB Postgres Distributed     | BDR   | HARP  | CLI | TPAexec                                                                          |
 | ------------ | ---------------------------- | ----- | ----- | --- | -------------------------------------------------------------------------------- |
-| 2023 Feb 14 | [4.3.0](pgd_4.3.0_rel_notes) | 4.3.0 | 2.2.2 | 1.1.0 | [23.9](https://techsupport.enterprisedb.com/customer_portal/sw/tpa/trunk/239/)   |
+| 2023 Feb 14 | [4.3.0](pgd_4.3.0_rel_notes) | 4.3.0 | 2.2.2[^*] | 1.1.0 | [23.9](https://techsupport.enterprisedb.com/customer_portal/sw/tpa/trunk/239/)   |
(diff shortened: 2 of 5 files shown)


irionr:
The changes LGTM, I like the footnote, maybe we could add an asterisk on the `2.2.2`  so that users would really look for that footnote.

drothery-edb (dismissed):

drothery-edb:
Is this ready to be published?

product_docs/docs/pgd/4/rel_notes/index.mdx
```diff
@@ -13,24 +13,24 @@ navigation:
...
+  - /pgd/latest/rel_notes/pgd_4.3.0_rel_notes/
 
 ---
 
 The EDB Postgres Distributed documentation describes the latest version of EDB Postgres Distributed 4, including minor releases and patches. The release notes provide information on what was new in each release. For new functionality introduced in a minor or patch release, the content also indicates the release that introduced the feature.
 
 | Release Date | EDB Postgres Distributed     | BDR   | HARP  | CLI | TPAexec                                                                          |
 | ------------ | ---------------------------- | ----- | ----- | --- | -------------------------------------------------------------------------------- |
-| 2023 Feb 14 | [4.3.0](pgd_4.3.0_rel_notes) | 4.3.0 | 2.2.1 | 1.1.0 | [23.9](https://techsupport.enterprisedb.com/customer_portal/sw/tpa/trunk/239/)   |
+| 2023 Feb 14 | [4.3.0](pgd_4.3.0_rel_notes) | 4.3.0 | 2.2.2[^*] | 1.1.0 | [23.9](https://techsupport.enterprisedb.com/customer_portal/sw/tpa/trunk/239/)   |
```
smaher-edb, line 33: Release date is still not clear for harp like when it was released confusing support and others. We need to have the separate row for harp release either by modifying this table or having separate table. I think Doc can help here. Otherwise for every release this will be problem. I think we have discussed similar thing for [v5 release](https://www.enterprisedb.com/docs/pgd/latest/rel_notes/) as well.

smaher-edb: Just to add, this tabular output is not only problem for harp but for cli, proxy, utilities etc...

irionr: I'm not sure this is the best approach either... 
I mean if we had to release bdr 5.0.1 for the next release (mid-May) we would have PGD5.0.2 BDR5.0.1 HARP5.0.1 etc... 
I don't know what would be the best solution though :( 
Was this discussed with Petr or Kelly?

smaher-edb: > I mean if we had to release bdr 5.0.1 for the next release (mid-May) we would have PGD5.0.2 BDR5.0.1 HARP5.0.1

Yes, thats what we have mentioned [in this comment](https://github.com/EnterpriseDB/docs/pull/3796#discussion_r1138694435)

product_docs/docs/pgd/3.7/harp/01_release_notes/harp2.2.2_rel_notes.mdx
```diff
@@ -0,0 +1,10 @@
+---
+title: "Version 2.2.2"
+---
+
+This is a patch release of HARP 2 that includes fixes for issues identified
+in previous versions.
+
+| Type | Description |
+| ---- |------------ |
+| Enhancement     | Upgrade dependencies to fix security vulnerabilities |
```
smaher-edb, line 10: I think this should be `Change` and may be we need change the description little. It was for dependabot alerts and not the direct vulnerabilities in code.

Jagdish-Kewat-EDB: Sure

product_docs/docs/pgd/4/rel_notes/index.mdx
Jagdish-Kewat-EDB: Hi @drothery-edb @PJMODOS @smaher-edb @irionr , can we please conclude on this so that we can bring it to closure ?

PJMODOS: I kind of agree it's not ideal, however having combinations listed together likethis is at request of customers.

PJMODOS (approved):

PR #3883, opened 2023-04-03
File structure to automatically generate install files for PEM 8

dwicinas:
Extensive restructure within the "Installing the PEM Server" and "Installing the PEM agent" directories.


drothery-edb:
@dwicinas this looks good to me. Did you check for broken links?

dwicinas:
@drothery-edb I fixed all broken links and I changed a few instances of "latest" to "8" in links. But there were a lot of those instances, even in just the installing-related directories. So I didn't make an effort to make the latest-to-8 change comprehensively. I can do so if you think that's appropriate.

drothery-edb (approved):

PR #3899, opened 2023-04-06
PEM: SQL Profiler package name corrections

drothery-edb:


drothery-edb:
Changes moved to https://github.com/EnterpriseDB/docs/pull/3914

PR #3902, opened 2023-04-06
Update index.mdx

ftouserkani-edb:


drothery-edb (approved):

drothery-edb:
@ftouserkani-edb Thanks for making this change. It didn't make it into the latest build, but it will get into the next one.

PR #3931, opened 2023-04-14
Table crunching - NOT FOR MERGIN

djw-m:
## What Changed?

Just some table contents


ebgitelman:
@nidhibhammar Can EPAS service account be changed to Service? If not, how can I prevent "EPAS service account" from breaking? @drothery-edb

nidhibhammar:
> @nidhibhammar Can EPAS service account be changed to Service? If not, how can I prevent "EPAS service account" from breaking? @drothery-edb

I think we can say "EDB account" but would like to see what recommendation Dee Dee has.

PR #3941, opened 2023-04-17
Add information about how efm interprets ping

EFM-Bobby:
Adding a note that the important part about pinging the VIP is the exit code of the ping command. This comes from a customer replacing /bin/ping with something else that always exited with zero even though the human-readable output. In that case EFM always thought the address being pinged was reachable, breaking failover when using a VIP.

### DIFF
--- a/product_docs/docs/efm/4/04_configuring_efm/05_using_vip_addresses.mdx
+++ b/product_docs/docs/efm/4/04_configuring_efm/05_using_vip_addresses.mdx
@@ -82,7 +82,7 @@ PING 172.24.38.239 (172.24.38.239) 56(84) bytes of data.
  time 3000ms
 ```
 
-You see 100% packet loss.
+You see 100% packet loss. Important: Failover Manager uses the exit code of the command to determine whether or not the address was reachable. In this case the exit code is not zero. If using a different command, it must return a non-zero exit code if the address is not reachable.
 
 2.  Run the `efm_address add4` command on the Primary node to assign the VIP, and then confirm with ip address:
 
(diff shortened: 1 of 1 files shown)


EFM-Bobby:
Adding jira link so I can find PR/Jira from the other. https://enterprisedb.atlassian.net/browse/EFM-1595

EFM-Bobby:
@drothery-edb Hi again. If there's a different person (or group) I should bother for these things, let me know please. No hurry on this one.

product_docs/docs/efm/4/04_configuring_efm/05_using_vip_addresses.mdx
```diff
@@ -72,75 +72,79 @@ ping.server.command=/bin/ping -q -c3 -w5
...
+   # /bin/ping -q -c3 -w5 172.24.38.239
+   PING 172.24.38.239 (172.24.38.239) 56(84) bytes of data.
+   --- 172.24.38.239 ping statistics ---
+   3 packets transmitted, 3 received, 0% packet loss, time 1999ms
+   rtt min/avg/max/mdev = 0.023/0.025/0.029/0.006 ms
+   ```
+
+   No packet loss occurs. 
+   !!!Important
+   Failover Manager uses the exit code of the command to determine whether or not the address was reachable. In this case the exit code is zero. If using a different command, it must return a zero exit code if the address is reachable.
```
drothery-edb, line 112: @EFM-Bobby , if you are ok with these edits I'll make them to the other note too.
```suggestion
   Failover Manager uses the exit code of the ping command to determine whether or not the address was reachable. In this case the exit code is zero. If using a command other than ping, it must return a zero exit code if the address is reachable.
```

EFM-Bobby: Yes, that looks good to me. Thank you.

drothery-edb (approved):

EFM-Bobby:
@drothery-edb Thank you again!

PR #3951, opened 2023-04-19
BigAnimal: users can now set Ultra Disk throughput

drothery-edb:


amrita-suresh (changes requested):
maybe it's worth explicitly mentioning that while BA calculates throughput based on IOPS, the user can optionally configure throughput themselves as well?

PR #3957, opened 2023-04-21
Ensure PEM 8 install files redirect from PEM 8

josh-heyer:
## What Changed?

Using latest in redirects for PEM 8 doesn't work out in this situation, as they'll either do nothing or intercept PEM 9 paths


josh-heyer:
Just FYI @dwicinas - nothing terribly interesting here other than that I used an expression instead of version-specific templates - the intent being that the version doesn't matter so much as just "less than [latest] vs [latest]"

PR #3971, opened 2023-04-24
EFM: Postgres prereq

drothery-edb:
## What Changed?

Updated EPAS prereq to apply to both PG and EPAS


drothery-edb:
Took another approach - see https://github.com/EnterpriseDB/docs/pull/3977

PR #3973, opened 2023-04-24
Incorrect link to "Eager replication" description

amacvar:
On this page page
 https://www.enterprisedb.com/docs/pgd/latest/
"CAMO, or Eager replication.", Eager text link points to 
  https://www.enterprisedb.com/docs/pgd/latest/consistency/eager/
it should be
 https://www.enterprisedb.com/docs/pgd/5/bdr/eager/
But that link doesn't exist


amacvar:
Waiting for review

amacvar:
Sorry clicked on wrong button

djw-m:
The current link goes to Eager Conflict Resolution (also known as eager replication), which is the correct page for PGD 5. Eager Replication content is integrated into that page, but the overall naming was changed. 

That said, this PR changes the link to one that does not exist (there is no bdr directory in PGD 5 documentation), and thus that would break the docs. So I am closing this PR without merging, but thanks for the contribution.

PR #3977, opened 2023-04-25
Failover Manager: install template changes

drothery-edb:
## What Changed?

Clarified that both PG and EPAS are supported and that you don't need to install the db for witness nodes

See related discussion https://edb.slack.com/archives/C016SDV9HLL/p1682341493767469


josh-heyer (approved):
I like that this makes the templates simpler than your previous PR 🙂

PR #3984, opened 2023-04-26
EPAS - Quick start reorg as per DF-322

nidhibhammar:


install_template/templates/products/edb-postgres-advanced-server/v15_index.njk
```diff
@@ -0,0 +1,8 @@
+{% extends "./index.njk" %}
```
nidhibhammar, line 1: @josh-heyer, we added this file, but it is not getting deployed.

product_docs/docs/epas/15/installing/linux_x86_64/epas_centos_7.mdx
```diff
@@ -46,39 +46,72 @@ Installing the server package creates an operating system user named enterprised
...
 INSERT INTO dept VALUES (10,'ACCOUNTING','NEW YORK');
 INSERT into dept VALUES (20,'RESEARCH','DALLAS');
 
-# You can use simple SQL commands to query the database and retrieve
-# information about the data you have added to the table
 SELECT * FROM dept;
-__OUTPUT__
+```
+
+This example includes the output of the commands so you can cross-check the output when you run the commands. Don't copy commands from this example as it includes the output and would result in errors.
```
nidhibhammar, line 94: @josh-heyer, I added this second example with outputs as I am getting the output at execution of each command and not at last after execution of all commands. Do you agree with this approach? or you have any other recommendation?

install_template/templates/products/edb-postgres-advanced-server/v15_index.njk
josh-heyer: The indentation was different between the redirects entries in install.njk and those in v15_install.njk - this resulted in unparsable yaml and the entire file was skipped. I've fixed the indentation, and also updated the deploy script to produce more visible and (slightly) more informative error messages when this happens.

product_docs/docs/epas/15/installing/linux_x86_64/epas_centos_7.mdx
josh-heyer: I don't like it; it's verbose, and it separates the commands that will be executed from their output, which is the whole point. 

I would say, if the output is too extensive to be comfortable... Break it up. This is a demo section - there is no particular need to have it easy to copy-paste en masse if that interferes with readability. Each command or logically-grouped set of commands can then have their own `__OUTPUT__` section without repetition.

josh-heyer (changes requested):
Recommend breaking up the examples such that output can be rendered next to the commands that generate it.

drothery-edb:
@josh-heyer we incorporated your suggestions in the Experiment section. Let us know what you think. Also, what is the deal with lock file version? Is the test block it is in something that should be removed? The version number got updated with my latest changes.

josh-heyer:
The lockfile version was my mistake - forgot to run the docker rebuild before committing. Cleaned up.

josh-heyer (approved):
The example changes look good to me!

PR #3990, opened 2023-04-26
Set products for search within extensions

josh-heyer:
## What Changed?

To fully enable search for these, we need two things:

1. Set a product name in directory defaults for each extension (this associates each extension's content with the product name and enables the facet in advanced search)
2. Add these names to products.js (this enables selecting the name from the search dropdown)


djw-m (commented):
But going forward we'll have a mix of spaced and underscored product names and no canonical tag as the names match the "product names".

Would I be more consistent to have canonical, unspaced names across the board?

edb_advanced_storage_pack
edb_ldap_sync
edb_postgres_tuner
pg_failover_slots

djw-m (approved):
LGTM

PR #4007, opened 2023-04-29
Install templates: updated Postgres prereq 

drothery-edb:
## What Changed?

Added PostgreSQL to connectors, FDWs, and P-products prereqs


josh-heyer (approved):
Looks good. Since you were updating the epas version number elsewhere, I did it for the PostGIS example as well - also took this opportunity to move that include into the product directory, as it is (and should be) only used by that product.

PR #4009, opened 2023-04-30
Update INSTALL.mdx

amacvar:
Based on the discussion here @josh-heyer suggested adding the word offline. I would also suggest adding the word air-gapped

## What Changed?
Added synonyms to the title, so hopefully they show up in the document search


sonotley:
Thanks for your PRs. The TPA docs are just copied to this website from their source repo so we can't merge them here but we'll include the changes in the upstream as we need to make some additions to this section anyway.

PR #4012, opened 2023-05-02
Update architectures.mdx

aswright491:
The link to RAFT is correct, but I think the spacing is wrong because when I click to it from the docs it doesn't work.


drothery-edb:
@aswright491 thanks for catching this. The root cause was related to the wrapping in the source file. Thanks to @djw-m, it is now fixed.

drothery-edb (approved):

Issue #3813, opened 2023-03-17
Add Linux install topics for PostgreSQL using the template framework


drothery-edb:
### Summary

In the prerequisite section provide the two repo options: EDB repos or community repos. See https://www.enterprisedb.com/docs/pem/latest/installing/linux_x86_64/pem_rhel_8/ for the links to the community repos.

### Additional notes

Feedback from the field: https://edb.slack.com/archives/CU5QEU5L7/p1678925210378679

PR #4015, opened 2023-05-02
More developed effort to templatize Linux install instructions for PostgreSQL

dwicinas:
## What Changed?

Fixes https://github.com/EnterpriseDB/docs/issues/3813


dwicinas:
@drothery-edb This is a first crack and restructuring introductory information for PostgreSQL. Maybe you can take a look at what information I preserved and where I put it. I created an "Installation overview" topic, which may not be ideal but it seemed like a pretty easy way to capture some of the introductory information. Happy to take another approach, if you have ideas. Also, note that the Windows topic is still using graphics. I can remove those when we're ready.

dwicinas:
@drothery-edb This PR is ready for your review and a technical review too.

dwicinas:
@drothery-edb, I think this is ready for your review, again. It now incorporates the "check your repo" step, like the other products. I've also updated the Windows and Mac topics, including graphics, to reflect version 15.

dwicinas:
@drothery-edb The changed files do include that package lock file. I'm not sure what to do about that.

drothery-edb:
Before we merge this we should add the topics for RHEL 9 for versions 12-15.

drothery-edb (approved):

PR #4018, opened 2023-05-03
Document support for HASH partitioning in AutoPartition

pavanvd:
This was added in PGD 5 release, but we missed updating the docs then. Also add note that SQL functions such as
`bdr.autopartition_create_partition` and `bdr.autopartition_drop_partition` must not be used by users directly.

BDR-3287


PJMODOS (approved):
Looks okay, I am not sure I like the "you" language but that's not new.

PR #4023, opened 2023-05-03
First draft of documentation changes for efm 4.7

EFM-Bobby:
## What Changed?

Version changes through user's guide.

Added 4.7 release notes page.

### DIFF
--- a/install_template/templates/products/failover-manager/base.njk
+++ b/install_template/templates/products/failover-manager/base.njk
@@ -21,7 +21,7 @@ redirects:
 {{ super() }}
 {% endblock product_prerequisites %}
 {% block postinstall %}
-Where `<4x>` is the version of Failover Manager that you are installing. For example, if you are installing version 4.6, the package name would be `edb-efm46`.
+Where `<4x>` is the version of Failover Manager that you are installing. For example, if you are installing version 4.7, the package name would be `edb-efm47`.
 
 The installation process creates a user named efm that has privileges to invoke scripts that control the Failover Manager service for clusters owned by enterprisedb or postgres.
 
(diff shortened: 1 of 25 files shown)


kelpoole (approved):
I've looked through the changes and the release notes. I'm assuming there's no inline content where the enhancement change should be discussed.

kelpoole (approved):
I've looked through the changes and the release notes. I'm assuming there's no inline content where the enhancement change should be discussed.

EFM-Bobby:
> I've looked through the changes and the release notes. I'm assuming there's no inline content where the enhancement change should be discussed.

Good thinking, and correct. The guide only mentions that we create a backup when a standby is reconfigured and that the user can remove it, which is still the case.

EFM-Bobby:
Hi @drothery-edb, can you give this a look when you get a chance? I don't know when the packages will actually be released, so we shouldn't merge these changes yet, but would like to do everything we can ahead of time.

EFM-Bobby:
Update for whomever -- I changed the 4.7 release date in the doc to tomorrow, May 25, just to pick *something*. Am happy to change it as needed.

drothery-edb:
@EFM-Bobby , the individual install topics are generated from a template. Any changes needed to those topics should be done in the templates and the files regenerated. You can see the change I made to the templates [here](https://github.com/EnterpriseDB/docs/pull/4023/commits/63067a1eb9dacf6557119f5d73620140f8d55fb3). You are welcome to familiarize yourself with our template framework or we can make those changes for you in the future.

drothery-edb (dismissed):

drothery-edb (dismissed):

PR #4024, opened 2023-05-04
Docs/new/arrow reveal markers

djw-m:
## What Changed?

Arrows as per design poc

Simple rule, add a frontMatter navTitle with unicode `▶` at end of title.


josh-heyer (approved):
This is good as a proof of concept. If it gains approval, we'll implement this in code and roll it out across all products.

PR #4049, opened 2023-05-10
First pass clean up

djw-m:
## What Changed?

Removing the legacy from synchronous replication.


PJMODOS:
With 5.1 we could potentially remove all the legacy (both PSR and logical) sync stuff from the table and just mention that we don't recommend using the `synchronous_standby_names`.

djw-m:
@PJMODOS The request to remove legacy as part of the naming was, as I understood it, because we still actively have users using it and the idea of recommending legacy things to customers was not a good thing. (See https://enterprisedb.atlassian.net/browse/DF-334?focusedCommentId=432385)

This seems like an apposite point to abandon this branch and start work on revising the whole replication section with the sync commit in scopes and deciding on what replication options we want to present. (and revert the discussion back to the Jira ticket).

PJMODOS:
> @PJMODOS The request to remove legacy as part of the naming was, as I understood it, because we still actively have users using it and the idea of recommending legacy things to customers was not a good thing. (See https://enterprisedb.atlassian.net/browse/DF-334?focusedCommentId=432385)
> 
> This seems like an apposite point to abandon this branch and start work on revising the whole replication section with the sync commit in scopes and deciding on what replication options we want to present. (and revert the discussion back to the Jira ticket).

It is legacy though, the non-legacy way of configuring it is the new SYNCHRONOUS_COMMIT in commit scopes.

I see two options:
- either we keep legacy logical sync and psr sync in the table (and put them close together, at the end of the table, just as a comparison for other options and recommend against using them
- or we simply remove both legacy logical sync and psr sync from the table

Either way we need to add the new SYNCHRONOUS_COMMIT in commit scopes to the table.

djw-m:
Moving this discussion elsewhere. This PR is no longer required except for reference.

PR #4051, opened 2023-05-11
Changes quickstart for 23.17 --pgd-proxy-routing

djw-m:
## What Changed?

Quickstart entries for --active-locations removed, added --pgd-proxy-routing, adjusted some other docs similarly.

### DIFF
--- a/product_docs/docs/pgd/5/quickstart/quick_start_docker.mdx
+++ b/product_docs/docs/pgd/5/quickstart/quick_start_docker.mdx
@@ -174,7 +174,7 @@ tpaexec configure democluster \
   --edb-postgres-advanced 15 \
   --redwood \
   --location-names dc1 \
-  --active-locations dc1 \
+  --pgd-proxy-routing global \
   --no-git \
   --hostnames-unsorted
 ```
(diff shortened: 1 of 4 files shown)


product_docs/docs/pgd/5/quickstart/quick_start_docker.mdx
```diff
@@ -196,7 +196,7 @@ Linux as the default image for all nodes.
 
 Specify that the data nodes will be running [EDB Postgres Advanced Server v15](/epas/latest/) (`--edb-postgres-advanced 15`) with Oracle compatibility (`--redwood`). 
 
-You set the notional location of the nodes to `dc1` using `--location-names`. You then activate the PGD proxies in that location using `--active-locations dc1` set to the same location.
+You set the notional location of the nodes to `dc1` using `--location-names`. You then set `--pgd-proxy-routing` to `global` so that proxy routing can route traffic to all nodes in the cluster.
```
sonotley, line 199: Would it be more accurate to say something a little more specific? Global means writes are routed to a single write leader for the whole cluster, and local means they are routed to a local write leader.

sonotley (approved):

product_docs/docs/pgd/5/quickstart/quick_start_docker.mdx
djw-m: This is in the context of the quickstart, where we won't be discussing write leaders per se till deeper into the explore topics. It does make me consider modifying the explore content as a secondary update though.

PR #4066, opened 2023-05-15
Standardize SLES topics so they do not use the "address other prerequisites" approach

dwicinas:
Changes to templates plus changes to all generated SLES topics


dwicinas:
@drothery-edb This should be ready for your review. All SLES topics are no longer using the "Address other prerequisites" approach in the bullet list of prereqs.

drothery-edb (approved):

PR #4069, opened 2023-05-15
BigAnimal: giving poolers more visibility

drothery-edb:
## What Changed?

Moved poolers and FDWs topics to their own topics with links to them from the Postgres extensions and tools topic

### DIFF
--- a/product_docs/docs/biganimal/release/overview/extensions_tools.mdx
+++ b/product_docs/docs/biganimal/release/overview/extensions_tools.mdx
@@ -5,14 +5,12 @@ navTitle: Supported extensions and tools
 
 BigAnimal supports a number of Postgres extensions and tools, which you can install on or alongside your cluster.
 
-## Postgres extensions
-
 - See [Postgres extensions available by deployment](/pg_extensions/) for the complete list of extensions BigAnimal supports.
 
 - See [Postgres extensions](/biganimal/release/using_cluster/03_modifying_your_cluster/extensions) for more information on installing and working with extensions.
 
-### EDB Postgres extensions
-EDB develops and maintains several extensions. These include:
+### EDB Postgres extensions and tools
+EDB develops and maintains several extensions and tools. These include:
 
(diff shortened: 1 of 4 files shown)


aswright491:
@drothery-edb thank you! I was actually thinking we should exclude pg-pool. I think it makes a lot of sense to include in knowledgebase if we are asked how to set it up on a VM, but most users will likely use pgbouncer and pgpool isn't actually IN the BA architecture.

drothery-edb:
> @drothery-edb thank you! I was actually thinking we should exclude pg-pool. I think it makes a lot of sense to include in knowledgebase if we are asked how to set it up on a VM, but most users will likely use pgbouncer and pgpool isn't actually IN the BA architecture.

@aswright491 Thanks for your feedback. We can certainly remove the information about using what we are calling "self-hosted" poolers as opposed to "BA-hosted" ones. 

I'll give you a little history on the pooler topics though because we have been getting opposing requests about the content. There was an initiative to move content that was deemed useful to have in the docs from the KB. I think Amrita led that initiative. The articles about self-hosted Pg-Pool and PgBouncer were deemed worthy to bring into the docs. Both were drafted although only the PgBouncer one finished the review process and made it into the docs. We had it in the docs for about a year, but decided to take it out and just link to the KB article when we made the changes for the BA-hosted PgBouncer feature. With Jason's help we have refined our strategy about what goes into the docs vs the KB and added more links to articles as well as added a new FAQs index topic with links to certain articles. 

I assume even with the background, you still feel we should pull the info on PgPool (and probably the self-hosted option we mention for PgBouncer too). I would be inclined to keep the links we have in the [FAQs topic](https://deploy-preview-4069--edb-docs-staging.netlify.app/docs/biganimal/latest/knowledge_base/#configuring-self-hosted-connection-poolers) in place though.

Just curious what you think about the [FDWs topic](https://deploy-preview-4069--edb-docs-staging.netlify.app/docs/biganimal/latest/overview/foreign_data_wrappers/). Is this a similar case?

PR #4099, opened 2023-05-17
BigAnimal: additional PgBouncer pricing info

drothery-edb:
## What Changed?

Addresses feedback from the CSAs


drothery-edb:
Jason approved via Slack

PR #4102, opened 2023-05-17
Update installation_upgrade.mdx

kelpoole:
Clarify that the user and password fields map to Repo and Token for Cloudsmith


djw-m:
The edited change is now also upstreamed as enterprisedb/pg4k-pgd#308.

drothery-edb (approved):

PR #4112, opened 2023-05-19
Simple KB excluder

djw-m:
## What Changed?

Adds

hideKBLink: 

as a front matter option.

When set to true, will omit the KB footnote from the headers.


josh-heyer (approved):
Needed a small tweak to make using it not annoying, otherwise looks great!

PR #4115, opened 2023-05-22
Update pgd_5.0.0_rel_notes.mdx

martinmarques:


djw-m:
@martinmarques we can't see any change in this PR. Are we missing something subtle?

martinmarques:
The user is expected to do this manually  after upgrading to the latest

(missing "do")

irionr (approved):
LGTM

PR #4134, opened 2023-05-24
Changed config dot yaml entries for OCL back to 14 and 15

dwicinas:
config.yaml
redirects for all generated OCL topics

### DIFF
--- a/install_template/config.yaml
+++ b/install_template/config.yaml
@@ -91,46 +91,46 @@ products:
     platforms:
       - name: CentOS 7
         arch: x86_64
-        supported versions: [14.1.0.1, 15.2.0.1]
+        supported versions: [14, 15]
       - name: RHEL 8
         arch: ppc64le
-        supported versions: [14.1.0.1, 15.2.0.1]
+        supported versions: [14, 15]
       - name: AlmaLinux 8 or Rocky Linux 8
         arch: x86_64
-        supported versions: [14.1.0.1, 15.2.0.1]
+        supported versions: [14, 15]
       - name: RHEL 7 or OL 7
         arch: x86_64
-        supported versions: [14.1.0.1, 15.2.0.1]
+        supported versions: [14, 15]
       - name: RHEL 8 or OL 8
         arch: x86_64
-        supported versions: [14.1.0.1, 15.2.0.1]
+        supported versions: [14, 15]
       - name: Debian 10
         arch: x86_64
-        supported versions: [14.1.0.1, 15.2.0.1]
+        supported versions: [14, 15]
       - name: Debian 11
         arch: x86_64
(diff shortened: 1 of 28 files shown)


dwicinas:
@drothery-edb this should be good to go

drothery-edb:
@dwicinas @josh-heyer I'm not 100% following the need to change all the redirects to update the templates after I changed the doc versions for OCL to major versions.  If we need a redirect (which I'm not sure we do since I did a redirect in the global _redirects file when I did the version consolidation) wouldn't we add one? not update the existing one? Dave, did you make all those changes manually?

josh-heyer:
Short answer is "test it" @drothery-edb - build this branch and try one of the old paths, see if it not only redirects to the major version but also to the new page.

In general, if you're adding a blanket redirect for a version change should try to avoid using the old version in specific redirects anywhere - they SHOULD work just fine, but... It depends on the order in which the redirects get processed. If the blanket redirect gets processed first, the specific redirect with the full version will never match.

josh-heyer (approved):
Looks good. Observe that this works: https://deploy-preview-4134--edb-docs-staging.netlify.app/docs/ocl_connector/14/04_open_client_library/01_installing_and_configuring_the_ocl_connector/install_on_linux_using_edb_repo/ibm_power_ppc64le/ocl_sles15_ppcle

...while this does not: https://deploy-preview-4134--edb-docs-staging.netlify.app/docs/ocl_connector/14.1.0.1/04_open_client_library/01_installing_and_configuring_the_ocl_connector/install_on_linux_using_edb_repo/ibm_power_ppc64le/ocl_sles15_ppcle

(No splat in redirect)

Nor does this:
https://deploy-preview-4134--edb-docs-staging.netlify.app/docs/ocl_connector/15.2.0.2/04_open_client_library/01_installing_and_configuring_the_ocl_connector/install_on_linux_using_edb_repo/ibm_power_ppc64le/ocl_sles15_ppcle

(Redirect missing entirely for 15)

So THIS PR is fine, and we need some more adjustments to the _static redirect

drothery-edb:
@josh-heyer, what is wrong with these redirects in the _redirect file? (redirects are not my friend)

/docs/ocl_connector/14.1.0.1/*  /docs/ocl_connector/14/:splat  301
/docs/ocl_connector/15.2.0.1/*  /docs/ocl_connector/15/:splat  301

Just to confirm, the redirects in this PR where automagically created, correct?

josh-heyer:
I typo'd the 15 path, so no issues there. For 14, there were overlapping rules - the first rule hit takes precedence, and it didn't have a splat. Fixed in develop

drothery-edb:
the yaml changes were made in another branch - https://github.com/EnterpriseDB/docs/pull/4139

Issue #4120, opened 2023-05-23
Scroll areas for left and right nav


djw-m:
Our scroll bars do not scroll freely or follow the users body content selection.

We should make them follow the users scrolling within the body on the right lighting up sections into the right based on what is currently in view or pointer is over.

The left hand nav should try and center on the current topic so that user has context.

We should look at building these enhancements to always give users full navigational context.

drothery-edb:
Fixed by #4138

PR #4138, opened 2023-05-25
Nav bar usability enhancements

josh-heyer:
## What Changed?

Build #4024 into the system as follows:

- record number of children when building nav tree
- apply CSS class to nav links where childCount > 0
- CSS class renders arrow using Bootstrap's weird-ass border technique

Result: arrows next to expandable items in list.
TODO: figure out if accessibility can be improved here (ARIA role?)

Partially address https://github.com/EnterpriseDB/docs/issues/4120 as follows:
- fix height of right nav bar, enable scrollbar when content exceeds
- indicate currently-selected section via CSS class
- on navigation, ensure currently selected items in both left and right nav
  are scrolled into view if not currently visible
- using gatsby hook, record current scroll state for left/right navs,
  restore this state when navigating back

### DIFF
--- a/gatsby-browser.js
+++ b/gatsby-browser.js
@@ -10,6 +10,13 @@ exports.onRouteUpdate = ({ location }) => scrollToAnchor(location);
  * @param {Number} [mainNavHeight] - the height of any persistent nav -> document.querySelector(`nav`)
  */
 function scrollToAnchor(location, mainNavHeight = 0) {
+  // left nav: scroll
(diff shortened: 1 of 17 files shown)


djw-m (dismissed):
Lovely, was tempted to add a commit removing the manual > from the PGD docs.

Seems to nail most locations.

Also pulled in a version of my Reference content branch to test against, and right hand navbar is now super responsive.

PR #4141, opened 2023-05-25
PGD: apply font treatment and fix case

drothery-edb:
## What Changed?

I lost the font treatment when I removed the link - fixing that and a few other little things


djw-m (approved):
LGTM

PR #4148, opened 2023-05-26
Update tpa_23.14_rel_notes.mdx

kelpoole:
Add missing change to release notes as requested by ACI


sonotley (approved):
Happy with change, although style guide from Dee Dee says no Jira numbers, so we'll see if they make the cut.

drothery-edb (approved):

drothery-edb:
It is ultimately PM's call whether to use our internal Jira numbers in tickets. I'm assuming there was a reason in this particular case.

PR #4159, opened 2023-05-29
Update bdr.mdx

martinmarques:
There was a back tick where a single quote should be inside the block code. This didn't show up too bad in the web docs, but did mangle the PDF a lot


djw-m (approved):
LGTM

PR #3995, opened 2023-04-27
Draft - Doc/new/raftdocs

djw-m:
## What Changed?

First pass at structuring the Raft content for visibility.

Looking for input on un-addressed issues.

Note the example cluster, for brand awareness, is now pgdgroup

The complete listing of config.yml has been replaced by annotated segments.

Alternate graphics are included but these are still placeholders for Graphic Design work.

## Questions to be addressed

With subgroups being baked into the product, the current "what's wrong with the way we did it" section seems redundant. Thoughts on possibly inverting that and driving the narrative from how we do subgroups now?

The migration section is somewhat light. Notes on additions there would be appreciated.

PR #4216, opened 2023-06-08
PGD5 Raft content (repick from older branch)

djw-m:
## What Changed?

All new(ish) from the older branch #3995


mwanner (approved):
LGTM now

PR #4232, opened 2023-06-13
Add workaround for EFM monitoring

sonotley:
Add workaround for case where the cluster properties file is not in the home directory of the efm user.


nidhibhammar (approved):

matthew123987:
Awesome!

drothery-edb (approved):

PR #4247, opened 2023-06-14
Update 03_overview_correction_strategies.mdx

djw-m:
Fixed code block language to remove "shql"


ccestes (approved):
Ah yes, good old shql

PR #4252, opened 2023-06-15
Connectors: Updated supported platform topics

francoughlin:
## What Changed?

Updated supported platform topics per https://enterprisedb.atlassian.net/browse/DF-338 for:

- EDB*Plus
- Hadoop FDW
- Mongo FDW
- JDBC connector
- OCL connector
- ODBC connector
- PgPoole
- PgBouncer
- PostGIS


drothery-edb:
@francoughlin can you check if any of these products are supported on Windows or MacOS and if they are remove Linux from the first sentence? For example, EDB*Plus is supported on Windows as well as Linux platforms.

francoughlin:
> @francoughlin can you check if any of these products are supported on Windows or MacOS and if they are remove Linux from the first sentence? For example, EDB*Plus is supported on Windows as well as Linux platforms.

Will do!

drothery-edb (approved):

PR #4297, opened 2023-06-23
BigAnimal: fix to pricing

drothery-edb:
## What Changed?

Per Slack message, Natalia says we don't need the 2nd PostgreSQL line in the pricing table


drothery-edb:
Approved by Natalia in Slack

PR #4321, opened 2023-06-30
Biganimal - describe edb_admin role as "user" instead of "role"

nidhibhammar:
## What Changed?

updated all reference of edb_admin to edb_admin user


drothery-edb (approved):

ba-edb:
What's the background behind this decision?

I'd suggest we bounce it off some of the PG community folks to get their reaction.

PR #4342, opened 2023-07-03
Update installation instructions for ASP

tureba:
- ASP packages are only available in version 12 onwards. Version 11 was never a possibility for ASP.
- EL 9 is now released, so added to the distro list
- Ubuntu is the only Debian derivative that we support, so mention it instead of letting customers think any other derivative is included
- The 1 and -1 were dropped from the package names a while back, and the packages have been released without them, with appropriate upgrade paths
- Newer versions than 1.0.0 have been out as well, so the installation instructions are not helping anyone by mentioning that past version explicitly
- Also fixing the last example that intends to install for EPAS, but shows the instruction for PostgreSQL


drothery-edb:
@tureba thanks for making these edits.

PR #4347, opened 2023-07-05
Added RHEL 9 to installing topics for PostgreSQL

dwicinas:
Changed config.yaml and files in install_template/products to support RHEL 9. Then generated new topics based on template changes


dwicinas:
@drothery-edb I think this is ready for review and possible merge. I'll start a separate PR to document the cookbook.

drothery-edb (approved):

PR #4355, opened 2023-07-06
EPAS: updating terminology

drothery-edb:


drothery-edb:
@francoughlin Nidhi and I think we should standardize on "Postgres" mode as opposed to "standard" mode or "PostgreSQL" mode. What do you think?

francoughlin:
> @francoughlin Nidhi and I think we should standardize on "Postgres" mode as opposed to "standard" mode or "PostgreSQL" mode. What do you think?

Agreed! I will add a task for that in the Phase 2 implementation, if that makes sense to you guys.

francoughlin (approved):

PR #4366, opened 2023-07-06
EPAS 15: Reorg cleanup

francoughlin:
## What Changed?

Minor edits to various topics that were not completed prior to reorg


drothery-edb (approved):
LGTM but you got a stray change in from PGD. It is really hard to avoid that currently and I don't think it will do any harm, but look out for that change when you commit in the future.

francoughlin:
> LGTM but you got a stray change in from PGD. It is really hard to avoid that currently and I don't think it will do any harm, but look out for that change when you commit in the future.

Dang it! I will have to watch those commits more closely and not assume. Thanks for pointing that out!

PR #4368, opened 2023-07-07
Added note on TDE presence check

djw-m:

### DIFF
--- a/product_docs/docs/tde/15/enabling_tde.mdx
+++ b/product_docs/docs/tde/15/enabling_tde.mdx
@@ -110,7 +110,7 @@ This example uses EDB Postgres Advanced Server 15 running on a Linux platform. I
 
 The presence of TDE on a server can be determined by querying the `data_encryption_version` column of the `pg_control_init` table. 
 
-A value of 0 means TDE is not enabled. Any non-zero value reflects the version of TDE in use. Currently, when TDE is enabled, this value will be 1.
+A value of 0 means TDE isn't enabled. Any non-zero value reflects the version of TDE in use. Currently, when TDE is enabled, this value is 1.
 
 ```sql
 # select data_encryption_version from pg_control_init;
--- a/product_docs/docs/tde/15/enabling_tde.mdx
+++ b/product_docs/docs/tde/15/enabling_tde.mdx
@@ -65,9 +65,6 @@ If you don't specify this parameter, the environment variable `PGDATAKEYUNWRAPCM
 
 Use the special value `-` if you don't want to apply any key unwrapping command.
 
-<!--- Should this be worded differently? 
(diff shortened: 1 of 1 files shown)


product_docs/docs/tde/15/enabling_tde.mdx
```diff
@@ -65,7 +65,8 @@ If you don't specify this parameter, the environment variable `PGDATAKEYUNWRAPCM
 
 Use the special value `-` if you don't want to apply any key unwrapping command.
 
-<!--Should this be worded differently? Is this the environment variable fallback? I changed the word "option" to parameter. -->
+<!--- Should this be worded differently?
```
nidhibhammar, line 68: I think we will be removing this and the following line

product_docs/docs/tde/15/enabling_tde.mdx
```diff
@@ -108,3 +109,17 @@ This example uses EDB Postgres Advanced Server 15 running on a Linux platform. I
    data_encryption_key_unwrap_command = 'openssl enc -d -aes-128-cbc -pass pass:ok -in %p'
    ```
 
+## Checking for TDE in use using SQL
```
nidhibhammar, line 112: Can we reword this like - Checking for TDE presence using SQL. or something else? Intend is not to use "use" along with using.

nidhibhammar (changes requested):
Rest looks good to me

product_docs/docs/tde/15/enabling_tde.mdx
djw-m: Removed... No idea how the line break appeared, but we should resolve questions about content without leaving long lived and untracked comments in the docs.

product_docs/docs/tde/15/enabling_tde.mdx
djw-m: Done

product_docs/docs/tde/15/enabling_tde.mdx
```diff
@@ -108,3 +106,17 @@ This example uses EDB Postgres Advanced Server 15 running on a Linux platform. I
    data_encryption_key_unwrap_command = 'openssl enc -d -aes-128-cbc -pass pass:ok -in %p'
    ```
 
+## Checking for TDE presence using SQL
+
+The presence of TDE on a server can be determined by querying the `data_encryption_version` column of the `pg_control_init` table. 
+
+A value of 0 means TDE is not enabled. Any non-zero value reflects the version of TDE in use. Currently, when TDE is enabled, this value will be 1.
```
drothery-edb, line 113:
```suggestion
A value of 0 means TDE isn't enabled. Any non-zero value reflects the version of TDE in use. Currently, when TDE is enabled, this value is 1.
```

drothery-edb (approved):

PR #4373, opened 2023-07-07
BigAnimal: update to note about csp support for bha

drothery-edb:
## What Changed?

removing "only" from not about BAH CSP support


drothery-edb:
per request from Natalia on Slack

francoughlin (approved):

PR #4393, opened 2023-07-11
Sample edits for HPE

ebgitelman:
Example edits in two files


jkitchens32:
I like the edits and their flow.  Added @scott-edb for the final say on this one to make sure it is all good with what he is thinking.

drothery-edb (approved):

PR #4411, opened 2023-07-12
update install version of efm

EFM-Bobby:
Changed to "47" to match the rest of the example.


EFM-Bobby:
Found a quick something I missed before. :)

drothery-edb (approved):

EFM-Bobby:
Thanks! That was fast.

PR #4427, opened 2023-07-14
Migration Portal: fix version number

drothery-edb:
## What Changed?

https://enterprisedb.atlassian.net/browse/MIG-4701


usmanmuzaffar-edb (approved):
Change looks good to me.

PR #4454, opened 2023-07-18
Home page: moved Migration block back to the top

drothery-edb:
## What Changed?

After the recent addition of new partner docs, the Migration block got pushed to the bottom of the second column. Did some rearranging to get it back to the top right position.


djw-m (approved):
LGTM

PR #4463, opened 2023-07-18
Edits to failover slots content

ebgitelman:


ebgitelman:
@drothery-edb I revised a long complicated sentence by making a bullet list. I think I unraveled it properly but please double check.

drothery-edb (approved):

PR #4470, opened 2023-07-19
Extensions: added bdr5

drothery-edb:
## What Changed?

Added BDR 5 and fixed table headings


drothery-edb:
Fixed by https://github.com/EnterpriseDB/docs/pull/4471

PR #4480, opened 2023-07-21
Update configuring_pgBouncer.mdx

matthew123987:
adding missing bracket, remove copy from comment


sonotley (approved):
Looks ok to me, not tested.

'COPY(' isn;t actually commented out, it's just that explicitly including the PSQL prompt confused the code formatter.

PR #4487, opened 2023-07-24
PGD: Updating TPA links

### LINKED ISSUE #4425: Queries regarding PGD 4 release notes doc
A couple of questions regarding the index page.

1. We have a column for TPA however the actual TPA change details are not mentioned in the respective page of PGD release notes. Instead, it's updated on the customer portal page which is outside docs I believe. So I am wondering if the TPA column should be added here or not. Maybe this can be mentioned on the compatibility matrix page. If we choose to add here then can the TPA release note entries also be added to the PGD release notes page like BDR and HARP ?

2. We have a component called bdr-utilities which is part of the PGD ecosystem, however, don't have a column for it on the index page. I can see a mention of this in release notes for 4.3.1 so wondering if we can add a column for this too.

ccestes:
## What Changed?

Fixes #4425


ccestes:
@josh-heyer I updated the v23 links, which are covered in our TPA docs, however v22 is not. Should we leave those links as is, or have them lead elsewhere?

ccestes:
@josh-heyer In that case, can we close this PR?

drothery-edb:
> @josh-heyer In that case, can we close this PR?

I don't think so. I think Dj just removed the v22 links, but didn't add the other links you have added in this PR. Let's request a review from Dj.

djw-m:
I did all of it in 4434, just didn't get to merge it. My bad. Now merged so you can see the changes in the conflict editor. I'd close this PR.

PR #4502, opened 2023-07-25
Revised upgrade language to be more consistent with other product

dwicinas:
Revisions to one topic describing Linux upgrades. These revisions parallel similar language used for other product upgrades.


dwicinas:
Yes, although I believe we talked about having Nidhi take a look at the PEM PR.

nidhibhammar (approved):
LGTM

PR #4510, opened 2023-07-26
Explicit 2PC usage w/ CAMO/Eager/GC is a limitation

mwanner:
## What Changed?

Remove the explicit 2PC usage w/ CAMO/Eager/GC from "known issues" to "design limitations"

BDR-3330


djw-m (approved):
LGTM

PR #4458, opened 2023-07-18
Parallel Apply : Updating content for 5.2

djw-m:
## What Changed?

First pass at updating - requires technical review.

PR #4545, opened 2023-08-01
Doc/pgd/upd/pafor5.2reload

djw-m:
## What Changed?

Extracted content by hand from source PR #4458 and effectively rebased. All changes should be in and this will be up for re-review.


djw-m:
Merged into #4560

PR #4551, opened 2023-08-01
Update index.mdx

mto79:
It is an upgrade from 14 to 15 and not 13 to 15


drothery-edb:
Thanks for making this correction, @mto79.

PR #4552, opened 2023-08-01
Edits to PR4505

ebgitelman:
and changes from EPAS to spelling out, eliminating "first use" call-out


ebgitelman:
The words "EPAS service account" exist on many reference pages under Configuration Parameters. Are those generated? I didn't change those cases.

Also, I wonder if on the index for the whole doc set, we might *there* call out the EPAS acronym in the first use of EDB Postgres Advanced Server and say (sometimes referred to as EPAS).

francoughlin:
> The words "EPAS service account" exist on many reference pages under Configuration Parameters. Are those generated? I didn't change those cases.
> 
> Also, I wonder if on the index for the whole doc set, we might _there_ call out the EPAS acronym in the first use of EDB Postgres Advanced Server and say (sometimes referred to as EPAS).

Totally agree with the idea about the call out on the doc set index page; good call! Not sure about the "EPAS service account" question, but maybe @nidhibhammar would know? Thanks for making the updates!

francoughlin (approved):

drothery-edb:
> > Also, I wonder if on the index for the whole doc set, we might _there_ call out the EPAS acronym in the first use of EDB Postgres Advanced Server and say (sometimes referred to as EPAS).
> 
> Totally agree with the idea about the call out on the doc set index page; good call! 

Do you mean introducing EPAS in the main landing page topic title like we do for PGD?
image

francoughlin:
> > > Also, I wonder if on the index for the whole doc set, we might _there_ call out the EPAS acronym in the first use of EDB Postgres Advanced Server and say (sometimes referred to as EPAS).
> > 
> > 
> > Totally agree with the idea about the call out on the doc set index page; good call!
> 
> Do you mean introducing EPAS in the main landing page topic title like we do for PGD? image
> 
> I think that makes sense. Let's implement that.
> 
> In terms of the EPAS service name, those topics aren't generated but I think it is a case where the acronym is called for.

@drothery-edb, as we discussed, I can take care of that update.

drothery-edb (approved):

PR #4556, opened 2023-08-02
BigAnimal: remove CLI limitation

drothery-edb:


drothery-edb:
Guang approved on Slack. https://edb-starlight.slack.com/archives/C02J6HXMH0F/p1690773852689889

PR #4589, opened 2023-08-08
Update index.mdx

djw-m:
Switch dates around to an actual human format.


drothery-edb (commented):
I'm the guilty party here. Somehow Kelly and I came up with this date format when we came up with the release note content model and it is used throughout the doc set. If we want to change it, I can task someone to do everywhere. Or we can live with it. I'm good with either approach, but would rather not fix it in just one version of one product.

djw-m:
> I'm the guilty party here. Somehow Kelly and I came up with this date format when we came up with the release note content model and it is used throughout the doc set. If we want to change it, I can task someone to do everywhere. Or we can live with it. I'm good with either approach, but would rather not fix it in just one version of one product.

Well... If it's only in release notes, it should be easy enough to find all references - here's a find to fond 

 find . -type f -exec grep -H "20[0-9][0-9] [A-Z][a-z][a-z] [0-9]* " {} \;

Shall ponder if I can globally automate that.

PR #4591, opened 2023-08-08
Switched to ticks and unified ordering as most current upgrade first

djw-m:
## What Changed?

Orders are all most current upgrades first (as in the life of any edition the most likely upgrade will be within that version - other older version upgrades will be likely consulted once and never again).

X changed to ✓ to indicate a positive upgrade path.

Typos fixed in old 3.7 docs.

### DIFF
--- a/product_docs/docs/pgd/3.7/bdr/upgrades/supported_paths.mdx
+++ b/product_docs/docs/pgd/3.7/bdr/upgrades/supported_paths.mdx
@@ -1,13 +1,13 @@
 ---
 title: Supported BDR upgrade paths
 ---
-### Upgrading from version 3.6 to version 3.7
-
-We recommend you upgrade your cluster to the latest version of 3.6 before upgrading your cluster to 3.7.  You will then be able to upgrade to a version of 3.7 released on the same date or later. 
-
 
 ### Upgrading within version 3.7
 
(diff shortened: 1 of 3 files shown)


product_docs/docs/pgd/3.7/bdr/upgrades/supported_paths.mdx
```diff
@@ -1,13 +1,13 @@
 ---
 title: Supported BDR upgrade paths
 ---
-### Upgrading from version 3.6 to version 3.7
 
-We recommend you upgrade your cluster to the latest version of 3.6 before upgrading your cluster to 3.7.  You will then be able to upgrade to a version of 3.7 released on the same date or later. 
+### Upgrading within version 3.7
 
+BDR 3.7 does not follow semantic versioning. Instead, it uses 2 position majors similar to PostgreSQL releases prior to version 10 (e.g. 9.6).  BDR 3.7 also considered releases prior to 3.7.9 as beta releases and they could include breaking changes. BDR 3.7.9 was declared a Long Term Support version under the previous versioning nomenclature. All releases 3.7.9 and later maintain backward compatibility allowing upgrades to later versions.
```
drothery-edb, line 7:
```suggestion
BDR 3.7 does not follow semantic versioning. Instead, it uses 2 position majors similar to PostgreSQL releases prior to version 10 (for example, 9.6).  BDR 3.7 also considered releases prior to 3.7.9 as beta releases and they could include breaking changes. BDR 3.7.9 was declared a Long Term Support version under the previous versioning nomenclature. All releases 3.7.9 and later maintain backward compatibility allowing upgrades to later versions.
```

drothery-edb (approved):

drothery-edb (approved):

PR #4597, opened 2023-08-09
dirty links change to make link to index work

### LINKED ISSUE #4596: Broken links in routing section
Detected in PDF linker

image

Links in https://www.enterprisedb.com/docs/pgd/latest/routing/installing_proxy/ don't work in PDF or web.

djw-m:
## What Changed?

Links inside routing pointed at index, but a reference to [index#configuration] is misread and comes out as index/#configuration which fails to resolve. Only two links in the codebase like this, so changed here to [../routing#configuration] which does resolve to the index page.

May need to raise issue as bug or as wider discussion of why don't we have a more classical folder/files hierarchy. 

Should close #4596.


Jagdish-Kewat-EDB (approved):
This links thing is always confusing. LGTM

PR #4624, opened 2023-08-14
BigAnimal: removing free trial reference from Terraform topic

drothery-edb:


djw-m (approved):
Assuming the change is only about tone (as the link is still to free-trial)...

LGTM

PR #4640, opened 2023-08-17
Explicitly call out psql -U for OpenShift

maxim-edb:
Update docs so it is clear users need to call the `kubectl cnp psql` command explicitly with a user on OpenShift due to the fact that OpenShift randomly assigns UID to users in containers.

## What Changed?

Added a Note to clarify the above under the block of text about the `kubectl cnp psql` command.


josh-heyer:
Thanks @maxim-edb !

PR #4648, opened 2023-08-17
Edits to install generated topics

ebgitelman:
to use as basis for updating templates


drothery-edb:
edits implemented in #4770

PR #4659, opened 2023-08-21
OCL Connector: 15.2.0.4 release

drothery-edb:


nidhibhammar (dismissed):

drothery-edb:
@moazzumali @ericmack I added a table to show EPAS compatibility but I was guessing a bit and not sure it works. If it doesn't work I think we can just mention the EPAS 15.4 requirement in the release notes. Let me know.

drothery-edb:
Eric approved via Slack.

PR #4669, opened 2023-08-22
test 2x self-hosted runner

josh-heyer:
## What Changed?

Tweaked draft deploy to build PDFs and run against EDB self-hosted 2x large runner


josh-heyer:
Test succeeded. See #4672

PR #4670, opened 2023-08-22
Fix notes for bdr.run_on_nodes and bdr.run_on_group

PJMODOS:
The text was copy pasted from `bdr.run_on_all_nodes` and the wording and warnings need to be somewhat different for these two functions


djw-m (commented):
Just a small suggestion, but generally fine.

josh-heyer (approved):

PR #4682, opened 2023-08-24
Speed up Netlify deploys

josh-heyer:
Key to accomplishing this is reducing the number of files changed from build-to-build to as close as possible to those corresponding to the set of pages actually *changed*. To this end, 

- Eliminates most client-side redirects by actually rewriting current version links to /latest/ instead of depending on a redirect to handle this (a server-side redirect is still created so that versioned URLs will work in production, but there shouldn't be any links *within* docs that use a versioned path to a latest-version page). The goal here is to eliminate the client-side redirects table, which is carried along with every page's data and will tend to change every time a file is added / removed / renamed - thus adding a single file could result in *every* page being re-deployed.
- Stop using the last commit SHA in static links for reporting errors or viewing page history. It was *nice to have this*, but it also meant that every page would change on every commit... So every page had to be redeployed on every commit. To retain the benefits, I've moved to rerendering these links client-side to reintroduce the hash. If JavaScript is disabled, the static version, containing the branch name, will be used - which is good enough in most cases. 
- Strip the webpackCompilationHash value from both generated HTML files and app-data.json. This is another thing that would cause tiny, localized changes to trigger redeployment of every single page. The intent here is to allow pages to recover gracefully when a new deployment lands after they've been loaded... That's cool and all, but I kinda feel like the number of folks likely to be affected by that is somewhere between few and none. It'd be nice if there was a way to make this just... Not change so readily... But right now, it's the difference between small content changes having deploy times measured in tens of minutes vs tens of seconds. Easily the biggest hack in this set of changes though; if we do see ugly errors, will have to revisit.

View the deploy history for this PR to see results.

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -3,7 +3,6 @@ const realFs = require("fs");
 const path = require("path");
 const gracefulFs = require("graceful-fs");
 gracefulFs.gracefulify(realFs);
-
 const { createFilePath } = require(`gatsby-source-filesystem`);
 const { exec, execSync } = require("child_process");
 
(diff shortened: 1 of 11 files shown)


djw-m (approved):
LGTM - The code looks a lot less gnarly too

PR #4697, opened 2023-08-25
Homepage: What's new - epas reorg

drothery-edb:


src/pages/index.js
```diff
@@ -99,18 +99,18 @@ const Page = () => (
                     className="homepage-headling-link"
                     to="/pgd/latest/reference/"
```
josh-heyer, line 100: This link and the one below need to point to EPAS 😉

josh-heyer (changes requested):

PR #4722, opened 2023-08-29
BigAnimal: pooler language updates

drothery-edb:
## What Changed?

moving away from "self-hosted" terminology for poolers deployed and managed separate from BigAnimal


francoughlin (approved):
Looks good!

PR #4726, opened 2023-08-30
Move online help redirects to PEM 9

josh-heyer:
## What Changed?

Online help section was eliminated from the PEM 9 published docs - redirects should go there.


nidhibhammar:
@josh-heyer, I have moved the redirects to actual topic files. If you disagree, you can revert the last commit.

nidhibhammar (approved):

PR #4752, opened 2023-09-04
Essential fix and relocate of information

djw-m:
## What Changed?

Moved items to to limitations, added entry for Group Commit (temporary as future new Durability limitations will include this).

### DIFF
--- a/product_docs/docs/pgd/5/limitations.mdx
+++ b/product_docs/docs/pgd/5/limitations.mdx
@@ -108,7 +108,7 @@ Be sure to disable transaction streaming when planning to use
 CAMO. You can configure this option globally or in the PGD node group. See
 [Transaction streaming configuration](../transaction-streaming#configuration).
 
--   Not all DDL can run when CAMO is used. The only supported DDL operations are:
+-   Not all DDL can run when CAMO is used. If unsupported DDL is used a warning will be logged and the transactions commit scope will be set to local only. The only supported DDL operations are:
     - non-concurrent CREATE INDEX
     - non-concurrent DROP INDEX
     - non-concurrent REINDEX of an individual table or index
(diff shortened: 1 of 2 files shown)


product_docs/docs/pgd/5/limitations.mdx
```diff
@@ -108,6 +108,27 @@ Be sure to disable transaction streaming when planning to use
 CAMO. You can configure this option globally or in the PGD node group. See
 [Transaction streaming configuration](../transaction-streaming#configuration).
 
+-   Not all DDL can run when CAMO is used. The only supported DDL operations are:
```
mwanner, line 111: Oh, I think we could be more specific here: CAMO and commit scopes gets automatically disabled for DDL that it does not support. You'll get for example:

```
WARNING:  0A000: commit scopes automatically disabled for ALTER TABLE
DETAIL:  This transaction's commit scope will be local only.
```

And yes, this is stupid. But no, we did not invent it. It was added upon specific (important) customer's request. :-) 

Same applies for Eager.

mwanner (changes requested):

product_docs/docs/pgd/5/limitations.mdx
djw-m, line 111:
```suggestion
-   Not all DDL can run when CAMO is used. If unsupported DDL is used a warning will be logged and the transactions commit scope will be set to local only. The only supported DDL operations are:
```

djw-m:
This branch is blocked (post mortem to come) - New PR incoming.

product_docs/docs/pgd/5/limitations.mdx
djw-m: Done with a small Eager section reffing the GC entry to avoid repeating the same info three times.

PR #4762, opened 2023-09-05
Constructor fix for older releases

dwicinas:


drothery-edb:
changes moved to https://github.com/EnterpriseDB/docs/pull/4768

PR #4785, opened 2023-09-07
Update 06_configuring_for_eager_failover.mdx

tureba:
The safest way for end users to edit service files is using `systemctl edit`. Alternatively, we can instruct them on how to edit the overrides files in `/etc/systemd/system/<unitname>.d/override.conf`. But creating a new file and `.include` shouldn't have been an option. Moreover, the `.include` feature has been removed from recent systemd versions, so it must be changed before it starts erroring out in the customer's hands.


drothery-edb (approved):

EFM-Bobby (approved):
Looks good to me. The example uses efm 4.5, which is fine -- doesn't have to be the latest. But I'll update it when I make the changes for the next release so it doesn't look like this section has been forgotten.

Issue #4783, opened 2023-09-07
Feedback on efm 4 - "Using Failover Manager"


tureba:
### Summary

The instructions suggest creating service files inside `/usr/lib/systemd/system`. But that is not where units created by the system administrator should be placed. Instead, they should be placed in `/etc/systemd/system`.

The instructions should be clear when saying the reader should copy the reference file from its source location `/usr/lib/systemd/system` into `/etc/systemd/system` with a new name, then edit it as necessary.

### Where did you see the problem?

https://www.enterprisedb.com/docs/efm/latest/05_using_efm/#rhelcentos-7x-or-rhelrocky-linuxalmalinux-8x

tureba:
I wasn't planning on proposing the changes, but I ended up writing https://github.com/EnterpriseDB/docs/pull/4788 for this.

drothery-edb:
Fixes https://github.com/EnterpriseDB/docs/issues/4783

drothery-edb:
Thank you for addressing this, @tureba! Your changes will be in today's build.

PR #4788, opened 2023-09-07
Update 05_using_efm.mdx

tureba:
This changes the instructions on how to customize systemd service files. The instructions were directing the reader to edit files in /usr/lib/systemd/system, which is not the adequate location. Whenever a system administrator edits service files, they should always do so in /etc/systemd/system, either with systemctl edit or editing the files directly, but never editing anything in /lib or /usr.

It also shows what the reader should expect to see in each file completely (i.e, exactly three lines), which helps them review their steps before proceeding further with the documentation. Previously, the instructions mentioned lines separately in an unnumbered set of instructions, which is more difficult to follow.

### DIFF
--- a/product_docs/docs/efm/4/05_using_efm.mdx
+++ b/product_docs/docs/efm/4/05_using_efm.mdx
@@ -265,27 +265,37 @@ After creating the `acctg.properties` and `sales.properties` files, create a ser
 
 ### RHEL/CentOS 7.x or RHEL/Rocky Linux/AlmaLinux 8.x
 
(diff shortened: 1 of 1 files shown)


EFM-Bobby (dismissed):
This looks good to me. I guess we need to update https://github.com/EnterpriseDB/efm/blob/master/src/main/resources/efm_unit_file as well.

EFM-Bobby:
Filed https://enterprisedb.atlassian.net/browse/EFM-1642 to get the instructions in the unit file we package changed. I can do it as part of the 4.8 release.

drothery-edb (approved):

drothery-edb:
Fixes https://github.com/EnterpriseDB/docs/issues/4783

Issue #3686, opened 2023-02-19
Regarding /pg_extensions/advanced_storage_pack/using.mdx


abip-sjarbini:
## Summary
Advanced example guidance doesn't work as expected

Version:
 PostgreSQL 15.2 (EnterpriseDB Advanced Server 15.2.0) on x86_64-pc-linux-gnu, compiled by gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-44)
, 64-bit

## Where did you see the problem?
USING autocluster is required to create nyse_trade
After inserting into nyse_trade_symbol, nyse_trade table stays empty

## Expected behavior
select ctid, * from nyse_trade gives result

## Screenshots
![image](https://user-images.githubusercontent.com/102211786/219907967-03749cbf-fd3c-4f72-bcc5-423176ced5d6.png)

## Browser / Platform
N/A

## Additional notes
N/A

drothery-edb:
Thanks for the feedback. We are looking into this.

ccestes:
The `select ctid, * from nyse_trade` command returns no output because the static data has been shortened for brevity in the example. So instead of inserting the full list of stock exchange symbols, numbers, and trades, the user is prompted to insert a shortened list, which doesn't include the data in the final output. 

However, the fact that the data is shortened is unclear to the user, so we plan to explain that better in the documentation and make all the data accessible in GitHub or elsewhere.

ccestes:
@josh-heyer Here is the file for the full example. The data sets are under `INSERT INTO` commands. 
https://drive.google.com/file/d/19xhng1Q5z0JhA1iMZ-OzwhKpuXd4qoXg/view?usp=sharing

josh-heyer:
Arthur is working on this now: https://edb.slack.com/archives/CU5QEU5L7/p1693314872322199

tureba:
I've submitted PR https://github.com/EnterpriseDB/docs/pull/4805 with a possible way of improving the examples. Reviews are much appreciated.

PR #4805, opened 2023-09-08
pg_extensions/ASP/using: improve examples

tureba:
The example for refdata was improved to show the different locking pattern that replaces the row locks for a table lock, leading to a stronger lock when the reference table is changed, but a lower overhead when the referenced rows are accessed. The demonstration uses pgrowlocks, which is available for heap, and will be available in a future version of our custom TAMs.

The example for autocluster was maintained.

The advanced example that uses both TAMs was improved to show the effect that autocluster has on the query plan, leading to fewer pages touched. Previously it showed the ctid of the rows, but that is already displayed in the autocluster example, so showing the different query plans fills a gap in the overall explanation of autocluster.

The examples were reworked to not need external data, through the usage of artificial data generated on the fly.

This is meant to address issue https://github.com/EnterpriseDB/docs/issues/3686

### DIFF
--- a/advocacy_docs/pg_extensions/advanced_storage_pack/using.mdx
+++ b/advocacy_docs/pg_extensions/advanced_storage_pack/using.mdx
@@ -3,20 +3,119 @@ title: Using EDB Advanced Storage Pack
 navTitle: Using
 ---
 
-The following are scenarios where the EDB Advances Storage Pack TAMs are useful.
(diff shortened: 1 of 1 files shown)


ccestes (approved):
Looks much improved--thanks @tureba!

markdilger (approved):
This looks better.  Thanks, Arthur!

PR #4810, opened 2023-09-11
Update index.mdx

kelpoole:
Requested via JIRA-4081

Changed title to match content which is about PGD upgrades not Application Schema Upgrades


LenzGr:
Thanks for the suggestion! This was already updated via #4808.

PR #4814, opened 2023-09-11
BigAnimal: commit scopes recommendation

drothery-edb:


drothery-edb:
Natalia approved on Slack

PR #4819, opened 2023-09-12
BigAnimal: fixing PEM remote monitoring link

drothery-edb:
## What Changed?

The Remote monitoring link in the third bullet on this [page](https://www.enterprisedb.com/docs/biganimal/latest/using_cluster/05_monitoring_and_logging/) 404s even though there is a redirect for it. This PR fixes the link, but it might be worth investigating why the redirect didn't work.


francoughlin (approved):
Looks good - redirect works

PR #4843, opened 2023-09-25
New style for release notes adding full title and date

dwicinas:


djw-m (approved):
LGTM - would have liked one or two older versions also dated but we can move forward from this. Did you check that the change on CNPG won't get overwritten on import or are the release notes still manually created?

nidhibhammar (approved):
LGTM. We can proceed with this and do the same task for two older versions for all the products in separate PR.

PR #4860, opened 2023-09-27
update dependencies: September 2023

josh-heyer:
## What Changed?

Notable updates:

- Husky
- Prettier
- Algolia 
- Bootstrap

The last two required substantial revisions to support the major version updates. As part of these, I've fixed a few related issues:
- incorrect path parsing for 404 page search (it was prefixing most searches with "docs")
- race condition in global search bar initialization (click too fast, no search)
- honor system dark mode preference when no local preference is set

Also replaced pretty-quick with lint-staged, as new prettier is unsupported by the former. May not have the configuration quite right for this.


djw-m:
The search box's category -> All Products etc. -> has flipped from white on blue to black on blue.

djw-m:
image

Little odd - I assume that it's the mismatch between the algolia db and the preview but wanted to flag that.

djw-m (dismissed):
Apart from the color change in the search box,LGTM. Couldn't snag anything else.

josh-heyer:
Took me way too long to figure out the color thing. Turns out Boostrap 5 picks a foreground color based on what will provide the best contrast for the background... And our blue has crappy contrast with white:

image

Options: 

1. Leave text black (and adjust styles on other buttons to match)
2. Use a slightly darker shade of blue (e.g. `#3e6e8e`)

djw-m:
I would likely go for a darker shade of blue and the black/blue contrast is
more painful.

josh-heyer:
Altered rules to achieve consistency for now; added #4868 to track future work on better contrast.

PR #4861, opened 2023-09-28
Add edb_sqlpatch information on minor version up page for EPAS 15 manual

yuki-tei:
Hi team,
Greetings.

We've found that your minor upgrade page lacks the information about edb_sqlpatch description, which is needed for example v15.4.

## What Changed?
Add edb_sqlpatch command sample and link to its manual page at bottom.

### DIFF
--- a/product_docs/docs/epas/15/upgrading/05_performing_a_minor_version_update_of_an_rpm_installation.mdx
+++ b/product_docs/docs/epas/15/upgrading/05_performing_a_minor_version_update_of_an_rpm_installation.mdx
@@ -27,3 +27,16 @@ yum update edb*
     The `yum update` command performs an update only between minor releases. To update between major releases, use `pg_upgrade`.
 
 For more information about using `yum` commands and options, enter `yum --help` at the command line.
+
+!!! If edb_sqlpatch result gives below message, it is necessary to execute edb_sqlpatch for system catalog.
+    Here is an output example:
+    ```shell
+    * database edb
+    0 patches were previously applied to this database.
+    58 patches need to be applied to this database.
+    ```
+    In this case , you need to execute this post rpm update:
+    ```shell
+    edb_sqlpatch -af
+    ```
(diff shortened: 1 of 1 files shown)


djw-m (commented):
This is very unclear. The instructions don't instruct you to run edb_sqlpatch, but you are supposed to take action based on it's result?

Does any of the preceding steps run edb_sqlpatch automatically? If not, shouldn't there be a patching step?

Happy to discuss.

yuki-tei:
Hi,
Thanks for your quick reply. Apologies for my description might be confusing.

The most important point for this page is that edb_sqlpatch doesn't exist till 15.4.
So the newbee for EDB might look into this minor update instruction page and doesn't aware of the need to apply edb_sqlpatch.

So the idea was first to at least give a link for edb_sqlpatch on this page.
However, even do so, especially the new customers might not know in which case they need edb_sqlpatch post minor release, one idea as you see is to decide by edb_sqlpatch's command result itself. If you have any other ideas, please tell us.

Thanks for your kind advice.
Regards,
Yuki Tei

djw-m:
So, these instructions should read...


!!! 
    
If on version 15.4 or later, you should run `edb_sqlpatch`

If the command responds that it has a number of patches needing to be applied like so:

```console
* database edb
0 patches were previously applied to this database.
58 patches need to be applied to this database.
```

Then it will be necessary to execute edb_sqlpatch to patch the system catalog. Run:

```shell
edb_sqlpatch -af
```

!!!

yuki-tei:
Hi,

Thanks for your quick review and proofreading. 
We believe that makes sense and would be a great help to many customers.
Appreciates to your continuous help.

Kind Regards,
Yuki Tei

djw-m:
@nidhibhammar This change will need back-porting to refer to 11.21.32, 12.16.20, 13.12.17 and 14.9.0 if we merge it.

nidhibhammar (approved):
LGTM. I agree with back-porting.

PR #4867, opened 2023-09-29
Update epas16_beta_rel_notes.mdx

theadamwright:
Before: 
Transparent data encryption (TDE) now supports the AES key length 256 option to TDE. The default is still AES-128, but the AES-256 option is available for use cases where 256-bit is a regulatory requirement and for feature parity with SQL Server's TDE technology that uses AES-256.

After: 
Transparent data encryption (TDE) now supports the AES key length 256 option to TDE. The default is still AES-128. AES-256 is typically required for regulatory compliance or recommended when looking for maximum security.


djw-m (approved):
LGTM (reviewed on phone)

djw-m (approved):
LGTM (reviewed on phone)0

PR #4876, opened 2023-10-04
Redirect unversioned product URLs to latest, search fixes

### LINKED ISSUE #4331: Search associates content from previous section with section header
## Summary

When a search matches a section header, it is possible for the excerpt to contain content from a previous section.

## Steps to reproduce

1. Search the PEM docs for "Scheduled job". https://www.enterprisedb.com/docs/search/?query=scheduled+job&product=pem&version=9
2. Observe result for [/pem/latest/managing_pem_agent/#managing-pem-scheduled-jobs](https://www.enterprisedb.com/docs/pem/latest/managing_pem_agent/#managing-pem-scheduled-jobs) includes an excerpt from the previous section ("Use the Email group field to specify the email group to send the email notification to") 
   
   image

## Expected behavior

Would expect to see the first (or at least, a relevant) sentence from the matching section.

## Notes

Not sure whether this arises due to how Algolia composes excerpts or how we index sections, but the specific appearance - section header, colon, sentence - looks suspiciously like how we pass the section header to ensure it is indexed when breaking up a long section into multiple "chunks". Would be good to double-check and verify we're not prepending this to the wrong text in this context.

### LINKED ISSUE #4872: Unversioned URLs 404
Example:

https://www.enterprisedb.com/docs/pgd/

results in a 404 when it _probably_ should redirect to

https://www.enterprisedb.com/docs/pgd/latest/

bonus issue:

image

None of the related URLs pointed at the top of the PGD docs. Some of them are some pretty deep cuts into the upgrade docs which seems weird.

josh-heyer:
## What Changed?

- Automatically add redirects from versioned product root to latest version. E.g., /pgd/ -> /pgd/latest/
- Choose object IDs for Algolia such that the default tie-breaker ordering (descending string order by ID) will mostly match navigation order.
- Include full hierarchy of section titles in search titles. E.g., # Installing ## Overview ### Linux -> Installing >> Overview >> Linux
- Pre-render global search bar (faster loading)
- Send search result click events to Algolia for better analytics, optimization
- Enable advanced syntax for searches (quoted terms, term exclusion). E.g., `"sql patch" -advisories`

Fixes #4872
Fixes #4331

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -582,11 +582,18 @@ exports.onPostBuild = async ({ graphql, reporter, pathPrefix }) => {
   //
   // get rid of compilation hash - speeds up netlify deploys
   //
+  const hashTimer = reporter.activityTimer("Removing compilation hashes");
+  hashTimer.start();
   const { globby } = await import("globby");
   const generatedHTML = await globby([
     path.join(__dirname, "/public/**/*.html"),
   ]);
-  for (let filename of generatedHTML) {
+  for (
+    let i = 0, filename;
+    i < generatedHTML.length, (filename = generatedHTML[i]);
+    ++i
+  ) {
+    hashTimer.setStatus(`${i + 1}/${generatedHTML.length}`);
     let file = await readFile(filename);
     file = file.replace(
       /window\.___webpackCompilationHash="[^"]+"/,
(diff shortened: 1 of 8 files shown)


djw-m (approved):
LGTM (as far as I can see without a deploy which would be pointless with the Algolia stuff)

PR #4895, opened 2023-10-11
Move scrolling to proper callback, avoid animation 

josh-heyer:
## What Changed?

Appears there were at least three different places where scrolling would potentially be forced on navigation:

1. a script injected by the autolink headers plugin (only on page load)
2. gatsby's router logic (page load and internal navigation)
3. navigation callback in gatsby-browser.js (page load and internal navigation)

...not to mention the default browser behavior, which would (should?) also work on page load.

Generally, this didn't seem to cause issues, but occasionally an odd behavior would emerge: navigating to a section within a page would jump the scroll position much further down. Oddly, this seems to disappear if the \#3 logic is eliminated in favor of a selective override of \#2.

### DIFF
--- a/gatsby-browser.js
+++ b/gatsby-browser.js
@@ -3,23 +3,6 @@ require("prismjs/plugins/command-line/prism-command-line.css");
 
 exports.onRouteUpdate = ({ location }) => scrollToAnchor(location);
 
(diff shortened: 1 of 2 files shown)


djw-m:
Seems to work here, much snappier browsing the reference too.

PR #4930, opened 2023-10-18
Bdr 4160 docs document the 2 phases of two phase commit on terminology page

djw-m:
## What Changed?

Removed pre-commit from language


mwanner (dismissed):
SGTM, thanks.

PR #4937, opened 2023-10-20
POC first commit for taskbook

djw-m:

### DIFF
--- a/product_docs/docs/pgd/5/taskbook/administration/rolling_database_upgrades.mdx
+++ b/product_docs/docs/pgd/5/taskbook/administration/rolling_database_upgrades.mdx
@@ -1,0 +1,329 @@
+---
+title: Rolling Database Upgrades
+---
+
+## Task
+
+Upgrading a PGD data node's Postgres installation between major versions.
+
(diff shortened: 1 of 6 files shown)


product_docs/docs/pgd/5/taskbook/administration/rolling_database_upgrades.mdx
```diff
@@ -0,0 +1,342 @@
...
+Stopping new cluster
+```
+
+### 10 - Modify the Postgres service
+
+The Postgres service on the system is configured to start the old version 
+of Postgres (PG14). We need to modify the `postgres.service` file to start the new 
+version (PG15). This can be done using sed to replace the old version 
+number "14" with "15" throughout the file. Once changed, we can tell the 
+systemd daemon to reload the configuration. On the target node, run:
```
mwanner, line 285: What? Shouldn't this be handled properly by the packages? At least on Debian (sorry, my bias), you can run multiple instances and different major versions in parallel. I would strongly recommend against manually fiddling with `.service` files there.

PR #4939, opened 2023-10-23
Removed defunct shared entry in reference

djw-m:
## What Changed?

Empty shared instance page (and reference section block) are gone.


djw-m:
This will, I assume, land in the next release of TPA in the code, but any reason not to merge it now @sonotley ?

PR #4957, opened 2023-10-30
EFM does not drop the VIP when the primary agent is stopped

EFM-Bobby:
This is occasionally a source of confusion, and we'd like to make it clear that EFM doesn't drop the vip when it exits. This would break a cluster that uses a VIP.

Am not sure if it makes sense to have two "Note" boxes next to each other at the top of the page, so if there's a better way to handle it am happy with any changes.

### DIFF
--- a/product_docs/docs/efm/4/04_configuring_efm/05_using_vip_addresses.mdx
+++ b/product_docs/docs/efm/4/04_configuring_efm/05_using_vip_addresses.mdx
@@ -15,6 +15,9 @@ Failover Manager uses the `efm_address` script to assign or release a virtual IP
 !!! Note
     Virtual IP addresses aren't supported by many cloud providers. In those environments, use another mechanism, such as an elastic IP address on AWS, that can be changed when needed by a fencing or post-promotion script.
 
+!!! Note
+    Failover Manager will not drop the Virtual IP address (if used) from the primary node when the agent is stopped. As a convenience for testing, the primary agent will acquire the VIP at startup if the node does not already have it, but otherwise starting and stopping Failover Manager has no effect on the node having the virtual IP address.
+
 By default, the script resides in:
 
  `/usr/edb/efm-4.<x>/bin/efm_address`
@@ -82,7 +85,7 @@ When instructed to ping the VIP from a node, use the command defined by the `pin
    time 3000ms
    ```
 
-   You see 100% packet loss. 
+   You see 100% packet loss.
    !!!important
    Failover Manager uses the exit code of the ping command to determine whether the address was reachable. In this case, the exit code isn't zero. If you're using a command other than ping, it must return a non-zero exit code if the address isn't reachable.
 
@@ -107,7 +110,7 @@ When instructed to ping the VIP from a node, use the command defined by the `pin
    rtt min/avg/max/mdev = 0.023/0.025/0.029/0.006 ms
    ```
 
-   No packet loss occurs. 
+   No packet loss occurs.
    !!!Important
    Failover Manager uses the exit code of the ping command to determine whether the address was reachable. In this case, the exit code is zero. If you're using a command other than ping, it must return a zero exit code if the address is reachable.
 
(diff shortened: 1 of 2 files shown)


EFM-Bobby:
@martinmarques Just pinging you for visibility.

product_docs/docs/efm/4/upgrading.mdx
```diff
@@ -33,6 +33,8 @@ Failover Manager provides a utility to assist you when upgrading a cluster manag
    ```shell
    /usr/efm-4.4/bin/efm stop-cluster efm
    ```
+!!! Note
+    The primary agent will not drop the virtual IP address (if used) when it is stopped.
```
josh-heyer, line 37: What's the rationale for noting this here? Is there a scenario where dropping the VIP during an upgrade is desirable / necessary?

product_docs/docs/efm/4/04_configuring_efm/05_using_vip_addresses.mdx
```diff
@@ -12,9 +12,12 @@ legacyRedirectsGenerated:
 
 Failover Manager uses the `efm_address` script to assign or release a virtual IP address.
 
-!!! Note
+!!! Note "Cloud provider support and alternatives"
     Virtual IP addresses aren't supported by many cloud providers. In those environments, use another mechanism, such as an elastic IP address on AWS, that can be changed when needed by a fencing or post-promotion script.
 
+!!! Note "Behavior when the agent is stopped"
+    Failover Manager will not drop the Virtual IP address (if used) from the primary node when the agent is stopped. As a convenience for testing, the primary agent will acquire the VIP at startup if the node does not already have it, but otherwise starting and stopping Failover Manager has no effect on the node having the virtual IP address.
```
josh-heyer, line 19: I suspect some of the confusion here ties into the final note in this file (line 156):

> The primary agent drops the VIP as needed during a failover, and Failover Manager verifies that the VIP is no longer available before promoting a standby. A failure of the bind address network leads to primary isolation and failover.

We should probably try to clarify that. Perhaps the two notes could be combined into a couple of paragraphs that lay out more clearly the scenarios in which EFM will drop / acquire a VIP?

product_docs/docs/efm/4/upgrading.mdx
EFM-Bobby: It's so they don't think there is db downtime when upgrading efm.

product_docs/docs/efm/4/04_configuring_efm/05_using_vip_addresses.mdx
EFM-Bobby: I'd rather make the smallest number of changes possible. Stopping efm != a failure -- I just want to let people know stopping efm doesn't break the cluster, and that's all separate from efm behavior while running.

product_docs/docs/efm/4/04_configuring_efm/05_using_vip_addresses.mdx
```diff
@@ -107,9 +111,10 @@ When instructed to ping the VIP from a node, use the command defined by the `pin
    rtt min/avg/max/mdev = 0.023/0.025/0.029/0.006 ms
    ```
 
-   No packet loss occurs. 
+   You will see 0% packet loss indicating the IP now reaches the primary node.
```
EFM-Bobby, line 114: I left a note here that I no longer see. This change isn't correct: there is no primary node in the testing scenario.

josh-heyer: That conflicts with step #2 above then, which has you running the script from the primary node!

EFM-Bobby: As discussed in slack, we will not handle any changes related to testing in this PR. Only the notes about not dropping the VIP at shutdown are relevant here.

product_docs/docs/efm/4/upgrading.mdx
josh-heyer: Let's be explicit about that then and give the note context:

```suggestion
    The primary agent will not drop the virtual IP address (if used) when it is stopped; the database will remain up and accessible on the VIP during the EFM upgrade. See also: [Using Failover Manager with virtual IP addresses](04_configuring_efm/05_using_vip_addresses.mdx) 
```

josh-heyer (approved):

PR #4958, opened 2023-10-30
Reorganize the EFM virtual IP addresses topic

josh-heyer:
## What Changed?

- collect all behavioral notes up-front
- link to all cited configuration parameters
- Cleanly delineate documentation of the control script
- Clarify expectations for each step of testing procedure

Goal is to fully summarize the behavior of EFM when configured for use with VIP in the top sections of the topic, without relying on notes elsewhere (testing section, configuration properties reference) to understand the basic behavior.

Builds on work started by @EFM-Bobby in #4957 

Preview: https://deploy-preview-4958--edb-docs-staging.netlify.app/docs/efm/latest/04_configuring_efm/05_using_vip_addresses/

### DIFF
--- a/product_docs/docs/efm/4/04_configuring_efm/05_using_vip_addresses.mdx
+++ b/product_docs/docs/efm/4/04_configuring_efm/05_using_vip_addresses.mdx
@@ -10,13 +10,39 @@ legacyRedirectsGenerated:
 
 <div id="using_vip_addresses" class="registered_link"></div>
 
-Failover Manager uses the `efm_address` script to assign or release a virtual IP address.
+Failover Manager can be used along with a virtual IP address (VIP) for routing requests to the current primary node. 
 
 !!! Note "Cloud provider support and alternatives"
     Virtual IP addresses aren't supported by many cloud providers. In those environments, use another mechanism, such as an elastic IP address on AWS, that can be changed when needed by a fencing or post-promotion script.
 
-!!! Note "Behavior when the agent is stopped"
-    Failover Manager will not drop the Virtual IP address (if used) from the primary node when the agent is stopped. As a convenience for testing, the primary agent will acquire the VIP at startup if the node does not already have it, but otherwise starting and stopping Failover Manager has no effect on the node having the virtual IP address.
+## Behavior when the EFM agent is stopped
+
+Failover Manager will not drop the virtual IP address from the primary node when the agent is stopped. As a convenience for testing, the primary agent will acquire the VIP at startup if the node does not already have it, but otherwise starting and stopping Failover Manager has no effect on the node having the virtual IP address. This allows for upgrades and maintenance of EFM services without interrupting access to the database.
+
+## Behavior during promotion of a node from standby to primary
+
+The VIP should be initially assigned to the primary node. When EMF detects failure of the primary node, the VIP will be released and then assigned to a standby node as it is promoted to be the new primary. 
+
+EFM verifies (via the command configured via the [`ping.server.command`](01_cluster_properties.mdx#ping_server_command) cluster property) that the VIP is not currently in use during promotion of a standby, and will skip assigning it to the new primary node if it is already assigned. This behavior can be disabled via the [`check.vip.before.promotion`](01_cluster_properties.mdx#check_vip_before_promotion) cluster property. 
+
+!!!important "Meaning of the ping command exit code"
+   Failover Manager uses the exit code of the ping command to determine whether an address is reachable. A zero exit code indicates the address is reachable (in this context, this means the VIP is assigned). A non-zero exit code indicates the address isn't reachable (in this context, this means the VIP is unassigned). 
+   
+   This matches the behavior of the standard [`ping(8)`](https://manned.org/ping.8) command; if you configure a different command via the [`ping.server.command`](01_cluster_properties.mdx#ping_server_command) cluster property, it should also conform to this behavior.
+
+## Configuring Postgres when using multiple addresses for nodes
+
+If a VIP address or any address other than the `bind.address` is assigned to a node, the operating system can choose the source address used when contacting the database. Be sure to modify the `pg_hba.conf` file on all monitored databases to allow contact from all addresses within your replication scenario.
+
+## Using multiple interfaces
+
+The network interface used for the VIP doesn't have to be the same interface used for the Failover Manager agent's [`bind.address`](01_cluster_properties.mdx#bind_address) value. The primary agent drops the VIP as needed during a failover, and Failover Manager verifies that the VIP is no longer available before promoting a standby. A failure of the bind address network leads to primary isolation and failover.
+
+If the VIP uses a different interface from the `bind.address`, you might encounter a timing condition in which the rest of the cluster checks for a reachable VIP before the primary agent drops it. In this case, Failover Manager retries the VIP check for the number of seconds specified in the [`node.timeout`](01_cluster_properties.mdx#node_timeout) property to help ensure that a failover happens as expected.
+
+## The efm_address script
+
+Failover Manager uses the `efm_address` script to assign or release a virtual IP address.
 
 By default, the script resides in:
 
@@ -54,12 +80,12 @@ For more information about properties that describe a virtual IP address, see [T
 
 Invoke the `efm_address` script as the root user. The efm user is created during the installation and is granted privileges in the sudoers file to run the `efm_address` script. For more information about the `sudoers` file, see [Extending Failover Manager permissions](04_extending_efm_permissions/#extending_efm_permissions).
 
-!!! Note
-    If a VIP address or any address other than the `bind.address` is assigned to a node, the operating system can choose the source address used when contacting the database. Be sure to modify the `pg_hba.conf` file on all monitored databases to allow contact from all addresses within your replication scenario.
-
 ## Testing the VIP
 
-When using a virtual IP (VIP) address with Failover Manager, it's important to test the VIP functionality manually before starting Failover Manager. This catches any network-related issues before they cause a problem during an actual failover. While testing the VIP, make sure that Failover Manager isn't running.
+When using a virtual IP (VIP) address with Failover Manager, it's important to test the VIP functionality manually before starting Failover Manager. This catches any network-related issues before they cause a problem during an actual failover. 
+
+!!!important
+   While testing the VIP, make sure that Failover Manager isn't running.
 
 The following steps test the actions that Failover Manager takes. The example uses the following property values:
 
@@ -73,7 +99,7 @@ ping.server.command=/bin/ping -q -c3 -w5
 !!! Note
     The `virtual.ip.prefix` specifies the number of significant bits in the virtual IP address.
 
-When instructed to ping the VIP from a node, use the command defined by the `ping.server.command` property.
+When instructed to ping the VIP from a node, use the command defined by the [`ping.server.command`](01_cluster_properties.mdx#ping_server_command) property and run it from the machine configured in EFM for the appropriate role (primary / secondary / witness).
 
 1. Ping the VIP from all nodes to confirm that the address isn't already in use:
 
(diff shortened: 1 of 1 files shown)


EFM-Bobby:
This all makes more sense now that I can look at the whole doc instead of little bits yesterday while trying to add the note about not dropping the vip. Thanks for handling this separately.

product_docs/docs/efm/4/04_configuring_efm/05_using_vip_addresses.mdx
```diff
@@ -10,13 +10,39 @@ legacyRedirectsGenerated:
...
 
-!!! Note "Behavior when the agent is stopped"
-    Failover Manager will not drop the Virtual IP address (if used) from the primary node when the agent is stopped. As a convenience for testing, the primary agent will acquire the VIP at startup if the node does not already have it, but otherwise starting and stopping Failover Manager has no effect on the node having the virtual IP address.
+## Behavior when the EFM agent is stopped
+
+Failover Manager will not drop the virtual IP address from the primary node when the agent is stopped. As a convenience for testing, the primary agent will acquire the VIP at startup if the node does not already have it, but otherwise starting and stopping Failover Manager has no effect on the node having the virtual IP address. This allows for upgrades and maintenance of EFM services without interrupting access to the database.
+
+## Behavior during promotion of a node from standby to primary
+
+The VIP should be initially assigned to the primary node. When EMF detects failure of the primary node, the VIP will be released and then assigned to a standby node as it is promoted to be the new primary.
```
EFM-Bobby, line 24: EMF -> EFM :)

product_docs/docs/efm/4/04_configuring_efm/05_using_vip_addresses.mdx
```diff
@@ -10,13 +10,39 @@ legacyRedirectsGenerated:
...
 
-Failover Manager uses the `efm_address` script to assign or release a virtual IP address.
+Failover Manager can be used along with a virtual IP address (VIP) for routing requests to the current primary node. 
 
 !!! Note "Cloud provider support and alternatives"
     Virtual IP addresses aren't supported by many cloud providers. In those environments, use another mechanism, such as an elastic IP address on AWS, that can be changed when needed by a fencing or post-promotion script.
 
-!!! Note "Behavior when the agent is stopped"
-    Failover Manager will not drop the Virtual IP address (if used) from the primary node when the agent is stopped. As a convenience for testing, the primary agent will acquire the VIP at startup if the node does not already have it, but otherwise starting and stopping Failover Manager has no effect on the node having the virtual IP address.
+## Behavior when the EFM agent is stopped
```
EFM-Bobby, line 18: Am concerned someone might think "is stopped" means "WHILE the agent is stopped" as opposed to the action of stopping it. It's obvious to you/me that efm doesn't do anything with the VIP while it's not running, but you never know how someone else might interpret it. Could it be more like "Behavior when the [primary] agent exits"? Am fine with/without "primary" in the text since it's clear from the description below.

josh-heyer: This is a great observation - I think let's be explicit about which process we're describing here. I'm gonna suggest "during shutdown" - this also aligns the wording with the section below.

I've also tweaked the paragraph that follows to further emphasize the distinction between the lifecycle of a *node* and the *agent* for that node.

product_docs/docs/efm/4/04_configuring_efm/05_using_vip_addresses.mdx
josh-heyer: d'oh!

product_docs/docs/efm/4/04_configuring_efm/05_using_vip_addresses.mdx
```diff
@@ -10,13 +10,41 @@ legacyRedirectsGenerated:
...
-    Failover Manager will not drop the Virtual IP address (if used) from the primary node when the agent is stopped. As a convenience for testing, the primary agent will acquire the VIP at startup if the node does not already have it, but otherwise starting and stopping Failover Manager has no effect on the node having the virtual IP address.
+## Behavior during shutdown of an EFM agent
+
+Failover Manager will not drop the virtual IP address from the primary node when the agent for that node shuts down. As a convenience for testing, the primary node's agent will *acquire* the VIP during startup if the node does not already have it, but otherwise starting and stopping Failover Manager has no effect on whether the node holds the virtual IP address. 
+
+This allows you to upgrade and perform maintenance on EFM services without interrupting access to the database.
+
+## Behavior during promotion of a node from standby to primary
+
+The VIP should be initially assigned to the primary node. When EFM detects failure of the primary node, it will release the VIP and then assign it to a standby node as that node is promoted to be the new primary.
```
EFM-Bobby, line 26: should be "...failure of the primary database..." and not "node." If the whole node fails, then it doesn't have the VIP any more anyway.

product_docs/docs/efm/4/04_configuring_efm/05_using_vip_addresses.mdx
```diff
@@ -10,13 +10,41 @@ legacyRedirectsGenerated:
...
+
+Failover Manager will not drop the virtual IP address from the primary node when the agent for that node shuts down. As a convenience for testing, the primary node's agent will *acquire* the VIP during startup if the node does not already have it, but otherwise starting and stopping Failover Manager has no effect on whether the node holds the virtual IP address. 
+
+This allows you to upgrade and perform maintenance on EFM services without interrupting access to the database.
+
+## Behavior during promotion of a node from standby to primary
+
+The VIP should be initially assigned to the primary node. When EFM detects failure of the primary node, it will release the VIP and then assign it to a standby node as that node is promoted to be the new primary. 
+
+EFM verifies (via the command configured via the [`ping.server.command`](01_cluster_properties.mdx#ping_server_command) cluster property) that the VIP is not currently in use during promotion of a standby, and will skip assigning it to the new primary node if it is already assigned. This behavior can be disabled via the [`check.vip.before.promotion`](01_cluster_properties.mdx#check_vip_before_promotion) cluster property.
```
EFM-Bobby, line 28: "...skip assigning it to the new primary node if it is already assigned." It wouldn't already be assigned to a standby node since standbys aren't supposed to have the vip. I think what you're going for is that efm won't promote if it sees that the VIP is reachable. "...during promotion of a standby, and will not promote [etc]."

product_docs/docs/efm/4/04_configuring_efm/05_using_vip_addresses.mdx
```diff
@@ -10,13 +10,41 @@ legacyRedirectsGenerated:
...
+
+The network interface used for the VIP doesn't have to be the same interface used for the Failover Manager agent's [`bind.address`](01_cluster_properties.mdx#bind_address) value. The primary agent drops the VIP as needed during a failover, and Failover Manager verifies that the VIP is no longer available before promoting a standby. A failure of the bind address network leads to primary isolation and failover.
+
+If the VIP uses a different interface from the `bind.address`, you might encounter a timing condition in which the rest of the cluster checks for a reachable VIP before the primary agent drops it. In this case, Failover Manager retries the VIP check for the number of seconds specified in the [`node.timeout`](01_cluster_properties.mdx#node_timeout) property to help ensure that a failover happens as expected.
+
+## The efm_address script
+
+Failover Manager uses the `efm_address` script to assign or release a virtual IP address.
 
 By default, the script resides in:
```
EFM-Bobby, line 49: I don't know why we had "by default" in this text before. Can you just change it to say that's where the script is? It isn't supposed to be anywhere else. Glad this section is getting some attention. :)

product_docs/docs/efm/4/04_configuring_efm/05_using_vip_addresses.mdx
```diff
@@ -103,23 +133,26 @@ When instructed to ping the VIP from a node, use the command defined by the `pin
...
+4. Use the `efm_address del` command to release the address on the primary node and confirm the VIP was released with the `ip address` command:
 
-   ```text
-   # efm_address del eth0 172.24.38.239/24
-   # ip address
+   ```console
+   efm_address del eth0 172.24.38.239/24
+   ip address
+   __OUTPUT__
    eth0 Link encap:Ethernet HWaddr 22:00:0A:89:02:8E
```
EFM-Bobby, line 156: Right after this there is some old text that says "The output from this step doesn't show an eth0 interface." Can you fix that? The interface is still there, just not the virtual IP address.

EFM-Bobby:
@josh-heyer Am all done with comments -- only found a couple technical things, half of which were in the original but hiding until you gave this some attention.

product_docs/docs/efm/4/04_configuring_efm/05_using_vip_addresses.mdx
josh-heyer: Good call!

product_docs/docs/efm/4/04_configuring_efm/05_using_vip_addresses.mdx
josh-heyer: Yeah, I worded that badly. AFAIK, EFM *will* still promote the node, but skip VIP assignment if the VIP is already reachable. Also there's the retry / monitor mentioned below that handles cases where async behavior results in the old primary node hanging onto it longer than expected. I've reworded.

product_docs/docs/efm/4/04_configuring_efm/05_using_vip_addresses.mdx
josh-heyer: Wondered about that :)

product_docs/docs/efm/4/04_configuring_efm/05_using_vip_addresses.mdx
josh-heyer: 👍

josh-heyer:
I don't know that we want to tackle this now @EFM-Bobby, but I noticed in EFM's VIP monitoring code logic that aims to automatically drop VIP assignment if it finds the VIP assigned to a non-primary node. I'm not sure this is worth mentioning because it doesn't seem like behavior that should be depended on for configuration (I assume it is aimed at ensuring that failover happens reliably) but figured I'd mention it in case you know more about the intent here.

EFM-Bobby:
> I don't know that we want to tackle this now @EFM-Bobby, but I noticed in EFM's VIP monitoring code logic that aims to automatically drop VIP assignment if it finds the VIP assigned to a non-primary node. I'm not sure this is worth mentioning because it doesn't seem like behavior that should be depended on for configuration (I assume it is aimed at ensuring that failover happens reliably) but figured I'd mention it in case you know more about the intent here.

EFM-Bobby:
@josh-heyer Whoops was trying to close a comment and closed the PR. Reopened. Sorry.

product_docs/docs/efm/4/04_configuring_efm/05_using_vip_addresses.mdx
```diff
@@ -10,36 +10,62 @@ legacyRedirectsGenerated:
...
+
+This allows you to upgrade and perform maintenance on EFM services without interrupting access to the database.
+
+## Behavior during promotion of a node from standby to primary
+
+The VIP should be initially assigned to the primary node. When EFM detects failure of the primary node's database, it will release the VIP and then assign it to a standby node as that node is promoted to be the new primary. 
+
+EFM verifies (via the command configured via the [`ping.server.command`](01_cluster_properties.mdx#ping_server_command) cluster property) that the VIP is not currently in use during promotion of a standby, and will not assign it to the new primary node until the ping indicates it is unreachable. You can disable this behavior via the [`check.vip.before.promotion`](01_cluster_properties.mdx#check_vip_before_promotion) cluster property.
```
EFM-Bobby, line 28: It not only won't assign it, it won't promote -- that's the important part. "...during promotion of a standby, and will not promote if the address is reachable." You've added a description of some logic here that isn't what efm really does.

EFM-Bobby: There is some logic to keep trying the vip check briefly, but the simplest thing to write is that efm won't promote if the VIP is reachable.

EFM-Bobby:
Ohhh..... if  you mean not worth mentioning _in the guide_, yes, i don't think it's worth mentioning. A lot of efm code is there to try to catch user error, and I don't think we gain much by trying to describe it all.

product_docs/docs/efm/4/04_configuring_efm/05_using_vip_addresses.mdx
josh-heyer: Ok, I see where this is going. Makes sense!

EFM-Bobby (approved):
Looks great -- thanks for giving this some time.

PR #4997, opened 2023-11-09
Update installer metadata; update links and version numbers for v16 ODBC driver

josh-heyer:
## What Changed?

Includes generated installer files.

I've chosen to omit the version number from the index description metadata, as this is tedious to keep updated with point releases. If desired, I recommend re-adding only a major version.

Could use a sanity check on that compat table; I just guessed. Primary goal was to make the links work, but realistically... Do we need to list the v13 stuff there? Maybe just a link to the v13 table...

### DIFF
--- a/install_template/config.yaml
+++ b/install_template/config.yaml
@@ -162,55 +162,55 @@ products:
     platforms:
       - name: CentOS 7
         arch: x86_64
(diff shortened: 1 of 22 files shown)


nidhibhammar (approved):
LGTM

PR #5003, opened 2023-11-14
Fix missing () in pg_control_init

djw-m:
## What Changed?

Missing () in pg_control_init as per https://edb.slack.com/archives/CSE2UK3EX/p1699952398387459


JohnCDB (commented):
Looks good to me. This will now work for both Postgres Extended and EPAS.

PR #5012, opened 2023-11-16
BDR-4288 - Docs: Improve role management documentation

djw-m:
## What Changed?

Text expanded to note where replication and joining may break when unreplicated roles or objects dependent on them are created.


LenzGr (approved):
LGTM! Thank you for the clarification.

PR #5019, opened 2023-11-17
BDR: remove reference to the "bdr-enterprise" extension in 5.x docs

ibarwick:
## What Changed?

As of BDR 4.0, the referenced `bdr-enterprise` extension no longer exists, and the catalogue table in question is always installed.

(I checked the BDR 4.x docs, it is not present there)


djw-m (approved):
LGTM

PR #5040, opened 2023-11-24
docs/tde/latest/enabling_tde - Adding -pbkdf2 to Examples

yuki-tei:
Hi team,
Here is our suggestion

## What Changed?

Added -pbkdf2 option to Example section. 
This might avoid "deprecated key derivation used." by openssl.

Kind Regards,
Yuki Tei


djw-m:
We have been advised that, because of openssl's variability in command line arguments and their effect, we should not change the given commands in the documentation. Not all openssl versions emit the warning.

yuki-tei:
Hi team,
Thanks for your reply. We understood.

However, there seems to be 2 samples on the same page, which is with -pbkdf2 and another is without. Could you think of making a consistent description to avoid unexpected confusion?

Kind Regards,
Yuki Tei

PR #5054, opened 2023-11-29
 BDR-4042-Docs-Update-release-notes-with-released-date 

djw-m:
## What Changed?

Added dates from index page
Normalized dates to 3 char month name as per standard.


LenzGr (approved):
LGTM! Thanks for taking care of this.

PR #5071, opened 2023-12-06
Docs/pgd/fix/bdr 4374 docs bdr sensitive to shared libraries ordering

djw-m:
## What Changed?

* Added text to the settings page (apparently, only configuration reference).


product_docs/docs/pgd/5/postgres-configuration.mdx
```diff
@@ -16,7 +16,8 @@ To run correctly, PGD requires these Postgres settings:
 
 -   `wal_level` &mdash; Must be set to `logical`, since PGD relies on logical decoding.
 -   `shared_preload_libraries` &mdash; Must contain `bdr`, although it can contain
-     other entries before or after, as needed. However, don't include `pglogical`.
+     other entries before or after, as needed. Some libraries, such as `pgaudit`,
+     must come after `bdr`. Don't include `pglogical` in the list.
```
mwanner, line 20: Is "some libraries" not too vague? Do we know of any other libraries that need to be loaded in any specific order?

I suggest we ask for `bdr` to be loaded first. Assuming that helps us tweak things ahead of time similar to what we did with `pgaudit`.

mwanner: @PJMODOS Agree to that wording? Or are there any extensions that we prefer to load *prior* to loading BDR? Maybe ACI's agent86 thing?

PR #5042, opened 2023-11-27
EPAS - Version fix to default component locations

nidhibhammar:
## What Changed?

https://enterprisedb.atlassian.net/browse/DB-2550

PR #5076, opened 2023-12-07
Error in component locations - move from 15 to 16

djw-m:
Discovered when doing clean install

## What Changed?

The log directory location


djw-m:
This will likely need to be merged into other work.

josh-heyer:
Covered by #5042

PR #5089, opened 2023-12-14
Fixed 4.3.3 required API versions

djw-m:
## What Changed?

Synced the 4.3.3 notes to the API breakage notes.

Fix for EW-10337


irionr (commented):
This is not actually true. 
The required API version requires a particular postgres version ( containing some particular changes). We enforce that now as API virtual packages and we don't need to add it in the docs. 
This line here instead is related to a change in EPAS (not bdr) that was later handled in bdr. So the latest versions of EPAS don't actually require the latest version of BDR.

djw-m:
Hmmm.... that seems to be something super hard to get over to users. What they see is version numbers in the upgrade docs out of sync with release notes. 

And in this case, the 4.3.3 requires versions note had versions that were before the previous release notes, as if the dependancy reversed.

product_docs/docs/pgd/4/rel_notes/pgd_4.3.3_rel_notes.mdx
```diff
@@ -10,7 +10,7 @@ Released: 14 Nov 2023
 EDB Postgres Distributed version 4.3.3 is a patch release of EDB Postgres Distributed 4, which includes bug fixes for issues identified in previous versions.
 
 !!! Note
-This version is required for EDB Postgres Advanced Server versions 12.14.18, 13.10.14, 14.7.0 and later.
+This version is required for EDB Postgres Advanced Server versions 12.15, 13.11, 14.10, 15.5 and later.
```
irionr, line 13: This is not true. 
This version is **not** _required_ for those EPAS versions. 
What **is** _required_ is that users do not use bdr < 4.3  with EPAS >= 12.14.18, 13.10.14, 14.7.0. 
I don't know what is the best way to let customers know this. 
Also, I think the critical period has passed. I don't think anyone would install one new EPAS version and an old BDR. 
This is one-year-old issue. 
For the sake of correctness, should we still keep it in the release notes?

djw-m:
So both versions old and replacement are wrong,  in which case, we should
remove the entire line.

irionr (approved):

PR #5096, opened 2023-12-19
Updates needed to the extensions by offering table - also sheet switch

djw-m:
## What Changed?

Updated against new sheet for extensions by offering.
pljava (11) removed
dbmsscheduler for BA EPAS set to true


catsouthard (approved):
LGTM, thank you

PR #5110, opened 2023-12-22
Update 04_efm_client_connect_failover.mdx

vadius45:
## What Changed?
Add drivers for popular languages python / golang


josh-heyer (approved):
Thanks for the additions, @vadius45 ! I've removed lib/pq before merging, as I can't find any indication that it can actually be used here, and given it's in maintenance mode I don't see that changing. If you have reason to believe otherwise, happy to reconsider!

PR #5124, opened 2024-01-02
Correct PEM compatibility table

sonotley:
PGE is supported as a backend for 13 and 14

11 is now out of support


nidhibhammar (approved):
LGTM

Issue #5035, opened 2023-11-22
Feedback on pgd 5 - "Backup and recovery" should reference postgresql.conf / postgresql.auto.conf, NOT recovery.conf


harada11111:
### What topic do you have feedback on?

https://github.com/EnterpriseDB/docs/blob/main/product_docs/docs/pgd/5/backup.mdx

### Your feedback

Seeing the Postgresql documents , "recovery.conf" cannot be used in PostgreSQL versions 12 and later.
It prevents the instance startup.

Is this supported for EDB ? 

Also would like you add some sample process (preparation for recovery/DR , command samples for PITR , node recovery  ,etc ) on this page.

Thanks in advance.

josh-heyer:
Not sure what you're encountering here - recovery.conf should be usable as documented. Could you clarify?

josh-heyer:
Ah, I see what you're talking about. Yes, we need to correct this (and should probably also add a note to the 3.7 BDR docs) - good catch, thanks!

PR #5125, opened 2024-01-02
Replace use of recovery.conf outside of PGD 3.7

josh-heyer:
## What Changed?

Bare minimum fixes for #5035 - remove references to recovery.conf except in PGD 3.7 where it is still nominally supported 

(Note: originally contained a somewhat iffy change to PGD 5 in a now-eliminated file.)


josh-heyer:
I think this is sufficient to avoid misleading information for now @djw-m

djw-m (approved):
LGTM to address the immediate issue.

PR #5139, opened 2024-01-10
Update managing_replicas.mdx

amrita-suresh:
clarify that the user can create a faraway replica in the same region as the source cluster


djw-m (approved):
LGTM

PR #5140, opened 2024-01-11
PGD: fix description of bdr.node_group_type() function

ibarwick:
## What Changed?

From PGD 5, the default group name is "global", not "normal".


djw-m (approved):
LGTM

PR #5141, opened 2024-01-11
PGD: add "bdr.node_group_summary" reference entry.

ibarwick:
## What Changed?

Added "bdr.node_group_summary" reference entry, which has been missing since the view was added in BDR 5.0.0.

Cross-reference from "bdr.node_group" added too (and cleaned up the column formatting there).

BDR-4467

### DIFF
--- a/product_docs/docs/pgd/5/reference/catalogs-visible.mdx
+++ b/product_docs/docs/pgd/5/reference/catalogs-visible.mdx
@@ -490,7 +490,9 @@ Currently configured conflict resolution for all known conflict types.
 
 ### `bdr.node_group`
 
-This catalog table lists all the PGD node groups.
+This catalog table lists all the PGD node groups. See also
+[`bdr.node_group_summary`](#bdrnode_group_summary) for a view containing
+user-readable details.
 
 #### `bdr.node_group` columns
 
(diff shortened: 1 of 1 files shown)


LenzGr:
Thank you! LGTM. What do you think about retroactively adding an item to the 5.0.0 release notes about this as well?

ibarwick:
> What do you think about retroactively adding an item to the 5.0.0 release notes about this as well?

Hmm, looking at the BDR release notes (for 5.0.0: https://www.enterprisedb.com/docs/pgd/latest/rel_notes/pgd_5.0.0_rel_notes ) they don't really go into that level of detail (i.e. noting new catalogue relations). It's not a major feature, from the context of the original commit ([50e02dfe](https://github.com/EnterpriseDB/bdr/commit/50e02dfe2939a453f2183093eed0ea45a15e8bf2)) I  assume it's covered under `Enhanced routing capabilities`. 

FWIW there are a bunch of other views and functions added in 5.0.0 (e.g. `bdr.node_group_routing_config_summary`, `bdr.node_group_routing_summary`) which are also not in the notes.

djw-m (approved):
LGTM

djw-m:
And I concur re release notes. The 5.0 release notes were quite high level and did not go down to table level.

PR #5144, opened 2024-01-12
PGD: rearrange internal functions reference

ibarwick:
## What Changed?

The [Internal system functions](https://www.enterprisedb.com/docs/pgd/latest/reference/functions-internal/) reference page starts off with the functions listed in alphabetical order, which breaks down part-way through, then there's a subsection titled "Task Manager Functions", which contains a couple of taskmanager functions, then a bunch of unrelated functions which were presumably just added to the end of the page.

As the functions start off in alphabetical order, a human scanning the list will tend to assume a function is not listed if it's not found where expected alphabetically. And it's a bit of a mess (I just wasted a bit of time assuming a function is not documented, when it was actually there).

This commit sorts the non-taskmanager functions alphabetically to make the page more usable.

It would probably make sense to group functions by purpose (view helpers, KV store, consensus, backwards compatibility) etc. but that is a larger refactoring.

### DIFF
--- a/product_docs/docs/pgd/5/reference/functions-internal.mdx
+++ b/product_docs/docs/pgd/5/reference/functions-internal.mdx
@@ -97,7 +97,7 @@ PGD message payload function that decodes the consensus payloads to a more human
 
 ### `bdr.decode_message_response_payload`
 
-PGD message payload function that decodes the consensus payloads to a more human-readable output.Used primarily by the `bdr.global_consensus_journal_details` debug view.
+PGD message payload function that decodes the consensus payloads to a more human-readable output. Used primarily by the `bdr.global_consensus_journal_details` debug view.
 
 ### `bdr.difference_fix_origin_create`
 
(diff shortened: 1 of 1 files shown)


product_docs/docs/pgd/5/reference/functions-internal.mdx
```diff
@@ -9,21 +9,162 @@ Listed below are internal system functions. Many are used in the creation of var
...
+!!! Warning
+    Don't use this function in user applications.
 
 ### `bdr.decode_message_payload`
 
 PGD message payload function that decodes the consensus payloads to a more human-readable output.Used primarily by the `bdr.global_consensus_journal_details` debug view.
 
+### `bdr.decode_message_response_payload`
+
+PGD message payload function that decodes the consensus payloads to a more human-readable output.Used primarily by the `bdr.global_consensus_journal_details` debug view.
```
djw-m, line 100:
```suggestion
PGD message payload function that decodes the consensus payloads to a more human-readable output. Used primarily by the `bdr.global_consensus_journal_details` debug view.
```

djw-m (approved):
LGTM

PR #5157, opened 2024-01-16
undo text->ini change

EFM-Bobby:
Looks like someone did a bulk change of "text" to "ini" and it clobbered the information about one efm property. This change adds the proper name/text back in.


djw-m (approved):
LGTM

PR #5174, opened 2024-01-23
Edits to PGD PR 5149

ebgitelman:


ebgitelman:
@djw-m I understand you want emphasis on the word "all," but I don't recommend using all caps. In general, we don't use emphasis (instead using a note when something needs to stand out), but I'm okay with the italics here without the all caps.

djw-m:
The issue that this started with was that people were not getting the need for all nodes to respond. Not sure how a lower case italic all would look here to resolve the essential issue.

djw-m:
A note could easily be skipped. I'm fairly happy with

**All** eligible data notes must agree to grant a global DML lock before the lock is granted.

Using bold over italic.

ebgitelman:
> A note could easily be skipped. I'm fairly happy with
> 
> **All** eligible data notes must agree to grant a global DML lock before the lock is granted.
> 
> Using bold over italic.

Cool. This should be ready to be merged then. Thanks!

djw-m (approved):
LGTM

PR #5176, opened 2024-01-23
Edits to Big Animal PR5123

ebgitelman:


ebgitelman:
Future tense seemed wrong in spots and it looked like there was missing content (version number?)

PR #5188, opened 2024-01-25
Update managing_cluster.mdx

amrita-suresh:
add that it may take time for the compute instance to shut down


nidhibhammar (approved):
LGTM

PR #5193, opened 2024-01-26
TDE - suggested fixes for CSP KMS

valeriodelsarto:
## What Changed?

Updating TDE docs that describe how to use CSP KMS to encrypt/decrypt using a wrapping key. Adding recent AZ CLI command (needs to be confirmed that the usage is correct from a TDE expert, I've not tested a TDE cluster locally I've just tested the AZ CLI commands)


gvasquezvargas (commented):
Hey @valeriodelsarto! 

It seems like this PR fell through the cracks. Is it still needed? 
If yes, I added a quick question for you to review, and then we would be ready to merge. 

Thank you!

product_docs/docs/tde/15/key_stores.mdx
```diff
@@ -95,13 +95,22 @@ PGDATAKEYUNWRAPCMD='crypt decrypt aws --in %p --region us-east-1'
...
+
+```shell
+PGDATAKEYWRAPCMD='az keyvault key encrypt --name pg-tde-master-1 --vault-name pg-tde --algorithm A256GCM --value @- --data-type plaintext --only-show-errors --output json | jq -r .result > "%p"'
+PGDATAKEYUNWRAPCMD='az keyvault key decrypt --name pg-tde-master-1 --vault-name pg-tde --algorithm A256GCM --value @"%p" --data-type plaintext --only-show-errors --output json | jq -r .result'
+```
+!!! Note
+    Shell commands with pipes, as in this example, are problematic because the exit status of the pipe is that of the last command. A failure of the first, more interesting command isn't reported properly. Postgres handles this somewhat by recognizing whether the wrap or unwrap command wrote nothing. However, it's better to make this more robust. For example, use the `pipefail` option available in some shells or the `mispipe` command available on some operating systems. Put more complicated commands into an external shell script or other program instead of defining them inline.
+
+Alternatively, you can use the [crypt utility](https://github.com/VirtusLab/crypt) to wrap and unwrap the data encryption key:
```
GabriFedi97, line 113: I strongly recommend not adding a reference to this project in our official documentation. That project is no longer maintained and uses outdated dependencies that result in unexpected behaviour.

gvasquezvargas (approved):
Technical review from @GabriFedi97  ✅ 
LGTM as well ✅

PR #5199, opened 2024-01-30
pgd: fixes for bdr.create_node_group() reference documentation

ibarwick:
## What Changed?

The values for the `node_group_type` parameter were changed in BDR 5.0 (commit 205c6aa1) but the documentation still describes the values valid for BDR 4.x and earlier.

Also while I was at it, generally improved the reference entry wording.

BDR-4554.


djw-m (approved):
LGTM apart from the reference to the main node group, which tends to be called the top level node group currently (though main node group does sound better to my ear)

PR #5201, opened 2024-01-30
PGD: Removed physical standby reference

djw-m:
## What Changed?

Removed limitations around physical standby from durability limitations (as physical standbys are no longer supported).


LenzGr (approved):
LGTM! Thanks.

PR #5237, opened 2024-02-06
Adding manual instructions for Postgres major version rolling upgrade - DOCS-95

jpe442:
@djw-m is working on a series of docs for TPA admin'd procedures on PGD. This PR covers the general procedure of a rolling upgrade of major versions of postgres on a PGD cluster. It then goes into extensive detailed steps to perform this rolling upgrade specifically from a TPA admin'd PGD cluster running PG 15 to a TPA admin'd PGD cluster running PG 16. 

Tracked in [JIRA by DOCS-95](https://enterprisedb.atlassian.net/browse/DOCS-95).


djw-m:
Is this a cluster deployed with TPA (I assume so) - if so, then say. You will probably have to also add notes on what to update after the upgrade to keep the TPA config.yml in sync with the currently upgraded cluster (otherwise future redeployments would go back a version).

The opening needs to frame *why* do a rolling upgrade and how PGD allows you to do so.

jpe442:
> Is this a cluster deployed with TPA (I assume so) - if so, then say. You will probably have to also add notes on what to update after the upgrade to keep the TPA config.yml in sync with the currently upgraded cluster (otherwise future redeployments would go back a version).
> 
> The opening needs to frame _why_ do a rolling upgrade and how PGD allows you to do so.

@djw-m Does one need to update anything in the config.yml other than this line:
postgres_version: '15'

djw-m:
That should be enough....

jpe442:
Would make the update _after_ upgrading all the nodes?

djw-m:
Yes - there’s a thing on reconciliation post updates in the  Tpa docs

jpe442:
Ok I found it here: https://www.enterprisedb.com/docs/tpa/latest/reference/reconciling-local-changes/

I will add a note in this procedure that TPA is being used in the worked example with the extra step of reconciliation and potentially link back to this page as well.

jpe442:
So I have successfully upgraded all 3 nodes of a PGD 5 PG 15 cluster to PG 16 according to the instructions as written in this PR, as updated now. Although I am able to upgrade each node individually, I am unable to successfully redeploy my cluster  (tpaexec deploy democluster) after changing the config.yml to the correct postgres version (i.e. from 15 -> 16). I am getting this error when I try to redeploy after making the change in the config.yml:

```
fatal: [localhost]: FAILED! => {
  "assertion": "cluster_tags['Owner'] not in [ '', 'root', 'postgres', 'ec2-user', 'admin', 'ubuntu', 'vagrant', 'tpa', 'user', ]",
  "changed": false,
  "evaluated_to": false,
  "msg": "Please set cluster_tags[Owner] or use '--owner Name' to identify the person responsible for this cluster"
}
```

Which is the same error I would get if I failed to set `--owner <owner_name>` when running `tpaexec provision democluster`. I have checked the TPA docs on reconciliation and I seem to be following instructions correctly (i.e. change the config.yml, then redeploy the cluster). I will dig further into the TPA docs and see if I missed something but this is where I am at now. I have tried adding `--owner jpe442` to the deploy command, which is the same owner I set when I provisioned the cluster, but that doesn't seem to do the trick.

MattWalliEDB:
What link can I use to see the taskbook procedures in the staging environment?  Specifically this major version upgrade procedure, I'd like to review this PR since I will likely be giving it to customers soon... but I can't find the rendered page using the nav on the docs staging link.

jpe442:
I have rebuilt the deployment so you, @MattWalliEDB, can see the preview.

MattWalliEDB:
Thanks @jpe442 .  I haven't tested the procedure, but reading through I don't see any information about performing a planned switchover prior to upgrading the write-leader node.  If the procedure is followed as currently written, a failover will occur (vs. a planned switchover) when postgres is stopped on whatever node is the current write leader.   "Rolling" availability needs to take into account pgd-proxy, the write-leader concept, and performing a planned switchover.  You can see an example of determining write-leader and performing planned switchover in the upgrade procedure I wrote for EPAS here:  https://github.com/EnterpriseDB/se-demos/blob/main/PGD/PGD%20POC%20Guide/PGD-Postgres-Major-Version-Upgrade-EPAS.md

MattWalliEDB:
Thanks @jpe442. No there isn't a need to do anything with pgd-proxy.  I like what you've done here with the refreshed content and I have some comments:

- I am not following why in the "rename the PGDATA directories" section you first create a directory with `sudo mkdir /opt/postgres/dataold` then rename an existing directory to the directory you just created `sudo -u postgres mv /opt/postgres/data /opt/postgres/dataold`.  Why not just rename data to dataold?  

- The permissions you're changing for ownership of /opt/postgres make me  nervous (you're changing ownership of /opt/postgres, presumably from root, to postgres).  Aren't those permissions put in place by TPA, presumably intentionally? I think you could just chown the /opt/postgres/datanew folder to postgres:postgres instead of chown'ing at the /opt/postgres level which overrides what TPA put in place.

- What should a user do if all does _not_ go well at some point during the procedure?  A customer will want to know how they could rollback/revert a node to the current major version in the event something went wrong while running the procedure.  I think they can do this by swapping (renaming) data directories again, moving `dataold` back to `data` and `data` to `datafailed` (or something like that).  This is the advantage of running the procedure this way, they can always revert.

- The cleanup steps at the end don't mention deleting the "dataold" folder.  This is a "use with caution" step, but we probably want to at least mention that they could remove the "dataold" folder once they've verified/tested and will no longer need to rollback (see prior comment on covering how to rollback)

-  Wouldn't it be logical to first upgrade all shadow (non write-leader) nodes then do a single switchover to upgrade the last (former write-leader) node?  That seems like the fastest approach because it only requires one switchover, vs possibly doing multiple switchovers.  

- As a suggestion, maybe they should upgrade just one shadow node then perform testing directly against that upgraded node (they could do this by performing the upgrade then connecting directly to the postgres/epas port on that node, without going through pgd-proxy and without making that upgraded node the write leader) in order to validate the new major version works as expected.

jpe442:
Thanks @MattWalliEDB for the comments. I think I have addressed them all and integrated them tidily into the document.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,510 @@
+---
+title: Performing a Postgres major version rolling upgrade 
+---
+
+## Performing a Postgres major version rolling upgrade
+
+This document outlines the procedure for performing a major version upgrade of a Postgres cluster using Postgres Distributed (PGD) as installed by Trusted Postgres Architect (TPA). The rolling upgrade process allows updating individual cluster nodes to a new major Postgres version while maintaining cluster availability and operational continuity. This approach minimizes downtime and ensures data integrity by allowing the rest of the cluster to remain operational as each node is upgraded sequentially. Upgrading your Postgres cluster to a newer major version is recommended to access improved features, performance enhancements, and security updates. This guide provides step-by-step instructions to facilitate a smooth and controlled upgrade process.
```
djw-m, line 7: Break this up. It's just a wall of text.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,510 @@
+---
+title: Performing a Postgres major version rolling upgrade
```
djw-m, line 2: Missing a navTitle and add deepToC: True to get more than two headings in the right hand nav bar

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,510 @@
...
+
+This document outlines the procedure for performing a major version upgrade of a Postgres cluster using Postgres Distributed (PGD) as installed by Trusted Postgres Architect (TPA). The rolling upgrade process allows updating individual cluster nodes to a new major Postgres version while maintaining cluster availability and operational continuity. This approach minimizes downtime and ensures data integrity by allowing the rest of the cluster to remain operational as each node is upgraded sequentially. Upgrading your Postgres cluster to a newer major version is recommended to access improved features, performance enhancements, and security updates. This guide provides step-by-step instructions to facilitate a smooth and controlled upgrade process.
+
+!!! Note
+Ensure that your databases and configuration files are backed up before starting the upgrade process to help prevent data loss.
+!!!
+
+To perform the rolling upgrade of the Postgres major version on your PGD cluster, carry out the following steps on each node in the cluster:
+
+- Confirm the current Postgres version.
```
djw-m, line 15: That's a bare heading. At least provide some direction for how you would do this.
"Versions can be viewed from PGD by logging in and running: `select * from bdr.group_version_details;`"

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,510 @@
...
+- Check to see if the target node is the write leader of the cluster.
```
djw-m, line 16: As above, repeat with the minimal direction.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,510 @@
...
+- Start Postgres.
+- Validate the new Postgres version. 
+- Test the upgrade directly against the newly upgraded node.
+- Cleanup post-upgrade. 
+
+We will look at these steps in more detail below, specifically for upgrading a node from Postgres 15 to Postgres 16. 
+
+## Worked example
+
+A three-node cluster (kaftan, kaolin, and kaboom) we [deployed on AWS using TPA](https://www.enterprisedb.com/docs/pgd/latest/quickstart/quick_start_aws/) is currently running Postgres 15 on a Linux OS (Debian in this case). We want to upgrade a single node to Postgres 16 using `bdr_pg_upgrade`, and then upgrade the rest of the cluster similarly. This walkthrough focuses on the detailed steps for the target node, kaboom.
```
djw-m, line 36: For this worked example, we start with a TPA cluster deployed using the AWS Quickstart. 

There are three nodes .....

We want to upgrade a single node to Postgres 16 using `bdr_pg_upgrade`, and then upgrade the rest of the cluster similarly. -> We want to upgrade a node to Postgres 16 as the first step of upgrading the entire cluster.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,510 @@
...
+
+A three-node cluster (kaftan, kaolin, and kaboom) we [deployed on AWS using TPA](https://www.enterprisedb.com/docs/pgd/latest/quickstart/quick_start_aws/) is currently running Postgres 15 on a Linux OS (Debian in this case). We want to upgrade a single node to Postgres 16 using `bdr_pg_upgrade`, and then upgrade the rest of the cluster similarly. This walkthrough focuses on the detailed steps for the target node, kaboom. 
+
+!!! Note
+Some steps of this process will involve you running commands as the Postgres owner. We will refer to this user as `postgres` throughout when appropriate. If you are running EDB Postgres Advanced Server, you will need to substitute that user with `enterprisedb` in all relevant commands.
+!!!
+
+### Confirm the current Postgres version
+
+Connect to any node that is still running Postgres 15 and use the following command to connect to PGD's `bdrdb` database:
```
djw-m, line 44: Connect how? psql, ssh, http?

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,510 @@
...
+ 2710197610 | kaboom    | 15.5 (Debian 15.5-1EDB.buster) | 5.3.0
+ 2111777360 | kaolin    | 15.5 (Debian 15.5-1EDB.buster) | 5.3.0
+(3 rows)
+```
+
+Confirm that the Postgres version is the expected version. Then, use the `exit` command to return to the node's command prompt.
+
+### Plan upgrade order
+
+The cluster must be available throughout the process (i.e., a *rolling* upgrade). To achieve continuous cluster availability, we must perform a [planned switchover](#performing-a-planned-switchover) of the write leader node before upgrading that node. We also want to keep the number of planned [switchovers](https://www.enterprisedb.com/docs/pgd/latest/cli/command_ref/pgd_switchover/) to a minimum for availability and efficiency. Accordingly, we will upgrade the current write leader last in our upgrade order, so we only have to perform one planned switchover.
```
djw-m, line 70: This is content which should be up top in the overall directions, not buried in the worked example. It applies to all rolling upgrades.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,510 @@
...
+
+The `pgd-show-groups` command shows all groups in the EDB Postgres Distributed cluster and their summary, including the current write leader.
+
+Our upgrade plan can be to upgrade kaboom, upgrade kaolin, perform a planned switchover from kaftan to kaboom, and finally upgrade kaftan.
+
+#### Performing a planned switchover
+
+Since our target node in this walkthrough is kaboom, we won't need to perform a planned switchover yet. 
+
+When you are ready to upgrade kaftan, switchover the write leader from kaftan to kaboom with the following command:
```
djw-m, line 95: Confusingly out of order.

Fix it by modifying the output of the show-groups to be kaboom, and then explaining that as kaboom is the write leader we will need to perform a switchover.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,510 @@
...
+-----        --------   ----   ------------ -------- ---- ------- ------------ 
+democluster  1935823863 global                       true false                
+dc1_subgroup 1302278103 data   democluster  dc1      true true    kaboom  
+```
+
+It is then safe to stop Postgres on kaftan.
+
+### Stop Postgres on the target node
+
+If you are not already on the target node, use SSH to connect to the target node, in this case, kaboom, and then run:
```
djw-m, line 120: Why not suggest they connect to the target node at the start?

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,510 @@
...
+
+```bash
+sudo systemctl stop postgres
+```
+
+This command halts the server on the current node, kaboom. Your cluster will continue running using the other two nodes.
```
djw-m, line 126: ... in this case, kaboom. 

Pausing review here.

djw-m (commented):
General changes tagged up until the Install new Postgres section.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
jpe442: Hmmm... yeah that is the order I had it in before, but because @MattWalliEDB suggested we would want to upgrade the shadow nodes before the write leader, I opted to have the worked example be a shadow node, as I thought the worked example should be the first node in the upgrade order. However, maybe the worked example can be of the last node to upgrade in the cluster, kaboom, which is the write-leader? I can give that a shot.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
jpe442: Is it a TPA cluster or a PGD cluster though? With what I had written I was trying to convey that it was a PGD cluster deployed using TPA, isn't that accurate?

djw-m: A pgd cluster deployed with TPA

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
djw-m: The flow is all wrong though - there should be advice to target shadow nodes first and a quick way to identify them but you will need to move the write leader on at some point

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
jpe442: I went ahead and added minimal direction to all of the steps in this next draft.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
jpe442: I think I figured out a way with doing the shadow nodes first but also bringing the important content to the general directions.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
jpe442: I changed it accordingly.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
jpe442: Moved it.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
jpe442: Changed to SSH.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
jpe442: Thanks!

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
jpe442: Yeah that needed breaking up.

product_docs/docs/pgd/5/admin-tpa/index.mdx
```diff
@@ -22,3 +22,5 @@ This section of the manual covers how to use TPA to deploy and administer EDB Po
   * Deploying the configuration with TPA
 
 The installing section provides an example cluster which will be used in future examples.
+
+You can also [perform a rolling major version upgrade](upgrading_major_rolling.mdx) with PGD administered by TPA.
```
ebgitelman, line 26: We don't use words like "guide," "section," "document," etc, as these are "documenting the documentation." I think I fixed the rest of the file when editing DJ's stuff but I'm not sure...

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -6,85 +6,85 @@ deepToC: true
...
 
-This document outlines the procedure for performing a major version upgrade of Postgres on a Postgres Distributed (PGD) cluster deployed by Trusted Postgres Architect (TPA).
+You can perform a major version upgrade of Postgres on a Postgres Distributed (PGD) cluster deployed by Trusted Postgres Architect (TPA).
 
 The rolling upgrade process allows updating individual cluster nodes to a new major Postgres version while maintaining cluster availability and operational continuity. This approach minimizes downtime and ensures data integrity by allowing the rest of the cluster to remain operational as each node is upgraded sequentially. 
 
-Upgrading your Postgres cluster to a newer major version is recommended to access improved features, performance enhancements, and security updates. This guide provides general step-by-step instructions and a worked example to help facilitate a smooth and controlled upgrade process.
+We recommend upgrading your Postgres cluster to a newer major version to access improved features, performance enhancements, and security updates. These step-by-step instructions and a worked example help to provide a smooth and controlled upgrade process.
```
ebgitelman, line 13: The one case we use first person.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -6,85 +6,85 @@ deepToC: true
...
 
 ### Plan the upgrade
 
-The cluster must be available throughout the process (i.e., a *rolling* upgrade). To achieve continuous cluster availability, you must [perform a planned switchover](#performing-a-planned-switchover) of the [write leader](https://www.enterprisedb.com/docs/pgd/latest/terminology/#write-leader) node before upgrading it so that there is always a write leader available. 
+The cluster must be available throughout the process (that is, a *rolling* upgrade). To achieve continuous cluster availability, you must [perform a planned switchover](#performing-a-planned-switchover) of the [write leader](https://www.enterprisedb.com/docs/pgd/latest/terminology/#write-leader) node before upgrading it so that a write leader is always available.
```
ebgitelman, line 17: House style is no latin abbreviations. 

"there is" tends to be a weak phrase

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -6,85 +6,85 @@ deepToC: true
...
 
-You can see which node is the write leader by connecting to the PGD cluster (e.g. using SSH) and running the `pgd show-groups` command:
+You can see which node is the write leader by connecting to the PGD cluster (for example, using SSH) and running the `pgd show-groups` command:
 
 ```bash
 sudo -u postgres pgd show-groups
 ```
 
-To keep the number of planned [switchovers](https://www.enterprisedb.com/docs/pgd/latest/cli/command_ref/pgd_switchover/) to a minimum, you should upgrade the current write leader last in your upgrade order, so you only have to perform [one planned switchover](#performing-a-planned-switchover) during the upgrade process.
+To keep the number of planned [switchovers](https://www.enterprisedb.com/docs/pgd/latest/cli/command_ref/pgd_switchover/) to a minimum, upgrade the current write leader last in your upgrade order, so you have to perform only [one planned switchover](#performing-a-planned-switchover) during the upgrade process.
```
ebgitelman, line 25: Don't use "should" as it introduces uncertainty. 

Misplaced modifier (only).

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -6,85 +6,85 @@ deepToC: true
...
 
 ### Perform the upgrade on each node
 
 !!! Note
-Ensure that your databases and configuration files are backed up before starting the upgrade process to help prevent data loss.
+To help prevent data loss, ensure that your databases and configuration files are backed up before starting the upgrade process.
```
ebgitelman, line 30: Goal before action

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -6,85 +6,85 @@ deepToC: true
...
 !!!
 
-Next, carry out the following steps on each node in order:
+Next, perform the following steps on each node in order:
```
ebgitelman, line 33: "carry out" is idiomatic and might not  be clear to non-native English speakers.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -6,85 +6,85 @@ deepToC: true
...
-- Validate the new Postgres version. 
-  - Connect to the bdrdb database and run `SELECT * FROM bdr.group_versions_details;`
-- Test the upgrade directly against the newly upgraded node.
-  - Run your test suite against while still SSH'd into the newly upgraded node.
-- Cleanup post-upgrade. 
+1. Validate the new Postgres version. 
+  - Connect to the `bdrdb` database and run `SELECT * FROM bdr.group_versions_details;`.
+1. Test the upgrade directly against the newly upgraded node.
+  - Run your test suite again while still connected by way of SSH into the newly upgraded node.
+1. Clean up post-upgrade.
```
ebgitelman, line 65: This should have been an ordered list.

You might want to also do numbers under some of these sub-steps.

I think adding the detail makes this a lot harder to read given that all this info is in the details that follow. Just my two cents on this addition.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -6,85 +6,85 @@ deepToC: true
...
 
-- Change the config.yml of the TPA-managed cluster to the new version.
+- Change the `config.yml` of the TPA-managed cluster to the new version.
   - `cluster_vars: postgres_version: '<new_version_number>'`
-- Use `tpaexec` to redeploy the PGD cluster with the updated config.yml.
+- Use `tpaexec` to redeploy the PGD cluster with the updated `config.yml`.
   - `tpaexec deploy <cluster_name>`
 
-We will look at these steps in more detail below, specifically for upgrading a [shadow node](https://www.enterprisedb.com/docs/pgd/latest/terminology/#write-leader) from Postgres 15 to Postgres 16. 
+Details of these steps are shown in the worked example, specifically for upgrading a [shadow node](https://www.enterprisedb.com/docs/pgd/latest/terminology/#write-leader) from Postgres 15 to Postgres 16.
```
ebgitelman, line 78: We don't write in first person but anyhow this text uses a mix. Let's stick with second person.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -6,85 +6,85 @@ deepToC: true
...
+This worked example starts with a [TPA-managed PGD cluster deployed using the [AWS Quickstart](https://www.enterprisedb.com/docs/pgd/latest/quickstart/quick_start_aws/). The cluster has three nodes: kaboom, kaolin, and kaftan, all running Postges 15.
 
-In this scenario, the current write leader is kaftan, and we will upgrade the first of two shadow nodes, kaboom.
+In this scenario, the current write leader is kaftan. This example upgrades the first of two shadow nodes, kaboom.
 
 !!! Note
-Some steps of this process will involve you running commands as the Postgres owner. We will refer to this user as `postgres` throughout when appropriate. If you are running EDB Postgres Advanced Server, you will need to substitute that user with `enterprisedb` in all relevant commands.
+Some steps of this process involve running commands as the Postgres owner. We refer to this user as postgres throughout, when appropriate. If you're running EDB Postgres Advanced Server, substitute the postgres user with enterprisedb in all relevant commands.
```
ebgitelman, line 87: First person is hard to avoid in this case.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -115,32 +116,32 @@ Confirm that the Postgres version is the expected version. Then, use the `exit`
...
 
-If you are not already on the target node, use SSH to connect to the target node, in this case, kaboom, and then run:
+If you're not already on the target node, use SSH to connect to the target node (in this case, kaboom), and then run:
 
 ```bash
 sudo systemctl stop postgres
 ```
 
-This command halts the server on the current node, in this case, kaboom. Your cluster will continue running using the other two nodes.
+This command halts the server on the current node, in this case, kaboom. Your cluster continues running using the other two nodes.
```
ebgitelman, line 125: No future tense unless it's truly a future action.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -203,22 +204,21 @@ When running the first or second command to stop the Postgres instances, you may
...
 
 ### Check that you can upgrade safely
 
-The `bdr_pg_upgrade` tool has a `--check` option, which dry runs some of the upgrade process. We can use this option to ensure the upgrade will go smoothly. 
-
-However, first, we need a directory for the files created by `bdr_pg_upgrade`, which we can give ownership of to the user `postgres`. 
+The `bdr_pg_upgrade` tool has a `--check` option, which performs a dry run of some of the upgrade process. You can use this option to ensure the upgrade goes smoothly.
 
-Accordingly, create an `/upgrade` directory in the `/home` directory and then give ownership of that directory to the user `postgres`:
+However, first you need a directory for the files created by `bdr_pg_upgrade`. For this example, create an `/upgrade` directory in the `/home` directory. Then give ownership of the directory to the user postgres.
```
ebgitelman, line 214: I'm not a fan of the word "accordingly." It sounds academic and it raises the reading level.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -438,20 +438,20 @@ Next, execute your test suite, including application transactions, queries, and
...
-We want to run a vacuum over to the database at this point as best practice. When the upgrade ran, you may have noticed the post-upgrade report included:
+As best practice, run a vacuum over the database at this point. When the upgrade ran, you may have noticed the post-upgrade report included:
 
 ```
 Once you start the new server, consider running:
     /usr/lib/postgresql/16/bin/vacuumdb --all --analyze-in-stages
 ```
 
-Let's do this now. On the target node, run:
+You can do run the vacuum now. On the target node, run:
```
ebgitelman, line 448: I don't know why I'm not able to edit this, but there's an extra word.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -461,11 +461,11 @@ sudo rm -r /opt/postgres/dataold
 
 After completing the upgrade on kaboom, run the same steps to upgrade the other shadow node, namely kaolin.
 
-After kaolin, only kaftan needs to be upgraded. However, because kaftan is our write leader, we should perform a planned switchover before running the upgrade steps on the node to ensure availability.
+After kaolin, only kaftan needs to be upgraded. However, because kaftan is the write leader, perform a planned switchover before running the upgrade steps on the node to ensure availability.
 
 #### Performing a planned switchover
 
-When you are ready to upgrade kaftan, SSH into kaftan and switchover the write leader from kaftan to an already upgraded shadow node, for example kaboom, with the following command:
+When you're ready to upgrade kaftan, SSH into kaftan and switch over the write leader from kaftan to an already upgraded shadow node, for example, kaboom:
```
ebgitelman, line 468: Verb form vs adjective form.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -438,20 +438,20 @@ Next, execute your test suite, including application transactions, queries, and
 
 ### Clean up post-upgrade
 
-We want to run a vacuum over to the database at this point as best practice. When the upgrade ran, you may have noticed the post-upgrade report included:
+As best practice, run a vacuum over the database at this point. When the upgrade ran, you may have noticed the post-upgrade report included:
```
ebgitelman, line 441: Was "over to" an error? It looked like it but my edit could be wrong...

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -94,13 +94,14 @@ SSH into kaboom and use the following command to connect to PGD's `bdrdb` databa
 ```bash
 sudo -u postgres psql bdrdb
 ```
-Then, when connected to bdrdb, use the following query:
+
+When connected to bdrdb, use the following query:
```
ebgitelman, line 98: Don't know why I can't edit, but bdrdb should be tagged.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
jpe442: I went ahead and went through this section to add more minimal direction at each step to be more tractable.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
jpe442: it was an error yes :-)

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
jpe442: I fixed it.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,537 @@
+---
+title: Performing a Postgres major version rolling upgrade on a PGD cluster built with TPA
+navTitle: Upgrade Postgres major version
+deepToC: true
+---
+
+## Overview
+
+You can perform a major version upgrade of Postgres on a Postgres Distributed (PGD) cluster deployed by Trusted Postgres Architect (TPA).
```
djw-m, line 9: Sounds like you are introducing a feature...

```suggestion
Upgrading a Postgres database's major version to access improved features, performance enhancements, and security updates is a common administration task. Doing the same for a Postgres Distributed (PGD) cluster deployed with Trusted Postgres Architect (TPA) is essentially the same process but performed as a rolling upgrade./
```

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,537 @@
...
+
+The rolling upgrade process allows updating individual cluster nodes to a new major Postgres version while maintaining cluster availability and operational continuity. This approach minimizes downtime and ensures data integrity by allowing the rest of the cluster to remain operational as each node is upgraded sequentially. 
+
+We recommend upgrading your Postgres cluster to a newer major version to access improved features, performance enhancements, and security updates. These step-by-step instructions and a worked example help to provide a smooth and controlled upgrade process.
+
+### Plan the upgrade
+
+The cluster must be available throughout the process (that is, a *rolling* upgrade). To achieve continuous cluster availability, you must [perform a planned switchover](#performing-a-planned-switchover) of the [write leader](https://www.enterprisedb.com/docs/pgd/latest/terminology/#write-leader) node before upgrading it so that a write leader is always available. 
+
+You can see which node is the write leader by connecting to the PGD cluster (for example, using SSH) and running the `pgd show-groups` command:
```
djw-m, line 19: Missing - discussion about groups and subgroups. You need to bear in mind that you need to know what group/subgroup the nodes are in because its this group you are going to be rolling upgrading.
Also may need a note for when there are two regions... and no regions (you are upgrading nodes in the top-level group only). We can discuss.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,537 @@
...
+
+```bash
+sudo -u postgres pgd show-groups
+```
+
+To keep the number of planned [switchovers](https://www.enterprisedb.com/docs/pgd/latest/cli/command_ref/pgd_switchover/) to a minimum, upgrade the current write leader last in your upgrade order, so you have to perform only [one planned switchover](#performing-a-planned-switchover) during the upgrade process.
```
djw-m, line 25:
```suggestion
To keep the number of planned [switchovers](https://www.enterprisedb.com/docs/pgd/latest/cli/command_ref/pgd_switchover/) to a minimum, put the current write leader last in your upgrade order. This will allow you to only perform [one planned switchover](#performing-a-planned-switchover) during the upgrade process.
```

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,537 @@
...
+
+### Perform the upgrade on each node
+
+!!! Note
+To help prevent data loss, ensure that your databases and configuration files are backed up before starting the upgrade process.
+!!!
+
+Next, perform the following steps on each node in order:
+
+1. If the target node is the current write leader, perform a [planned switchover](#plan-the-upgrade).
```
djw-m, line 35: Ok, couple of big comments here. This is the meat of the article, and it's one big step. And, what were numbered steps are not anymore because when you added notes to that broke the sequential count in markdown. Split them out into headings and make it sync with the worked example below.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,537 @@
...
+
+This worked example starts with a [TPA-managed PGD cluster deployed using the [AWS Quickstart](https://www.enterprisedb.com/docs/pgd/latest/quickstart/quick_start_aws/). The cluster has three nodes: kaboom, kaolin, and kaftan, all running Postges 15.
+
+In this scenario, the current write leader is kaftan. This example upgrades the first of two shadow nodes, kaboom.
+
+!!! Note
+Some steps of this process involve running commands as the Postgres owner. We refer to this user as postgres throughout, when appropriate. If you're running EDB Postgres Advanced Server, substitute the postgres user with enterprisedb in all relevant commands.
+!!!
+
+### Confirm the current Postgres version
```
djw-m, line 90: Should this not be in the plan your upgrade?

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,537 @@
...
+
+!!! Note
+To help prevent data loss, ensure that your databases and configuration files are backed up before starting the upgrade process.
+!!!
+
+Next, perform the following steps on each node in order:
+
+1. If the target node is the current write leader, perform a [planned switchover](#plan-the-upgrade).
+  - `sudo -u postgres pgd switchover --group-name <group_name> --node-name <new_write_leader_node_name>`
+1. Confirm the current major Postgres version.
```
djw-m, line 37: Probably want to move this "confirm the current postgres version" into planning

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,537 @@
...
+
+### Install the new Postgres version
+
+While still on the target node, kaboom, install the new version of Postgres (PG16) and the upgrade tool:
+
+```bash
+sudo apt install edb-bdr5-pg16 edb-bdr-utilities
+```
+
+### Initialize Postgres 16
```
djw-m, line 135:
```suggestion
### Initialize new Postgres version
```

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,537 @@
...
+  -D /opt/postgres/datanew \
+  -E UTF8 \
+  --lc-collate=en_US.UTF-8 \
+  --lc-ctype=en_US.UTF-8 \
+  --data-checksums
+```
+
+This command creates a PG16 data directory for configuration, `/opt/postgres/datanew`.
+
+### Copy configuration for Postgres 15 to Postgres 16
```
djw-m, line 157:
```suggestion
### Copy configuration from old Postgres to new Postgres
```

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,537 @@
...
+
+The next step copies the existing configuration (PG15) to the new PG16 data directory.
```
djw-m, line 159:
```suggestion
The next step copies the configuration files from the old Postgres version (PG15) to the new Postgres version's (PG16). Configuration files reside in each version's data directory.
```

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,537 @@
...
+Copy over the `postgresql.conf`, `postgresql.auto.conf`, and `pg_hba.conf` files and the whole `conf.d` directory:
+
+```bash
+sudo -u postgres cp /opt/postgres/data/postgresql.conf /opt/postgres/datanew/
+sudo -u postgres cp /opt/postgres/data/postgresql.auto.conf /opt/postgres/datanew/
+sudo -u postgres cp /opt/postgres/data/pg_hba.conf /opt/postgres/datanew/
+sudo -u postgres cp -r /opt/postgres/data/conf.d/ /opt/postgres/datanew/
+```
+
+### Rename the PGDATA directories
```
djw-m, line 170:
```suggestion
### Swap the data directories
```

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,537 @@
...
+
+Next, swap the PG15 and PG16 data directories: 
+
+```bash
+sudo mv /opt/postgres/data /opt/postgres/dataold
+sudo mv /opt/postgres/datanew /opt/postgres/data
+```
+
+!!! Important
```
djw-m, line 179: Important or a note. It's not the good path but something to refer to if it breaks.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,537 @@
...
+This rolls back/reverts the node back to the previous major version of Postgres.
+!!!
+
+### Stop all Postgres instances
+
+Although you previously stopped the Postgres service on the target node, kaboom, to be sure it's stopped, run the `stop` command for both instances of Postgres (PG15 and PG16) on
+kaboom:
+
+```bash
+sudo -u postgres /usr/lib/postgresql/15/bin/pg_ctl -D /opt/postgres/dataold stop
```
djw-m, line 196: Will that even work? I suspect only if it's not running. (maybe test with a running db and do the directory swap and then stop it?)

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,537 @@
...
+sudo -u postgres /usr/lib/postgresql/16/bin/pg_ctl -D /opt/postgres/data stop
+```
+
+!!! Note
```
djw-m, line 200: This is looking a little messy. Wonder if we can tidy it up a bit. @MattWalliEDB any thoughts on how you'd check servers were stopped without reading the chicken entrails of errors?

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,537 @@
...
+
+If you're not already on the target node, use SSH to connect to the target node (in this case, kaboom), and then run:
+
+```bash
+sudo systemctl stop postgres
+```
+
+This command halts the server on the current node, in this case, kaboom. Your cluster continues running using the other two nodes.
+
+### Install the new Postgres version
```
djw-m, line 127:
```suggestion
### Install the new Postgres version and upgrade utility
```

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,537 @@
...
+```
+
+### Validate the Postgres version
+
+Repeating the first step, check the version of Postgres to confirm that you upgraded kaboom correctly. While still on kaboom, run:
+
+```bash
+sudo -u postgres psql bdrdb
+```
+
```
djw-m, line 413: Or just `select version()` to return that nodes version? Might be simpler and less info to parse.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,537 @@
...
+### Test the upgrade
+
+After the upgrade is complete on kaboom, test your databases thoroughly to ensure the upgrade was successful. Being still connected to kaboom ensures that your tests run specifically against the new version without interference from cluster routing mechanisms.
+
+```bash
+sudo -u postgres psql <database_name>
+```
+
+Next, execute your test suite, including application transactions, queries, and other operations critical to your environment. Pay particular attention to performance metrics and compatibility with the new version. The insights gained from this controlled upgrade can inform your approach, helping to ensure a smoother transition for all nodes.
```
djw-m, line 437: Well thats all well and good but did we mention "You should have a test suite you can deploy on any node" in the preamble? 

Also isn't the node returning to the cluster at this point?

Have we checked it is/isn't returning?

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,537 @@
...
+
+### Clean up post-upgrade
```
djw-m, line 439: Ditto this vacuum step is somewhat scary. I'd check with someone (as I seriously don't know) whether a vacuum while joined to the cluster but not write leader could cause any issues.

djw-m (commented):
Lots to fix, structures need to sync up on both sides steps and example, and somethings might need to be moved or simplified.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,537 @@
...
+deepToC: true
+---
+
+## Overview
+
+You can perform a major version upgrade of Postgres on a Postgres Distributed (PGD) cluster deployed by Trusted Postgres Architect (TPA).
+
+The rolling upgrade process allows updating individual cluster nodes to a new major Postgres version while maintaining cluster availability and operational continuity. This approach minimizes downtime and ensures data integrity by allowing the rest of the cluster to remain operational as each node is upgraded sequentially. 
+
+We recommend upgrading your Postgres cluster to a newer major version to access improved features, performance enhancements, and security updates. These step-by-step instructions and a worked example help to provide a smooth and controlled upgrade process.
```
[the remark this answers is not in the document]
jpe442: Done.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
jpe442: I see how it makes sense to put it in the plan. I had thought it would be the first step of the node-by-node steps, such that we would check the versions each time we were on a node to ensure things were being updated as we expected. Let me see if I can balance both these goals.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
jpe442: Note seems better I agree.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
jpe442: Okay I will test this.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
jpe442: Yeah, I will add something about needing a test suite in the preamble. 

For your second/third point here, did we ever remove the node from the cluster in the first place? I don't think I documented that we need to remove the node only that we need to stop postgres on the node. I assumed PGD would just not use it with postgres turned off on the node and that it would use it again with postgres turned back on on the node. What do I need to do to remove and or rejoin a node? Rejoining presumably only happens after the test suite has been run, correct?

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
jpe442: Do we have an eng contact I can use or do I need to find one for this?

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
jpe442: Split the steps up into headings as opposed to using numbers? I am not 100% on what you are suggesting here.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
jpe442: Will do.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
jpe442: I am not following this 100%.  I could include a brief explanation of how groups and subgroups work in a PGD cluster. I thought the only relevant information using the show-groups command here was to learn which node is the write leader. We need to know which nodes belong to which group/subgroup? Could you explain this a bit more?

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
djw-m: Go look at the deployment.

Every step is numbered 1.

And there's a lot of steps.

Break them up.

Put headings on them.

Make the headings match the worked example's headings.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
djw-m: If you have two subgroups, you'll have a write leader in each group. (Possibly three, not sure if you could configure the top-level with routing too).... 

So, you need to know what sub-group you are upgrading because it has its own write leader and that's what the switchover will change.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
djw-m: Second/third point is about running a test suite on a node thats in the process of resyncing with the cluster. I don't recall this step in previous guides so not sure where this came from, but it feels potentially dangerous as a step.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
jpe442: 1. I see that the numbering isn't working; I could fix that by indenting the note under each by one, but it looks like you are suggesting to do away with the ordered list. 
2. When you say break up the steps and put headings on them, are you saying to break each step apart from one another with a little whitespace and make them headings or break the steps up into groups of steps and then put headings on each group of steps? 
3. Making the headings match the worked example's headings is a little tricky as the general steps aren't quite 1:1 with the worked example steps with respect to order. For instance, the first step in the general directions that apply to each node is to perform a planned switchover if the target node in question is the current write leader. However, this step is in ### Next Steps of the worked example because the worked example is the upgrade of a shadow node (as it would be weird to start with the write leader in the worked example.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
jpe442: It's only dangerous if the node is syncing back with its group at the time of the testing right? The step to test seems right as part of upgrading process, correct?

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
jpe442: Can we grep for the pid or something?

jpe442: Or maybe we can just use systemctl status postgresql?

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
jpe442: I am pulling the test suite content.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
jpe442: I am changing the step to verify that postgres is stopped and using `systemctl status postgres` as it is clear that the server isn't running using this method I believe.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
jpe442: Okay I will add a little something on this near the beginning on preparing/planning the upgrade.

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
jpe442: Reached out in #eng-pgd and this vacuum step is safe:

"This shouldn’t impact on data consistency as VACUUM isn’t replicated, but depending on the size of the database it may generate a bit of replication lag which shouldn’t take too long to catch up." William Ivanski

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,602 @@
...
+
+### Verify the Postgres service is inactive
+
+Although you should have stopped [previously stopped the Postgres service on the target node](#stop-postgres-on-the-target-node), kaboom, to be sure it's stopped, verify by running the `systemctl status postgres` command:
+
+```bash
+sudo systemctl status postgres
+```
+
+The output of the status command shows that the Postgres service has stopped running:
```
ebgitelman, line 275:
```suggestion
The output of the `status` command shows that the Postgres service has stopped running:
```

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,602 @@
...
+sudo rm -r /opt/postgres/dataold
+```
+
+You are now finished upgrading the target node.
+
+### Next steps
+
+After completing the upgrade on kaboom, run the same steps on kaolin and kaftan.
+
+If you followed along with this example and kaftan is the write leader, you must [perform a planned switchover](#perform-a-planned-switchover) to another, already upgraded, node before running the upgrade steps on kaftan to ensure availability.
```
ebgitelman, line 560:
```suggestion
If you followed along with this example and kaftan is the write leader, to ensure availability, you must [perform a planned switchover](#perform-a-planned-switchover) to another, already upgraded node before running the upgrade steps on kaftan.
```

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,602 @@
...
+
+#### Check PG versions across the cluster 
+
+After completing the upgrade on all nodes, while connected to one of the nodes, you can once again check your versions:
+
+```bash
+sudo -u postgres pgd show-version
+```
+
+The output should be similar to the following:
```
ebgitelman, line 570:
```suggestion
The output will be similar to the following:
```

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,602 @@
...
+
+You are now finished upgrading the target node.
+
+### Next steps
+
+After completing the upgrade on kaboom, run the same steps on kaolin and kaftan.
+
+If you followed along with this example and kaftan is the write leader, you must [perform a planned switchover](#perform-a-planned-switchover) to another, already upgraded, node before running the upgrade steps on kaftan to ensure availability.
+
+#### Check PG versions across the cluster
```
ebgitelman, line 562:
```suggestion
#### Check Postgres versions across the cluster 
```

ebgitelman (commented):
Let me know if you have any questions!

product_docs/docs/pgd/5/admin-tpa/upgrading_major_rolling.mdx
```diff
@@ -0,0 +1,602 @@
...
+
+### Perform the upgrade on each node
+
+!!! Note
+To help prevent data loss, ensure that your databases and configuration files are backed up before starting the upgrade process.
+!!!
+
+Using the [preliminary order](#prepare-the-upgrade), perform the following steps on each node while connected via SSH:
+
+* **Confirm the current Postgres version**
```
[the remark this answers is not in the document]
jpe442: I am just following the format DJ asked me too so that it is in line with what he is doing elsewhere in the PGD docs: https://www.enterprisedb.com/docs/pgd/latest/admin-manual/installing/05-creating-cluster/ 🤷

PR #5246, opened 2024-02-08
PG4K: Perform API HTML+link corrections on import

josh-heyer:
## What Changed?

A few corrections that are easier to make on import than fix at the source.
Also use a worktree for the script to avoid messing up the source repo (can now re-run local import without side-effects)


djw-m (dismissed):
LGTM (Though we need to discuss whether  is actually invalid.

PR #5249, opened 2024-02-09
Docs/fix/df-402 support disabling pdf generation and logo display

djw-m:
## What Changed?

Added hidePDF support to frontmatter. Does not descend directories.


josh-heyer (approved):
This is solid. +1 for remembering to add it to the schema :)

If we ever switch up and make this a toggle for *generating* the PDF, we'll probably want to invert it (e.g. `showPDFDownload: [yes|no]`) For now, the negative phrasing makes it pretty clear what the default is gonna be!

PR #5256, opened 2024-02-12
Added more information about upgrade-conf source flag

EFM-Bobby:
This removes a little information from the general 'efm' utility page and adds a note to see the upgrade page for more info instead. Hopefully this is better than having incomplete information spread out across pages. On the upgrade page there is now a whole section on the -source parameter and when root privileges are not required to run it.

There is also a note about what happens when files already exist in the target location.

The above changes are in response to a customer confused about when root privileges were required or not, and what happened when files are upgraded into the same directory as the source files (source files renamed). Any suggestions are welcome -- the "summary" could possibly be a table but that might be overkill.

### DIFF
--- a/product_docs/docs/efm/4/07_using_efm_utility.mdx
+++ b/product_docs/docs/efm/4/07_using_efm_utility.mdx
@@ -173,9 +173,9 @@ This command must be invoked by efm, a member of the efm group, or root.
 
 `efm upgrade-conf <cluster_name> [-source <directory>]`
 
-Invoke the `efm upgrade-conf` command to copy the configuration files from an existing Failover Manager installation and add parameters required by a Failover Manager installation. Provide the name of the previous cluster when invoking the utility. This command must be invoked with root privileges.
+Invoke the `efm upgrade-conf` command to copy the configuration files from an existing Failover Manager installation and add parameters required by a Failover Manager installation. Provide the name of the previous cluster when invoking the utility. If you're running Failover Manager in the default mode, this command must be invoked with root privileges.
 
-If you're upgrading from a Failover Manager configuration that doesn't use sudo, include the `-source` flag and specify the name of the directory in which the configuration files reside when invoking `upgrade-conf`.
+For information on the optional `-source` flag, or if you're upgrading from a Failover Manager configuration that doesn't use sudo, see [Upgrading Failover Manager](upgrading) for more information.
 
 ## efm node-status-json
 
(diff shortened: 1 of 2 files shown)


product_docs/docs/efm/4/07_using_efm_utility.mdx
```diff
@@ -173,9 +173,9 @@ This command must be invoked by efm, a member of the efm group, or root.
 
 `efm upgrade-conf <cluster_name> [-source <directory>]`
 
-Invoke the `efm upgrade-conf` command to copy the configuration files from an existing Failover Manager installation and add parameters required by a Failover Manager installation. Provide the name of the previous cluster when invoking the utility. This command must be invoked with root privileges.
+Invoke the `efm upgrade-conf` command to copy the configuration files from an existing Failover Manager installation and add parameters required by a Failover Manager installation. Provide the name of the previous cluster when invoking the utility. If you're running Failover Manager in the default mode, this command must be invoked with root privileges.
```
EFM-Bobby, line 176: The "default" mode is the one that uses sudo for various things. We don't have a good term for "sudo-using mode" I think. Hopefully the sentence below makes it more clear.

djw-m: It still feels somewhat confusing....

This command must be invoked with root privileges, by using sudo or logging in as root, if you are running Failover Manager in the default mode. 

Maybe?

EFM-Bobby: I think we moved to the "root privileges" thing a while back from the more wordy "as root or with sudo" idea, but now I don't remember. I'm fine with it though, will push the change. This command is really more of an installation kind of thing (as opposed to runtime), so it's the only one that needs this text.

djw-m: It's more about bringing up what you need to do to the front of the text, followed by the explanation, rather than explaining why you need to do what comes next.

EFM-Bobby: @djw-m Thinking a little more about it..... I think I remember now why we stopped using things like "root or with sudo." Depending on the system, there could be other users that actually have these privileges, and some customers run with a permissions system other than 'sudo.' We use the phrase "superuser privileges" in other parts of the docs. What do you think of this?

"This command must be invoked with superuser privileges if you are running Failover Manager in the default mode."

djw-m: Go for it.

EFM-Bobby: Pushed.....

djw-m (approved):
LGTM

PR #5197, opened 2024-01-29
Update pgpool_other_linux_9.mdx

david-olivari:
powertools is knows as `crb` in rocky 9 
```
[root@PG2 ~]# dnf config-manager --enable crb
[root@PG2 ~]# dnf install edb-pgpool44
Rocky Linux 9 - BaseOS           11 kB/s | 4.1 kB     00:00    
Rocky Linux 9 - AppStream        18 kB/s | 4.5 kB     00:00    
Rocky Linux 9 - CRB             2.0 MB/s | 2.3 MB     00:01    
Dependencies resolved.
================================================================
 Package      Arch   Version      Repository               Size
================================================================
Installing:
 edb-pgpool44 x86_64 4.4.3-1.el9  enterprisedb-enterprise 706 k
Installing dependencies:
 edb-pgpool44-libs
              x86_64 4.4.3-1.el9  enterprisedb-enterprise  59 k
 libmemcached-awesome
              x86_64 1.1.0-12.el9 crb                     110 k

Transaction Summary
================================================================
Install  3 Packages

Total download size: 874 k
Installed size: 3.7 M
Is this ok [y/N]: 

````

PR #5268, opened 2024-02-14
Change name of powertools to crb for Rocky and Alma Linux 9

dwicinas:
Changes the command for enabling powertools from "powertools" to "crb" for all products installed on Rocky or Alma Linux 9.

This PR also picks up a change to EPAS that must have been made in the templates but never propagated to all the versions of EPAS. I'm including it in this PR so all the templatized output is consistent.

This PR responds to [PR-5197](https://github.com/EnterpriseDB/docs/pull/5197) but I don't have access to that branch.

### DIFF
--- a/advocacy_docs/supported-open-source/postgresql/installing/linux_x86_64/postgresql_other_linux_9.mdx
+++ b/advocacy_docs/supported-open-source/postgresql/installing/linux_x86_64/postgresql_other_linux_9.mdx
@@ -40,7 +40,7 @@ Before you begin the installation process:
 - Enable additional repositories to resolve dependencies:
 
   ```shell
-  sudo dnf config-manager --set-enabled crb
+  sudo dnf config-manager --set-enabled powertools
   ```
 
 - Disable the built-in PostgreSQL module:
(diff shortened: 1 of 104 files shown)


dwicinas:
@nidhibhammar Please take a look at the files changed in this PR. The change for Rocky/Alma Linux 9 seems pretty straightforward, but it seems like we must have made a change to the templates for EPAS post-install configuration that we didn't propagate to all the various versions of EPAS. I'm including those changes here to keep the generated install files in sync with the templates. If you don't like that, I can work to separate out the Rocky/Alma Linux 9 change from the EPAS change and we can put them into two separate PRs. Adding @djw-m  because he seems to have a handle on the Rocky/Alma change.

djw-m (approved):
Reviewing the output, LGTM.

nidhibhammar:
@dwicinas, everything looks good except the one change you pointed out for EPAS. Let's fix that before moving ahead if it is not too much time-consuming.

nidhibhammar:
@dwicinas, yes, I meant to fix the psql output change. It should reflect the exact version it is. Previously it was 15.2.0 for all versions, of which we are unaware. Now that we know, let's fix it, and the psql output should show the exact version it is in. 
for eg: version 12 should show psql 12.x.0.

PR #5273, opened 2024-02-15
Edits to PWR PR5253

ebgitelman:


product_docs/docs/pwr/1/index.mdx
```diff
@@ -6,15 +6,9 @@ navigation:
...
-On the Postgres server, you should have `edb_wait_states` version 1.2 or later installed and loaded. The extension also must be created,
-preferably on the main database, so the `edb_wait_states` functions are available for Postgres Workload Report to collect saved snapshot data.
-
-For more information about PWR, see: 
-
-  - [Installing](installing)
-  - [Configuring](configuring)
-  - [Using](using)
+On the Postgres server, you must have `edb_wait_states` version 1.2 or later installed and loaded. You must also create the extension,
+preferably on the main database, so the `edb_wait_states` functions are available for Postgres Workload Report to collect saved snapshot data.
```
ebgitelman, line 14: It's not necessary to list a table of contents because our design provides that. As we add content, this becomes unruly (and unnecessary) on the landing page.

dwicinas: @ebgitelman I was responding to a comment from @djw-m who suggested that we can't expect users to find subordinate topics by looking at the left nav tree.

nidhibhammar: I agree that it is not necessary.

nidhibhammar (approved):
LGTM

product_docs/docs/pwr/1/index.mdx
djw-m: I would like to disagree - relying on the left hand nav bar which for navigation from the landing page of any top index is cognitive load for the reader. I've been applying true index pages in PGD and readers comment immediately on how they can find things more easily. 

That said, I would alter the layout and remove the "More information" presentation and use explanatory lines eg.

* Installing ....

ebgitelman:
A narrative explaining next steps could be added, as per our recent discussion. I'll be happy to look at that if it's added

dwicinas:
I think this is a place where the index cards should be used (assuming they work as expected). We put in a lot of work over the last two years standardizing (somewhat) the TOC structure for most products. I don't know that there's a lot more to be said about "Using" or "Installing". A boilerplate phrase of description in the index card seems sufficient. If the subordinate sections are less straightforward, then I would agree with DJ about the value of explanatory text.

PR #5274, opened 2024-02-15
Add note about acceptable PGD clients

eatonphil:


djw-m (commented):
Not sure if the Quickstart is the right place for this note.

PR #5286, opened 2024-02-20
Index cards for Postgres Workload Report

dwicinas:
Create simple index cards 
Repositioned release notes so they appear first


nidhibhammar:
This is how it looks on mobile.

Screenshot_20240222_111341_Samsung Internet

djw-m:
It looks fine, and I tested against the mobile device that gave the accessibility oddness, and that seems to have gone away with last night's firmware update, so I'm going to call it a transient bug. 

Next, we have to be consistent around this if you want index cards everywhere. :)

djw-m (approved):
LGTM

PR #5279, opened 2024-02-19
PEM - Updated the registry links

nidhibhammar:
## What Changed?

Updated the registry links

PR #5289, opened 2024-02-21
PEM: Fix some typos, clean up wording around Windows registry notes

josh-heyer:
## What Changed?

A few corrections to changes from #5279 - not sure what happened here, but some stray text got inserted in a couple of spots ([pem/8/managing_certificates.mdx](https://github.com/EnterpriseDB/docs/pull/5279/files#diff-4fc3d11f55e2a3a2bd17588c7ccf49728d638974c27b988a2159f914be3e2a56) and [pem/8/monitoring_performance/alerts.mdx](https://github.com/EnterpriseDB/docs/pull/5279/files#diff-0f904f524976e088e08d7277d571f58b69dfdec7c0ffc3dc6242a6126534f1ea)

While fixing that, I figured I'd standardize the phrasing around these notes - there no longer appears to be a difference between registry paths in 32-bit and 64-bit Windows, and AFAIK we don't support PEM on 32-bit Windows anymore anyway; these can all be simplified to just "Windows". Also there were a few places where double slashes were used (this likely came from the source, as in C++ strings double backslashes are needed to escape a single backslash). Finally, one place where spaces were used instead of slashes. 

...yes, I got very distracted from my original goal of fixing a build warning.


nidhibhammar (approved):
LGTM

PR #5291, opened 2024-02-21
Add styling for the "info" admonition tag

josh-heyer:
## What Changed?

This is almost exclusively used in the PG4K docs, where... Many should probably be SeeAlso instead. But regardless, in MkDocs the "info" styling is a dead ringer for "Note", so I've made them the same here too to avoid shock when writing for one system and then seeing it in the other.


djw-m (approved):
LGTM

PR #5296, opened 2024-02-22
pgd: remove deprecation notices for "check_constraints" node group option

ibarwick:
## What Changed?

This option was removed in BDR 3.7, but hastily reinstated with a deprecation notice as it turned out it was in use by at least one customer. Since then, the customer has continued using the option but no action has been or is likely to be taken to remove, replace or otherwise modify the behaviour of the option, and the presence of the deprecation notices over multiple major releases is causing ongoing confusion, so remove the notices.

BDR-4424.


djw-m (approved):
LGTM

PR #5298, opened 2024-02-22
Fix improperly indented admonitions

josh-heyer:
## What Changed?

Rather than continuing to discover these ad-hoc when they end up eating all content to (and through) the end of the topic, I've added build warnings for these. This PR includes fixes to a good big pile of them.

Reviewer notes: look at first commit to see only manually changed files; second commit is generated installer files.


djw-m (approved):
LGTM

Tho I still say we should make em explicitly fenced

PR #5305, opened 2024-02-26
pgd: improve reference for bdr.alter_node_group_option()

ibarwick:
## What Changed?

- Put items in alphabetical order.
  - for `bdr.alter_node_group_config()` it made sense to put each item in the order of the function parameters, but that's not relevant here, and alphabetical order is much easier to scan
- Note effective data type of each item.
  - the value of each item, even though parsed as `text`, will be parsed as a particular PostgreSQL data type, so note which one
- Note function return value.
  - I find this very useful to know when writing code which uses this function
- Add note to each item which affects top-level node group only.
- Note that function uses group consensus and is not transactional.
  - we had this for `bdr.alter_node_group_config()`, so carry through to this function, as it's the same
- Add entry for `route_writer_wait_flush`.
  - like `route_reader_max_lag`, this (AFAICT) is reserved for future use; both are listed in `bdr.node_group_summary`, so it's better to clarify the reason for their existence in case anyone wonders...
- Note that current node group options can be examined via the view `bdr.node_group_summary`.
  - this kind of cross-reference is IMO essential, particularly for a function which changes something and the caller would like to know how to verify that change
- Various smaller improvements/clarifications to individual option descriptions, including additional links to other documentation sections.

BDR-4671.

### DIFF
--- a/advocacy_docs/pg_extensions/pg_failover_slots/configuring.mdx
+++ b/advocacy_docs/pg_extensions/pg_failover_slots/configuring.mdx
@@ -5,9 +5,20 @@ navTitle: Configuring
 
 You must add the extension to `shared_preload_libraries` on both the primary instance as well as any standby that's used for high availability (failover or switchover) purposes.
 
+## Prerequisite settings
+
+The following settings are required: 
+
+- `hot_standby_feedback` should be `on`.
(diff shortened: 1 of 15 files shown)


djw-m (commented):
LGTM

PR #5306, opened 2024-02-26
Update docs on upgrade path, 3.7 to 5 is now supported

sonotley:


djw-m (approved):
LGTM (Huge if true :) )

mwanner (approved):

PR #5317, opened 2024-02-27
Edits for DOCS-138 - BA repo token instructions.

nataliawojcik27:
Need to make it clear that BA free trial customers can also access the token from the EDB Repos


nataliawojcik27:
lgtm

PR #5323, opened 2024-02-28
Corrections to install templates

dwicinas:
Fix to install example for all versions of EPAS so the output shows the correct version number.

Changed config file to show current version of JDBC connector.


dwicinas:
@nidhibhammar This ready for review. A couple points worth noting. All the psql version numbers are x.0.0, such as 16.0.0. That seems reasonable for an example. The actual version numbers might differ.

I hard-coded the redirect for JDBC connector. They way it was defined previously didn't make sense because the redirect was starting at a file that did not exist. I'm not certain they way I've defined it makes sense either because I don't have access to those older versions of JDBC.

nidhibhammar (approved):
LGTM

PR #5324, opened 2024-02-28
Updates list of object types

mlewandowski-edb:
Added "Packages" to the list of database object types that synonyms can be create for.  Support for creating synonyms on packages was added in EPAS 16.

Also, changed "Stored procedures" and "Stored functions" to "Procedures" and "Functions".  The "Stored" qualifier isn't needed in this context since we are already talking about database object types here.


gvasquezvargas (approved):
LGTM

PR #5330, opened 2024-03-01
Adding pre-GA Release Notes for transparency

gvasquezvargas:
## What Changed?

For https://enterprisedb.atlassian.net/browse/DOCS-148, the docs need to specify that the product has been available and in use for a long period of time.
Therefore, we add pre-GA release notes to the public documentation.


djw-m (approved):
LGTM

(Will need to check with @josh-heyer that this will survive any future imports)

PR #5341, opened 2024-03-05
ALL with GROUP COMMIT must use Raft

eatonphil:
One missed page from BDR-4575.


djw-m (approved):
LGTM (well the update looks good, commit scope syntax not so much :) )

PR #5362, opened 2024-03-08
pgd: fixes for bdr.node_summary documentation

ibarwick:
## What Changed?

The field "set_repl_ops" is not present from PGC 3.7 onwards, so remove this from the 3.7 and 4 documentation (it is not in the PGD 5 docs).

BDR-4774.


djw-m (approved):
LGTM

PR #5363, opened 2024-03-08
pgd: fix "bdr.raft_keep_min_entries" reference entry

ibarwick:
## What Changed?

The default value was changed from 100 to 1000 with PGD 5.4 (BDR commit 2588f022) but the reference entry wasn't updated.


djw-m (approved):
LGTM (note to self: need a better format for exceptional/old values)

PR #5364, opened 2024-03-08
pgd: add missing field to "bdr.node_group" reference entry

ibarwick:
## What Changed?

Column `node_group_default_commit_scope`, added in PGD 5.0, was missing, so added. The corresponding entry in `bdr.node_group_summary` is present.


djw-m (approved):
LGTM

PR #5369, opened 2024-03-08
Update 06_demonstration_oracle_compatibility.mdx

dbwagoner:
While testing this code with the password set to "password", I encountered the following error message:

chinook=> CREATE USER demo WITH PASSWORD 'password'; ERROR:  password must contain both letters and nonletters

Adding a number succeeded:

chinook=> CREATE USER demo WITH PASSWORD 'password1'; CREATE ROLE

### DIFF
--- a/product_docs/docs/biganimal/release/using_cluster/06_demonstration_oracle_compatibility.mdx
+++ b/product_docs/docs/biganimal/release/using_cluster/06_demonstration_oracle_compatibility.mdx
@@ -259,7 +259,7 @@ If you want to try these examples on your own BigAnimal cluster, follow these in
 3.  Create a limited-privilege user for connecting to this database:
 
     ```sql
-    CREATE USER demo WITH PASSWORD 'password';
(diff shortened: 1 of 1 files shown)


product_docs/docs/biganimal/release/using_cluster/06_demonstration_oracle_compatibility.mdx
```diff
@@ -259,7 +259,7 @@ If you want to try these examples on your own BigAnimal cluster, follow these in
 3.  Create a limited-privilege user for connecting to this database:
```
gvasquezvargas, line 259: Isn't the `password` supposed to be a placeholder? If yes, we could change to:

```suggestion
3.  Create a limited-privilege user for connecting to this database, replace the `<password>` placeholder with the password you want to assign to the user:
```

djw-m:
Closing this ticket as the Oracle demo is going to be retired.

PR #5377, opened 2024-03-12
Manual proxy installation changes

smaher-edb:
## What Changed?

- Remove the use of `proxy.endpoint` as it is an optional field in config file, and required only when proxy probes are enabled. As config file given in guide doesn't talk about probes user also doesn't need to add `proxy.endpoint`.
- Guide talks about creating a new `pgdproxy` database user but then other sections doesn't use this db user. Made changes to required sections to use this user.

### DIFF
--- a/product_docs/docs/pgd/5/admin-manual/installing/07-configure-proxies.mdx
+++ b/product_docs/docs/pgd/5/admin-manual/installing/07-configure-proxies.mdx
@@ -19,7 +19,7 @@ It is best practice to configure PGD Proxy for clusters to enable this behavior.
 To set up a proxy, you will need to first prepare the cluster and sub-group the proxies will be working with by:
 
 * Logging in and setting the `enable_raft` and `enable_proxy_routing` node group options to `true` for the sub-group. Use [`bdr.alter_node_group_option`](/pgd/latest/reference/nodes-management-interfaces#bdralter_node_group_option), passing the sub-group name, option name and new value as parameters.
-* Create as many uniquely named proxies as you plan to deploy using [`bdr.create_proxy`](/pgd/latest/reference/routing#bdrcreate_proxy) and passing the new proxy name and the sub-group it should be attached to.
+* Create as many uniquely named proxies as you plan to deploy using [`bdr.create_proxy`](/pgd/latest/reference/routing#bdrcreate_proxy) and passing the new proxy name and the sub-group it should be attached to. The `bdr.create_proxy` sets default values for different proxy options like 6432 will be set for proxy listen_port.
 * Create a `pgdproxy` user on the cluster with a password (or other authentication)
 
 ### Configure each host as a proxy
@@ -29,7 +29,7 @@ Once the cluster is ready, you will need to configure each host to run pgd-proxy
 * Creating a `pgdproxy` local user
 * Creating a `.pgpass` file for that user which will allow it to log into the cluster as `pgdproxy`.
 * Modify the systemd service file for pgdproxy to use the pgdproxy user.
-* Create a proxy config file for the host which lists the connection strings for all the nodes in the sub-group, specifies the name that the proxy should use when connected and gives the endpoint connection string the proxy will accept connections on.
+* Create a proxy config file for the host which lists the connection strings for all the nodes in the cluster, specifies the name that the proxy should use when fetching proxy options like listen_address, listen_port etc.
 * Install that file as `/etc/edb/pgd-proxy/pgd-proxy-config.yml`
 * Restart the systemd service and check its status.
 * Log into the proxy and verify its operation.
@@ -90,7 +90,7 @@ __OUTPUT__
 
  bdrdb=#
  ```
- 
+
 ## Create a pgdproxy user on the database
 
 Create a user named pgdproxy and give it a password. In this example we will use `proxysecret`
@@ -100,6 +100,16 @@ On any node, log into the bdrdb database as enterprisedb/postgres.
 ```
 CREATE USER pgdproxy PASSWORD 'proxysecret';
 GRANT bdr_superuser TO pgdproxy;
+```
+
+Set node level route_dsn option for each node with database user pgdproxy. The proxy uses these dsn while connecting to database nodes. Please note the initial dsn (on proxy start) is taken from proxy config file (see below). So, ideally, route_dsn given in this section and in config file should match.
+
+On any node, log into the bdrdb database as enterprisedb/postgres.
+
+```
+SELECT bdr.alter_node_option('host-one', 'route_dsn', 'host=host-one dbname=bdrdb port=5444 user=pgdproxy');
+SELECT bdr.alter_node_option('host-two', 'route_dsn', 'host=host-two dbname=bdrdb port=5444 user=pgdproxy');
+SELECT bdr.alter_node_option('host-three', 'route_dsn', 'host=host-three dbname=bdrdb port=5444 user=pgdproxy');
 ```
 
 ## Create a pgdproxy user on each host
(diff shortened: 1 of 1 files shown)


smaher-edb:
**Note:** I still feel we should remove the pgdproxy user part completely and go with default superuser which comes per-installed with database (including the OS user). It will simplify things for initial user. And then in some other section we could mention about how to use separate user with proxy if required. 

There are other things like in k8s env there won't be any systemd e.g. BA, so, they just need to run `./pgd-proxy -f config.yml`. Also, ideally, service should be run as root because proxy also changes file descriptor limit if have permission etc.

(As a side note, I think the whole guide should have been with 2 data nodes and 1 witness, and using cert based auth instead of md5; at least SCRAM should be used I guess)

May be can be taken care in future updates.

smaher-edb:
I see at other page, in pg_hba, we are using host all all. But if that is changed in future (like per user) then pgdproxy entry will be required there as well. With current things it looks ok.

djw-m:
> There are other things like in k8s env there won't be any systemd e.g. BA, so, they just need to run `./pgd-proxy -f config.yml`. Also, ideally, service should be run as root because proxy also changes file descriptor limit if have permission etc.

K8S will be a separate guide.

product_docs/docs/pgd/5/admin-manual/installing/07-configure-proxies.mdx
```diff
@@ -161,16 +173,14 @@ This is necessary for EDB Postgres Advanced Server instances.
...
 A proxy cannot be on the same port as the Postgres server and, ideally, should be on a commonly used port different from direct connections, even when no Postgres server is running on the host.
 We typically use port 6432 for PGD proxies.
 
 ```
   proxy:
     name: pgd-proxy-one
-    endpoint: "host=localhost dbname=bdrdb port=6432"
 ```
 
 In this case, by using 'localhost' in the endpoint, we specify that this proxy will listen on the host where the proxy is running.
```
djw-m, line 186:
```suggestion
```

Delete this as it refers to the endpoint specification.

product_docs/docs/pgd/5/admin-manual/installing/07-configure-proxies.mdx
```diff
@@ -102,6 +102,16 @@ CREATE USER pgdproxy PASSWORD 'proxysecret';
 GRANT bdr_superuser TO pgdproxy;
 ```
 
+Set node level route_dsn option for each node with database user pgdproxy. The proxy uses these dsn while connecting to database nodes. Please note the initial dsn (on proxy start) is taken from proxy config file (see below). So, ideally, route_dsn given in this section and in config file should match.
```
djw-m, line 105: General point: Do not use relative directions like (see below) in a sequential tutorial or example.

djw-m, line 105: This is a new tangible step and needs to be called out as such or it needs to be folded into the creation of the proxies.

It also needs to be replicated in the steps at the top of this page which serve as the general reference to the process. This is currently only in the worked example.

product_docs/docs/pgd/5/admin-manual/installing/07-configure-proxies.mdx
```diff
@@ -140,17 +150,19 @@ properties:
 
 The name of the PGD cluster's top-level group (as `name`).
 An array of endpoints of databases (as `endpoints`).
-The proxy definition object with a name and endpoint (as `proxy`).
+The proxy definition object with a name (as `proxy`).
+
+The endpoints given in config file are used by proxy only at startup. Once started, the actual list is taken from route_dsn option (see above section).
```
djw-m, line 155: Again: do not use relative pointers like (see above section).

Also:
```suggestion
The endpoints given in the config file are only used by the proxy at startup. Once started, the list is taken from the previously configured `route_dsn` setting for the group.
```

product_docs/docs/pgd/5/admin-manual/installing/07-configure-proxies.mdx
```diff
@@ -19,7 +19,7 @@ It is best practice to configure PGD Proxy for clusters to enable this behavior.
 To set up a proxy, you will need to first prepare the cluster and sub-group the proxies will be working with by:
 
 * Logging in and setting the `enable_raft` and `enable_proxy_routing` node group options to `true` for the sub-group. Use [`bdr.alter_node_group_option`](/pgd/latest/reference/nodes-management-interfaces#bdralter_node_group_option), passing the sub-group name, option name and new value as parameters.
-* Create as many uniquely named proxies as you plan to deploy using [`bdr.create_proxy`](/pgd/latest/reference/routing#bdrcreate_proxy) and passing the new proxy name and the sub-group it should be attached to.
+* Create as many uniquely named proxies as you plan to deploy using [`bdr.create_proxy`](/pgd/latest/reference/routing#bdrcreate_proxy) and passing the new proxy name and the sub-group it should be attached to. The `bdr.create_proxy` sets default values for different proxy options like listen_port which defaults to 6432.
```
djw-m, line 22: Not sure how this addition helps the reader. 

```suggestion
* Create as many uniquely named proxies as you plan to deploy using [`bdr.create_proxy`](/pgd/latest/reference/routing#bdrcreate_proxy) and passing the new proxy name and the sub-group it should be attached to. The `bdr.create_proxy` does not create a proxy, but creates a space for a proxy to register itself with the cluster. The space contains configuration values which can be modified later. Initially it is configured with default proxy options such as setting the listen_port to 6432.
```

djw-m (changes requested):

product_docs/docs/pgd/5/admin-manual/installing/07-configure-proxies.mdx
Jagdish-Kewat-EDB, line 22: The default value of 6432 will not be true after release of read scalability so should we add it?

product_docs/docs/pgd/5/admin-manual/installing/07-configure-proxies.mdx
smaher-edb: I'm not sure I get it but removing `(see below)`. Quick search suggest 12 places where `(see below)` is mentioned in docs repo.

djw-m: Yes, and our style guide says don't do it, and there's a lot of pre-style guide content in the PGD docs which is yet to be tidied up.

product_docs/docs/pgd/5/admin-manual/installing/07-configure-proxies.mdx
smaher-edb: > but creates a space for a proxy to register itself with the cluster

Not sure I get this. It is same as create node or create group (which creates nothing but group or node configs). It creates default configuration. But may be you meant same. Committing it.

smaher-edb: This is v5.4 guide. There will lots of changes I guess after read scalability. We can modify at that time or just mention something else like listen_address (0.0.0.0)

djw-m: What will the new default value be? And why is an existing default changing with the addition of read scalability?

smaher-edb: Changing it to`listen_address to `0.0.0.0``

smaher-edb: `listen_port` could be nullable after read scalability feature. listen_port will decide whether proxy is in default, read-only or any mode. We had added separate proxy mode to indicate that but after RFC discussion it was suggested to use this option for deciding proxy which mode proxy is configured in.

product_docs/docs/pgd/5/admin-manual/installing/07-configure-proxies.mdx
smaher-edb: Yes, but added it in same section, so, it is clear it is required only if someone is creating separate user for proxy i.e. step mentioned in section `## Create a pgdproxy user on the database`. So, kept them together. But yes please change it as you find suitable. My only intention was to add the step which is required if someone decides to go with separate user (i.e. altering route_dsn part).

djw-m:
Closing this PR - superseded by #5448

PR #5399, opened 2024-03-18
Clarify that timestamp args of PWR accept input with or without time zone

barthisrael:
When no time zone is explicitly set, PWR falls back to the local system time zone.

References: PGCP-71.

## What Changed?

Added a note to the PWR docs explaining that in the absence of a time zone in the timestamp arguments, PWR falls back to the system time zone.


gcalacoci (approved):

djw-m (commented):
LGTM

PR #5406, opened 2024-03-19
Clarify in the docs that PWR configuration file should be in YAML format

barthisrael:
Although the PWR configuration file is named with a `.conf` suffix, its content must be in YAML format.

This commit adds a note with that information, so we avoid having users to face config parsing issues.

References: PGCP-63.

## What Changed?

Adds a note that PWR config file content must be in YAML format.


gcalacoci (approved):

djw-m (approved):
LGTN

PR #5418, opened 2024-03-20
Add missing nav sections so pdf can include them

djw-m:
## What Changed?

Added the two missing sections from the navigation to restore them to the PDF.


josh-heyer (approved):
Turns out they weren't missing, they were in the wrong spot, sections 5 and 6. In this case, the length of the last section made them very difficult to see. Putting them in the navigation is a correct fix and less confusing for the next person; fixing the script to behave consistently with the site should also reduce confusion in the future.

PR #5419, opened 2024-03-20
Update dha_bulk_migration.mdx

aswright491:
remove 'DHA' acronym


djw-m (approved):
LGTM - but does this mean we say Distributed High Availability but only refer to PGD for short?

PR #5429, opened 2024-03-21
update featured topics

josh-heyer:
## What Changed?

Feature latest releases for PGD & TPA.

Also bring in @ebgitelman's fixes for TPA's Tower topic, since linking to it in its current state is not great


jpe442 (approved):
LGTM! 🚢

PR #5437, opened 2024-03-25
Add more information about efm startup log

EFM-Bobby:
This changes adds notes in two places to tell the user to check the startup log if there is a problem starting efm. Too often the user only checks 'systemctl status' output, which does not contain the problem.

(I'd like to get rid of the startup log because users forget it's there and it's not needed any more. I don't know when it'll happen though: https://enterprisedb.atlassian.net/browse/EFM-1619)

### DIFF
--- a/product_docs/docs/efm/4/05_using_efm.mdx
+++ b/product_docs/docs/efm/4/05_using_efm.mdx
@@ -39,6 +39,8 @@ To start the Failover Manager cluster on RHEL/CentOS 7.x or RHEL/Rocky Linux/Alm
 
 `systemctl start edb-efm-4.<x>`
 
+!!! Note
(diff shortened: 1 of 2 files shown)


product_docs/docs/efm/4/05_using_efm.mdx
```diff
@@ -39,6 +39,8 @@ To start the Failover Manager cluster on RHEL/CentOS 7.x or RHEL/Rocky Linux/Alm
 
 `systemctl start edb-efm-4.<x>`
 
+!!! Note
+    If the agent fails to start, see the startup log `/var/log/efm-4.<x>/startup-efm.log` for more information.
```
EFM-Bobby, line 43: I used the 4.<x> format here to be consistent with the page, but a more general efm-<version> format on the troubleshooting page.

djw-m (approved):
LGTM

PR #5442, opened 2024-03-25
Sync PG4K plugin docs with upstream

josh-heyer:
## What Changed?

This info hasn't been synchronized with the cloudnativepg version in like 8 months.  Some information is missing; some is out of date; some is straight-up wrong.

This PR does a quick update to get correct information out. Will need to also make some corrections in the internal repo (and figure out a better sync process) as well as updates upstream.

### DIFF
--- a/product_docs/docs/postgres_for_kubernetes/1/kubectl-plugin.mdx
+++ b/product_docs/docs/postgres_for_kubernetes/1/kubectl-plugin.mdx
@@ -34,52 +34,67 @@ them in your systems.
 
 #### Debian packages
 
-For example, let's install the 1.18.1 release of the plugin, for an Intel based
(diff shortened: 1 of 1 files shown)


djw-m:
This only touches on one page and not the one that was reported:

https://deploy-preview-5442--edb-docs-staging.netlify.app/docs/postgres_for_kubernetes/latest/installation_upgrade/

This is missing the --server-side flag which is present on https://cloudnative-pg.io/documentation/1.22/installation_upgrade/

josh-heyer:
Oh wow, thanks for the catch @djw-m - I'm just gonna blame that on lack of coffee. This is even weirder though: the changes are present upstream in main, but [not in 1.22.2](https://github.com/EnterpriseDB/cloud-native-postgres/blob/v1.22.2/docs/src/installation_upgrade.md)!

PR #5453, opened 2024-03-28
Timestamp check fails with example

martinmarques:
The example fails with the timestamp check library we use. Added missing seconds to timestamp and minutes to timezone


gcalacoci (approved):
WFM

PR #5458, opened 2024-03-29
Revise sqlpatch docs to sync with customer support content.

DonaldKai:
Making a change to the PGD cluster section to match that off the KB article:

https://techsupport.enterprisedb.com/kb/a/august-2023-advanced-server-security-update/

Customer in 103516 also was confused when they applied the edb_sqlpatch to the leader node and not all patches were applied to other nodes in the PGD cluster, like the documentation says.


djw-m (approved):
LGTM

PR #5496, opened 2024-04-11
update jdbc.properties note

EFM-Bobby:
Instead of adding this property to the efm properties file template, we're updating the note about how to use it instead. This is the only case that the property is used without working with EDB first for instructions on what to use for the property information.


EFM-Bobby:
This change is for https://enterprisedb.atlassian.net/browse/EFM-1662

josh-heyer (approved):

PR #5497, opened 2024-04-11
Correct and update generated installer files

josh-heyer:
## What Changed?

Clean up generation & deployment for installer templates:
    
- update versions for JDBC connector and PostGIS
- use major version for PostGIS deployment (match new directory versioning) - also add static redirect for this
- use fake but realistic major version for EPAS psql examples (vs. using 16.whatever for ALL versions)
- update dependencies, add product.versionMajor context var

### DIFF
--- a/advocacy_docs/supported-open-source/postgresql/installing/index.mdx
+++ b/advocacy_docs/supported-open-source/postgresql/installing/index.mdx
@@ -9,8 +9,6 @@ indexCards: none
 
 redirects:
   - /supported-open-source/postgresql/installer
-  - /postgresql_journey/02_installing/03_linux/
-  - /postgresql_journey/02_installing/
 
 navigation:
   - requirements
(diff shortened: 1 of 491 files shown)


dwicinas (commented):
@josh-heyer, In general this looks good, but I have a question about the new redirects. For those files where you have created redirects, why are you redirecting from "latest" to the current location. Did the files you are referencing ever exist in "latest"? The way I understand it, those files just existed in previous versions of the product. Or, does "latest" somehow work as a wild card and it refers to all previous versions of the current product, not just the current product? Hopefully I'm expressing myself clearly. After getting this issue resolved, I would approve this PR.

josh-heyer:
"latest" is special *for redirects only* - the redirects engine will map it to whatever version the file belongs to, e.g. a "latest" redirect in epas/14/index.mdx will result in a redirect from epas/14/whatever, while the same redirect in epas/16/index.mdx will become epas/16/whatever. This has the same effect as injecting the version in our templates, so I could have done this either way - I originally switched from injecting the version because of the disconnect between installer version and path version for JDBC, but then added the `versionMajor` variable to support that in deployment paths.

dwicinas:
@josh-heyer in light of your comment about "latest" I approve this PR. Should I request another review or can you handle it all on your end?

josh-heyer:
I got it - will rebase and merge

PR #5500, opened 2024-04-15
Add information about slot privileges needed

EFM-Bobby:
If using physical replication slots, the database user efm uses must have permissions for creating and advancing slots. This commit adds a note about it and the command needed to grant the privileges.


djw-m (approved):
LGTM

PR #5501, opened 2024-04-16
pgd: update bdr_read_all_stats / bdr_monitor documentation

ibarwick:
## What Changed?

The lists of objects for which each of these predefined roles has privileges on, are somewhat outdated.

Update the lists based on a query of the actual database privileges installed for each role.

BDR-4899 (incorporating the additional permissions added via BDR-4842 and BDR-4885).


djw-m (approved):
LGTM

PR #5510, opened 2024-04-17
Add table of metric names for 3PM integrations

sonotley:
This is a quick edit to get this information published for a customer. We should consider embedding these names in the list of metrics rather than a separate table as a future improvement.


nidhibhammar (approved):
LGTM

PR #5529, opened 2024-04-23
Docs/poc/newinfobox - NOT FOR DEPLOYMENT

djw-m:
## What Changed?

Experimental new info box for pages. Works on PGD, milage elsewhere may vary depending on where release notes are.


josh-heyer (approved):
This looks great! I was able to simplify the code a bit, and fix that issue where navigating back to the top level didn't restore scroll position - that's probably been a bug forever, so nice to have it solved finally!

PR #5553, opened 2024-04-29
Outdated link in TPA harp documentation.

gvasquezvargas:
Target link does not exist. DJ suggested this could be the correct link https://www.enterprisedb.com/docs/pgd/4/harp/. However, I am wondering if we should be linking to older docs at all? Maybe the recommended path now is to have PGD instead of installing BDR or HARP components?


gvasquezvargas:
Closing in favour of upstream: https://github.com/EnterpriseDB/tpa-internal/pull/203

PR #5554, opened 2024-04-29
Fix release years that were changed for 4.0/4.1

EFM-Bobby:
Back in commit 0fcdb9f1c37e24e8ca7e2e80c228e368dd5c7237, the release years for efm versions 4.0 and 4.1 got bumped up by a year (so that 4.0 and 4.1 were actually released AFTER 4.2). Fixing this based on the diff of the commit above and the dates of the changes in efm code itself.


djw-m (approved):
LGTM

PR #5563, opened 2024-04-30
Remove minimum db version info in 4.8 release notes

EFM-Bobby:
We don't put supported database version information into the release notes -- users should check the platform compatibility page instead.

Adding it here was my mistake last time around, and the information is wrong now that v11 has gone out of support.


anilkumar29 (approved):

EFM-Bobby:
@nidhibhammar Hi, can you ok this? Anil is the PM for efm and ok'ed the change, but I need a code owner to review before I can merge (and github suggested you).

djw-m (approved):
LGTM

PR #5575, opened 2024-05-03
Use github-slugger for section header links

josh-heyer:
## What Changed?

Switch to [GithubSlugger](https://github.com/Flet/github-slugger) for generating card section links. This is the same library used for the rest of the ToC and should therefore avoid most issues with punctuation, etc.

One thing to note: I've made no effort to avoid collisions here between section headers and other headers already used in the same topic. While normally this is handled by incrementing a suffix, that won't happen here - so if you use a header in markdown and the same header text for a section title, the section title link will take you to the section in the markdown. This is rare, and easy enough to avoid by just using unique section titles. Note that using the same title for multiple navigation sections *will* generate a unique suffix and work just fine, but is still a terrible idea and you shouldn't do it.

### DIFF
--- a/src/components/card-decks.js
+++ b/src/components/card-decks.js
@@ -72,7 +72,7 @@ const CardDecks = ({ cards, cardType = "simple", deckTitle = "" }) => {
   return (
     <>
       {deckTitle && (
-        <h2 className="mt-3" id={slugger.slug(deckTitle)}>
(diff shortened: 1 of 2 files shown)


djw-m (approved):
Have made a small tweak (to precede section links with "section-") to reduce liklihood of TOC and Section anchor collisions. 

Otherwise LGTM

PR #5576, opened 2024-05-03
Admonition cleanup

josh-heyer:
## What Changed?

Fix a few admonitions that were improperly indented and/or fenced.
Most of these (all but one?) were still rendering fine as they were implicitly closed by another admonition, list item, etc. But that's a hazard waiting to trip up someone modifying nearby text.


djw-m (approved):
LGTM

PR #5584, opened 2024-05-08
Full nav cards in versioned docs + bugfixes

josh-heyer:
## What Changed?

- Section headings now render for card navigation section in unversioned docs
- Full card navigation no longer renders icons, when there are no icons defined for child pages
- Navigation order no longer ignored for grandchildren in "full" card navigation cards
- Sidebar navigation now expands normally even when page title is "NULL".
- As discussed on IA call, there's a new "Up" button that takes you to the parent, and the Prev button now behaves in a consistent fashion everywhere: to the last descendant of the previous sibling (IOW, the page for which the "next" button will take you to the current page). Existing behavior took you to the last child of the previous sibling, unless filename of previous sibling sorts alphabetically before the filename of current node in which case prev button takes you to previous sibling.

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -17,7 +17,6 @@ const {
   buildProductVersions,
   reportMissingIndex,
   treeToNavigation,
-  treeNodeToNavNode,
   findPrevNextNavNodes,
   preprocessPathsAndRedirects,
   configureRedirects,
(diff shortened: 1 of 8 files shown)


djw-m (approved):
LGTM - got it up and running locally and everything seems as advertised

Tho it might be nice to move the Prev/Up/Next higher in the page as part of an updated nav element, this will let people traverse more easily.

PR #5590, opened 2024-05-08
Add warning recommending strict adherence to EPRS install steps

dwicinas:
Added a global warning for install and another for upgrade.

Addresses https://enterprisedb.atlassian.net/browse/DOCS-630.


dwicinas:
@josh-heyer I added you in case you want to review the template change. The way I handled it is not very elegant so you might have some other idea.

anwar-sahib (approved):
Changes look fine.

josh-heyer (approved):

josh-heyer:
It's good. For a one-off bit of verbiage, simple is better - makes changes easier in the future.

Issue #5486, opened 2024-04-09
Regarding "Using EDB Advanced Storage Pack"


tusharahuja:
### Summary

The Autocluster example mentioned in our DOC  doesn't seem to work(https://www.enterprisedb.com/docs/pg_extensions/advanced_storage_pack/using/#autocluster-example)

edb=# INSERT INTO iot (thermostat_id, recordtime) VALUES (456, 12:01);
ERROR:  syntax error at or near ":"
LINE 1: ...RT INTO iot (thermostat_id, recordtime) VALUES (456, 12:01);
                                                                  ^
edb=# INSERT INTO iot (thermostat_id, recordtime) VALUES (8945, 04:55);

ERROR:  syntax error at or near ":"
LINE 1: ...T INTO iot (thermostat_id, recordtime) VALUES (8945, 04:55);

### Where did you see the problem?

https://github.com/EnterpriseDB/docs/blob/main/advocacy_docs/pg_extensions/advanced_storage_pack/using.mdx

### Expected behavior

insert command should work

PR #5601, opened 2024-05-10
fix autocluster example

tureba:
The example had timestamptz as datatype and inserted only hour:minute unquoted.

Besides the obvious quoting, we should either change the data to match the datatype, or the datatype to match the hour:minute data. Using full timestamptz probably wouldn't improve the example, which is well served by simple data. So using `time` as column datatype is adequate.

This intends to cover issue #5486 .


nidhibhammar (approved):
LGTM

PR #5602, opened 2024-05-13
Test attempt to combine PEM prereqs topic with install template

dwicinas:


install_template/templates/products/postgres-enterprise-manager-server/debian-10.njk
```diff
@@ -1,7 +1,17 @@
 {% extends "products/postgres-enterprise-manager-server/base.njk" %}
 {% set platformBaseTemplate = "debian-10" %}
+{% set upgradeCommand = "update" %}
 {% block repocheck %}
 To determine if your repository exists, enter this command: 
   
 `apt-cache search enterprisedb`
 {% endblock repocheck %}
+{% block debianUbuntuNote %}!!! Note
```
sonotley, line 9: I don't think we need this note on any page other than Debian 10, and that's about to go EOL so I think we can just drop it.

sonotley (approved):

install_template/templates/products/postgres-enterprise-manager-server/debian-10.njk
nidhibhammar: We are removing Debian 10 from a separate PR, so we are keeping this for now.

PR #5628, opened 2024-05-16
Update registering_agent.mdx

sonotley:
Dash should be underscore (yes that parameter doesn't match the rest...)


nidhibhammar (approved):
LGTM

PR #5658, opened 2024-05-21
Updated with new error

djw-m:
## What Changed?

Added Daylight Savings Time error (prefixed with Incorrect Timestamps which would be the user observed error, rather than the cause.


djw-m:
Approved in https://enterprisedb.atlassian.net/browse/DOCS-553?focusedCommentId=593315

nidhibhammar (approved):
LGTM

PR #5684, opened 2024-05-23
Edits to Migration Toolkit PR5658

ebgitelman:

### DIFF
--- a/product_docs/docs/migration_toolkit/55/09_mtk_errors.mdx
+++ b/product_docs/docs/migration_toolkit/55/09_mtk_errors.mdx
@@ -259,19 +259,18 @@ To correct the problem, specify `-fetchSize 1` as a command-line argument when y
 
 ### Incorrect timestamps and Daylight Saving Time
 
-When migrating from SQL Server to PostgreSQL using the MSSQL JDBC driver there may be no errors observed. If the source database contains rows with timestamp values which are within a range of time when Daylight Savings Time was in effect, they will be migrated to the target database with the wrong timestamps.
+When migrating from SQL Server to PostgreSQL using the MSSQL JDBC driver, the migration may error without reporting any errors. However, if the source database contains rows with timestamp values that are within a range of time when Daylight Savings Time was in effect, those rows will be migrated to the target database with the wrong timestamps.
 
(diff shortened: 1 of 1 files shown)


ebgitelman:
@djw-m I took a little heavier hand with this than usual because it was hard to follow some of the sentence logic. Let me know if this is correct.

product_docs/docs/migration_toolkit/55/09_mtk_errors.mdx
```diff
@@ -259,19 +259,18 @@ To correct the problem, specify `-fetchSize 1` as a command-line argument when y
 
 ### Incorrect timestamps and Daylight Saving Time
 
-When migrating from SQL Server to PostgreSQL using the MSSQL JDBC driver there may be no errors observed. If the source database contains rows with timestamp values which are within a range of time when Daylight Savings Time was in effect, they will be migrated to the target database with the wrong timestamps.
+When migrating from SQL Server to PostgreSQL using the MSSQL JDBC driver, the migration may error without reporting any errors. However, if the source database contains rows with timestamp values that are within a range of time when Daylight Savings Time was in effect, those rows will be migrated to the target database with the wrong timestamps.
```
gvasquezvargas, line 262: This was a bit confusing to me in both versions. How about:
```suggestion
When migrating from SQL Server to PostgreSQL using the MSSQL JDBC driver, the migration may not report any errors in spite of a timestamp migration failure. If the source database contains rows with timestamp values that are within a range of time when Daylight Savings Time was in effect, those rows will be migrated to the target database with the wrong timestamps.
```

djw-m:
```suggestion
When migrating from SQL Server to PostgreSQL using the MSSQL JDBC driver, there can be an error in the migration that is not reported. Specifically, if the source database contains rows with timestamp values that are within a range of time when Daylight Savings Time was in effect, those rows will be migrated to the target database with the wrong timestamps.
```

gvasquezvargas (approved):
LGTM after updating the timestamp wording.

djw-m (approved):
LGTM

PR #5687, opened 2024-05-24
Update sequences.mdx

lucianobotti:
Suggesting a fix for the query ( needs to use jsonb instead of json and is missing an ending ')'

## What Changed?

The following query example is not working:

-- determine highest sequence value across all nodes
SELECT max((x->'response'->'command_tuples'->0->>'nextval')::bigint)
    FROM json_array_elements(
        bdr.run_on_all_nodes(
            E'SELECT nextval(\'public.sequence\');'
            )::jsonb AS x;
            
Missing an ending ')' and then to use 'jsonb' function


djw-m (approved):
LGTM - code works and doesn't error.

PR #5691, opened 2024-05-27
Added trademark in titles

gvasquezvargas:
## What Changed?

According to https://www.enterprisedb.com/trademarks, the trademark sign ® must go in: 
- first occurrence of trademark in titles
- first occurrence of trademark in the text 

I fixed the first occurrence in text in https://github.com/EnterpriseDB/docs/pull/5680 for the new EDB Postgres AI section, and now fixing titles for the same section of our documentation. 

I need a quick sanity check that this is the direction we want to take.


gvasquezvargas:
Had a sync with @djw-m  on this and we won't add the symbol to titles as it will make them unsearchable. Hence, closing this one.

PR #5694, opened 2024-05-28
Fix text to match recent edits

djw-m:
## What Changed?

Fixes output text (embedded in code as so little text, a template seemed overkill)


gvasquezvargas (approved):
LGTM!

PR #5725, opened 2024-06-03
DOCS-690 First draft changes for mixed versions

djw-m:
## What Changed?

PGD 4 upgrades
PGD 5 upgrades and limitations

### DIFF
--- a/product_docs/docs/pgd/4/upgrades/index.mdx
+++ b/product_docs/docs/pgd/4/upgrades/index.mdx
@@ -56,16 +56,18 @@ A rolling upgrade starts with a cluster with all nodes at a prior release,
 then proceeds by upgrading one node at a time to the newer release, until
 all nodes are at the newer release. There should never be more than two versions
 of any component running at the same time, which means the new upgrade must not
-be initiated until the previous upgrade process has fully finished on all nodes.
+be initiated until the previous upgrade process has fully finished on all nodes. 
 
-An upgrade process can take more time when caution is required to reduce
-business risk. Do not run with mixed versions of the software for any longer
-than is necessary to complete the upgrade. The longer you run with mixed
-versions, the more likely you are to encounter issues, the more difficult it is
-to diagnose and resolve them, and you will not be able to use new features until
-all nodes are upgraded. We recommend upgrading in off peak hours for your
-business, and over a short period of time. For maximum caution, we suggest
-performing the upgrades in separate a test environment first.  
+Where additional caution is required to reduce business risk, more time may be
+required to perform an upgrade. For maximum caution and to reduce the time
+required upgrading production systems, we suggest performing the upgrades in a
+separate test environment first. 
+
+Don't run with mixed versions of the software for any longer than is absolutely
+necessary to complete the upgrade. The longer you run with mixed versions, the
+more likely you are to encounter issues, the more difficult it is to diagnose
+and resolve them.  We recommend upgrading in off peak hours for your business,
+and over a short period of time. 
 
 While Rolling Upgrade can be used for upgrading major version of the software
 it is not supported to mix PostgreSQL, EDB Postgres Extended and
--- a/product_docs/docs/pgd/4/upgrades/index.mdx
+++ b/product_docs/docs/pgd/4/upgrades/index.mdx
@@ -58,9 +58,14 @@ all nodes are at the newer release. There should never be more than two versions
 of any component running at the same time, which means the new upgrade must not
 be initiated until the previous upgrade process has fully finished on all nodes.
 
-An upgrade process may take an extended period of time when the user decides
-caution is required to reduce business risk, though it's not recommended
-to run the mixed versions of the software indefinitely.
+An upgrade process can take more time when caution is required to reduce
+business risk. Do not run with mixed versions of the software for any longer
+than is necessary to complete the upgrade. The longer you run with mixed
+versions, the more likely you are to encounter issues, the more difficult it is
+to diagnose and resolve them, and you will not be able to use new features until
+all nodes are upgraded. We recommend upgrading in off peak hours for your
+business, and over a short period of time. For maximum caution, we suggest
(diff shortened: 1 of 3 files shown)


product_docs/docs/pgd/5/planning/limitations.mdx
```diff
@@ -78,9 +78,14 @@ and current listing.
...
-We expect users to run mixed versions only during upgrades and, once an upgrade starts, that they complete that upgrade.
-We don't support running mixed versions of PGD except during an upgrade.
+PGD was developed to [enable rolling upgrades of PGD](/pgd/latest/upgrades) by
+allowing mixed versions of PGD to operate during the upgrade process. We expect
+users to run mixed versions only during upgrades. Once an PGD cluster upgrade
+has begun, you should complete the whole cluster upgrade as quickly as possible.
+
+We don't support running mixed versions of PGD except during an upgrade, and we
+don't support clusters running mixed versions, while being upgraded, for a
+extended periods such as over a month.
```
mwanner, line 88: While this is sufficiently vague, I think I'd still prefer to not mention any time frame at this point in time (Ribbon may interpret this as a confirmation of their 30 day figure, which I'd like to avoid).

mwanner (commented):
Generally sounds very good to me. However, the dev side still needs to contribute by clarifying what exactly does or doesn't work. Or why running a mixed version cluster poses risks. And what risks.

product_docs/docs/pgd/5/planning/limitations.mdx
djw-m:
```suggestion
extended periods.
```

product_docs/docs/pgd/5/planning/limitations.mdx
```diff
@@ -78,9 +78,14 @@ and current listing.
 
 ## Mixed PGD versions
 
-PGD was developed to [enable rolling upgrades of PGD](/pgd/latest/upgrades) by allowing mixed versions of PGD to operate during the upgrade process. 
-We expect users to run mixed versions only during upgrades and, once an upgrade starts, that they complete that upgrade.
-We don't support running mixed versions of PGD except during an upgrade.
+PGD was developed to [enable rolling upgrades of PGD](/pgd/latest/upgrades) by
```
nataliawojcik27, line 81: @djw-m need to adjust this to explain more

One of PGD's key value propositions is in its ability to enable rolling upgrades, allowing users to avoid downtime associated with moving to new versions. We recommend testing rolling upgrades in a test environment before going to production. When performing a rolling upgrade, users add a new node to an existing cluster, which results in a mixed version cluster. We recommend performing rolling upgrades during off-peak hours and completing the process in one go in order to minimize the time the cluster is in a mixed version state. You can't use the new features in the new version until the rolling upgrade is complete, and the node with the old version has been removed. PGD does not support mixed version clusters for production. As with any mixed version software, there are common pitfalls that might arise if the upgrade is not completed including but not limited to: incompatibility issues between the different versions causing errors and inefficiencies and increased complexity during troubleshooting due to version discrepancies. 

While performing a rolling upgrade, you cannot change database types i.e. from Postgres to EDB Postgres Extended.

product_docs/docs/pgd/4/upgrades/index.mdx
```diff
@@ -58,9 +58,14 @@ all nodes are at the newer release. There should never be more than two versions
 of any component running at the same time, which means the new upgrade must not
 be initiated until the previous upgrade process has fully finished on all nodes.
 
-An upgrade process may take an extended period of time when the user decides
-caution is required to reduce business risk, though it's not recommended
-to run the mixed versions of the software indefinitely.
+An upgrade process can take more time when caution is required to reduce
```
nataliawojcik27, line 61: @djw-m would update this to match the version 5 language

product_docs/docs/pgd/5/upgrades/manual_overview.mdx
```diff
@@ -59,14 +59,19 @@ nodes are at the newer release. There must be no more than two versions of the
...
 
-An upgrade process can take more time when 
-caution is required to reduce business risk. However, we don't recommend
-running mixed versions of the software indefinitely.
-
-While you can use a rolling upgrade for upgrading a major version of the software,
-we don't support mixing PostgreSQL, EDB Postgres Extended, and
-EDB Postgres Advanced Server in one cluster. So you can't use this approach 
-to change the Postgres variant.
+An upgrade process can take more time when caution is required to reduce
```
nataliawojcik27, line 62: @djw-m update to match other language

nataliawojcik27 (changes requested):

product_docs/docs/pgd/5/planning/limitations.mdx
djw-m: This is the Limitations section where we are historically brief on the issue so that the explicit limitation is directly visible. If anything, this section has expanded to the point of being too large as we don't have an explicit limitation on the use on mixed versions.

product_docs/docs/pgd/5/upgrades/manual_overview.mdx
djw-m: The other language is in a limitations section. This is a difference section and is already explicit in its recommendations.

product_docs/docs/pgd/4/upgrades/index.mdx
djw-m: It does match the version 5 language *for the upgrades section*.

product_docs/docs/pgd/5/upgrades/manual_overview.mdx
nataliawojcik27: @djw-m it is explicit, but it's written in a way that makes it hard for the customer to understand the true impact. Recommend we explain more like suggested here https://github.com/EnterpriseDB/docs/pull/5725#discussion_r1626341673

PR #5730, opened 2024-06-04
Add support for Debian 12 to install documentation

dwicinas:


product_docs/docs/epas/11/installing/index.mdx
```diff
@@ -15,6 +15,7 @@ redirects:
 navigation:
   - linux_x86_64
   - linux_ppc64le
+  - linux_arm64
```
nidhibhammar, line 18: This is not needed in version 11

product_docs/docs/epas/11/installing/linux_x86_64/index.mdx
```diff
@@ -22,6 +22,7 @@ navigation:
   - epas_sles_12
   - epas_ubuntu_22
   - epas_ubuntu_20
+  - epas_debian_12
```
nidhibhammar, line 25: This is not needed in version 11.

nidhibhammar:
@dwicinas, We need to remove the entries of Debian 12 from the navigation in index files of earlier versions of EPAS, OCL connector, ODBC connector, PEM, and PGE.

dwicinas:
@nidhibhammar The template creates those navigation entries in the index pages. I would need Josh's help to rewrite the template so it doesn't automatically add those kind of entries to all index pages. I could manually delete the entries from index pages, but every time I generate new topics the template would want to add them back in. If there's no underlying topics related to those navigation entries, the entries don't create anything that's visible to the user, so I decided to leave them there. Of course, I can manually remove them and enlist Josh's help if that's what you want.

nidhibhammar:
@josh-heyer, we are adding Debian 12 install topics for all the products. It is supported only for the latest versions. We need to remove the Debian 12 entries from the navigation of index files of earlier versions of EPAS, OCL connector, ODBC connector, PEM, and PGE. For reference, please take a look at my above two comments. Can you please help us get the correct templates so index entries will only be for the latest version?

josh-heyer:
Ok, @nidhibhammar @dwicinas - I've modified these templates such that nav sections are generated based on config data. This ensures there are no entries that are not needed... It ALSO enforces a particular grouping and ordering. This should make it easier to work with in the future. It is still possible to customize the navigation for products that need it.

I've rebased this branch, and rewritten past commits to streamline the work here a bit. All generated files are now contained within [d568e8e](https://github.com/EnterpriseDB/docs/pull/5730/commits/d568e8ea10b5536e1b4d272c221cf05543ed2cf2) (the latest as I write this). You don't *have* to maintain that approach, but I do recommend keeping template changes and generated files in separate commits for ease of review. 

Make sure to `git pull --rebase` locally before continuing work.

nidhibhammar:
Thanks, @josh-heyer, for all your support on this.

nidhibhammar (approved):
LGTM

PR #5765, opened 2024-06-12
Remove confusing text about minor releases and patches

EFM-Bobby:
EFM doesn't have "minor" releases or patches. Every release is separate from the others, even if it's just the "minor" number that changed in the version.


djw-m (approved):
LGTM

PR #5787, opened 2024-06-19
Ensure path prefix is used for SVG links

josh-heyer:
## What Changed?

Root cause was lack of automatic prefixing; fixing this was enough work that it made sense to just abandon the use of the `/static` tree for SVGs in favor of an approach that allows them to be used more or less the same as any other image (by rewriting relative paths based on the referencing page, and then adding the path prefix).

This can (and should) be expanded for other file types that aren't natively supported but which we might want to reference in this manner.


gvasquezvargas (approved):
Looking good on preview, we'll have to push to prod to be sure. 
Thanks for the quick fix!

djw-m (approved):
LGTM - the paths should work on all deployments now

PR #5810, opened 2024-06-26
Update 01_portal_access.mdx

liudanbj2021:
the note is not correct in https://www.enterprisedb.com/docs/biganimal/latest/administering_cluster/01_portal_access/#organization-level-roles

## What Changed?

update the note format. please verify if it is correct.


djw-m (approved):
LGTM

PR #5830, opened 2024-07-02
Fix render loop on advanced search

josh-heyer:
## What Changed?

Root cause of this was passing a string to `<Configure>` where it expected a number. The irritating bit about that was that it *worked*, it just kept re-rendering as the internal state was never an exact match with what was passed. 

Took this opportunity to clean up a few other places that were triggering unnecessary re-rendering; page should be noticeably faster now.

Finally, re-enable analytics (which I'd disabled the other dayas a temporary mitigation; didn't fix the root issue, but did stop it from quickly crashing a browser by filling logs with errors)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -81,157 +81,141 @@
       }
     },
     "node_modules/@algolia/cache-browser-local-storage": {
-      "version": "4.23.3",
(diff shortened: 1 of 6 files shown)


djw-m (approved):
LGTM (but the proof is in the pudding)

djw-m:
Checked out on preview deploy, moving into production.

PR #5859, opened 2024-07-17
Suggestions for Getting Started "Planning" and "Overview" sections

gvasquezvargas:
## What Changed?

When looking at the rebranding branch, I noticed some overlap between the **Getting Started > Overview of Cloud Service** and the **Getting Started > Planning** sections. 

In this PR, I propose unifying both sections to ensure that information is not in two places but in a single place, namely, the planning section. 

I have also added some minor suggestions for bits I noticed along the way.  

Let me know what you think. 

PS: I mistakenly named the branch `epas`, feel free to ignore the mistake. The contents of this PR refer to the rebranding and reorganization of BA docs.


nidhibhammar (approved):
LGTM

PR #5872, opened 2024-07-23
Document deep spl_check enhancement

dwicinas:
Added verbiage to Using section of SPL Check

Addresses https://enterprisedb.atlassian.net/browse/DOCS-763.

### DIFF
--- a/advocacy_docs/pg_extensions/spl_check/using/index.mdx
+++ b/advocacy_docs/pg_extensions/spl_check/using/index.mdx
@@ -188,7 +188,7 @@ ORDER BY (pcf).functionid::regprocedure::text, (pcf).lineno;
 ```
 ## Validating function calls
 
-When the GUC `spl_check.validate_function_calls` is set to true, `spl_check` validates a function and any function bodies encountered while validating the initial function.
+When the GUC `spl_check.validate_function_calls` is set to true, `spl_check` validates a function and any function calls encountered while validating the initial function.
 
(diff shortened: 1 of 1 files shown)


advocacy_docs/pg_extensions/spl_check/using/index.mdx
```diff
@@ -186,6 +186,15 @@ OFFSET 0
...
 ```
+## Validating function calls
+
+When the GUC `spl_check.validate_function_calls` is set to true, `spl_check` validates a function and any function bodies encountered while validating the initial function.
+
+For example, if function `f1()` calls function `f2()` and `spl_check.validate_function_calls` is set to true, `spl_check` vaidates function `f1()` and function `f2()`. If `spl_check.validate_function_calls` is set to false, `spl_check` validates only function `f1()` and checks the parameters and return type of `f2()`.
+
+Use the `SET` command to set the value of `spl_check.validate_function_calls`. By default `spl_check.validate_function_calls` is set to false.
+
+`spl_check` does not check any functions with an OID less than `FirstNormalObjectId` (that is, functions defined in the `sys` or `pg_catalog` schema). They are assumed to be system functions. Also, `spl_check` skips system packages and functions written in languages other than EDB SPL.
```
ebgitelman, line 197:
```suggestion
`spl_check` doesn't check any functions with an OID less than `FirstNormalObjectId`, that is, functions defined in the `sys` or `pg_catalog` schema. They are assumed to be system functions. Also, `spl_check` skips system packages and functions written in languages other than EDB SPL.
```

ebgitelman (commented):
Just a couple of small comments. Didn't review installs since they are generated.

nidhibhammar (approved):
LGTM

josh-heyer:
⚠Rebased - history has changed, please reset your local branch before continuing work.

PR #5878, opened 2024-07-26
DOCS-911 - Added warning to reference entry for bdr.default_replica_identity

djw-m:
## What Changed?

Added warning to reference entry

### DIFF
--- a/product_docs/docs/pgd/5/reference/pgd-settings.mdx
+++ b/product_docs/docs/pgd/5/reference/pgd-settings.mdx
@@ -60,6 +60,15 @@ above.
 Setting the replica identity of tables to `full` increases the volume of
 WAL written and the amount of data replicated on the wire for the table.
 
+!!! warning On setting bdr.default_replica_identity to default
+When setting `bdr.default_replica_identity` to `default` using `ALTER SYSTEM`
+always quote the value, like this:
+```sql
+ALTER SYSTEM SET bdr.default_replica_identity="default";
+```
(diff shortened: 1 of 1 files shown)


sumarjee (approved):
Should it be better to say "PGD default which is auto"  instead of "actual default which is auto"
Up there is mention for auto as "This is PGD default behavior".

If you feel current one is better, I am fine with that too. Hence approving.

sumarjee (approved):

product_docs/docs/pgd/5/reference/pgd-settings.mdx
```diff
@@ -60,6 +60,15 @@ above.
 Setting the replica identity of tables to `full` increases the volume of
 WAL written and the amount of data replicated on the wire for the table.
 
+!!! warning On setting bdr.default_replica_identity to default
+When setting `bdr.default_replica_identity` to `default` using `ALTER SYSTEM`
+always quote the value, like this:
+```sql
+ALTER SYSTEM SET bdr.default_replica_identity="default";
+```
+This is because default, unquoted, is a special value to the [ALTER SYSTEM](https://www.postgresql.org/docs/current/sql-altersystem.html) command which triggers the removal of the setting from the configuration file. This results in the system using the actual default setting which is `auto`.
```
djw-m, line 69:
```suggestion
This is because default, unquoted, is a special value to the [ALTER SYSTEM](https://www.postgresql.org/docs/current/sql-altersystem.html) command which triggers the removal of the setting from the configuration file. This results in the system using the PGD default setting which is `auto`.
```

djw-m:
No, fair comment. Changed,

sumarjee (approved):
looks fine.

PR #5896, opened 2024-08-01
Fixup on epas upgrade and security pages to fix earlier merge

djw-m:
## What Changed?

Change link title in epas


djw-m:
The link in the security advisory is absolute and will only resolve once the entire PR and staging is published.
It is fine.

PR #5912, opened 2024-08-06
Transporter/reader suggestions

gvasquezvargas:
## What Changed?

Some suggestions to use the reader. 
In a different PR, I saw the reader was being called "EDB Transporter Writer". If that is the case, we'll need to update here as well.

### DIFF
--- a/product_docs/docs/transporter/ba_preview/getting_started/config_reader.mdx
+++ b/product_docs/docs/transporter/ba_preview/getting_started/config_reader.mdx
@@ -1,5 +1,5 @@
 ---
-title: "Configuring the source reader"
+title: "Configuring and running the source reader"
 ---
 
 ## Getting credentials
--- a/product_docs/docs/transporter/ba_preview/getting_started/config_reader.mdx
+++ b/product_docs/docs/transporter/ba_preview/getting_started/config_reader.mdx
@@ -82,16 +82,20 @@ cd /opt/cdcreader
 ### Parameters explanation
 
 #### DBZ_ID
-ID to identify the reader. Users can define the ID they think will be useful with below constraints:
 
-1. ID length should lower or equal to 255
-2. ID can only contain lowercase characters, uppercase characters, numbers, underscore(_) and hyphen(-)
-3. ID should be unique, two reader instances should not use the same ID
+This is the name you assign to a migration to identify it. 
+Take into account the following limitations: 
+
+- The ID cannot be longer than 255 characters.
+- The ID can contain lowercase characters, uppercase characters, numbers, underscores(_) and hyphens(-)
+- The ID must be unique; two reader instances should not have the same ID.
 
 #### RW_SERVICE_HOST
-URL of the transporter-rw-service is always https://transporter-rw-service.biganimal.com
+
+Uncomment the line. The URL of the `transporter-rw-service` is always https://transporter-rw-service.biganimal.com.
 
 #### TLS_PRIVATE_KEY_PATH
+
 Private key of user migration credential. Reader's http client uses it to mTLS with transporter-rw-service
 
 #### TLS_CERTIFICATE_PATH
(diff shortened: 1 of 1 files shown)


product_docs/docs/transporter/ba_preview/getting_started/config_reader.mdx
```diff
@@ -72,37 +72,41 @@ Set the following environment variables in `/opt/cdcreader/run-cdcreader.sh` wit
 #export QUARKUS_LOG_CATEGORY__COM_ENTERPRISEDB__LEVEL=DEBUG
 ```
 
-## Running the reader
```
gvasquezvargas, line 75: Moved this to the bottom of the page, as this is the command the customer will have to run _after_ configuring all values.

product_docs/docs/transporter/ba_preview/getting_started/config_reader.mdx
```diff
@@ -72,37 +72,41 @@ Set the following environment variables in `/opt/cdcreader/run-cdcreader.sh` wit
...
-
 ### Parameters explanation
 
 #### DBZ_ID
-ID to identify the reader. Users can define the ID they think will be useful with below constraints:
 
-1. ID length should lower or equal to 255
-2. ID can only contain lowercase characters, uppercase characters, numbers, underscore(_) and hyphen(-)
-3. ID should be unique, two reader instances should not use the same ID
+This is the name you assign to a migration to identify it.
```
yidian1997, line 79: This DBZ_ID is assigned to a reader not a migration

product_docs/docs/transporter/ba_preview/getting_started/config_reader.mdx
```diff
@@ -72,37 +72,41 @@ Set the following environment variables in `/opt/cdcreader/run-cdcreader.sh` wit
...
+Take into account the following limitations: 
+
+- The ID cannot be longer than 255 characters.
+- The ID can contain lowercase characters, uppercase characters, numbers, underscores(_) and hyphens(-)
+- The ID must be unique; two reader instances should not have the same ID.
 
 #### RW_SERVICE_HOST
-URL of the transporter-rw-service is always https://transporter-rw-service.biganimal.com
+
+Uncomment the line. The URL of the `transporter-rw-service` is always https://transporter-rw-service.biganimal.com.
```
yidian1997, line 88: Why do we need to add "Uncomment the line." here

product_docs/docs/transporter/ba_preview/getting_started/config_reader.mdx
gvasquezvargas: Is this also the name that will later appear as the source name in the EDB Cloud Service (BigAnimal) UI? If yes, we could say: 
```suggestion
This is the name you assign to identify a reader. This name will later appear as a _source_ in the EDB Cloud Service UI. 
```
If not, we can simply say: 
```suggestion
This is the name you assign to a reader.
```

product_docs/docs/transporter/ba_preview/getting_started/config_reader.mdx
gvasquezvargas: You are right, we are not suggesting to uncomment all other lines of the reader when configuring them, so we should not explicitly call it out here. What about this? 

```suggestion
Specifies the URL of the service that will host the migration. `transporter-rw-service` is always https://transporter-rw-service.biganimal.com.
```

product_docs/docs/transporter/ba_preview/getting_started/config_reader.mdx
Xander-run, line 1: How about changing the title to "Configuring & starting the source reader" or something similar to indicate that the startup script is included?

gvasquezvargas: Good point, I altered the title as well ✅

Xander-run (approved):
LGTM!

PR #5921, opened 2024-08-07
fix efm auto.allow.hosts property description

EFM-Bobby:
I think something automated clobbered this, because the descriptive text and the property comment description both were garbled. The descriptive text now makes sense again, though it's a long sentence and I'm open to suggestions.

I changed "update" to "set" because there isn't actually a list to update since the cluster doesn't exist before starting this node. Am ok with going back though.


djw-m (approved):
LGTM

PR #5928, opened 2024-08-08
Pgd/5/consistency/crdt/docs 287

jpe442:
Break up CRDT page into multiple sections as outlined in https://enterprisedb.atlassian.net/browse/DOCS-287.


djw-m (commented):
The overview seems to contain most of the practical usage information and little actual overview the way it is currently split.

I'd take the opening section and the closing parts on CRDTs in Postgres and make that the overview and move the remaining content into a simple "Using CRDTs" section.

djw-m (approved):
LGTM (see changes)

PR #5908, opened 2024-08-04
fix(ET-1265): Updates migration service documentation

ksnavely:
## What Changed?

I've updated some documentation content based on testing and review.

## Not In Scope

Anything unrelated to MP/MTK schema management. I made some small updates but am stopping to avoid scope creep.

PR #5934, opened 2024-08-13
Data Migration Service: general edits

gvasquezvargas:
## What Changed?

- General edits for clarity
- Included Kyle's edits from https://github.com/EnterpriseDB/docs/pull/5908/files
- Renamed EDB Transporter to EDB Data Migration Service or EDB DMS Reader depending on the context. (I have not renamed mentions in the reader install yet to avoid conflicts)


product_docs/docs/transporter/ba_preview/limitations.mdx
```diff
@@ -19,6 +19,6 @@ Unsupported Oracle data types include:
 - XML
 - Spatial
 
-Don't use EDB Transporter for replicating Oracle tables that contain blob, clob, or nclob columns and are expected to get UPDATE changes. Streaming UPDATE changes currently aren't handled correctly.
+EDB Data Migration Service supports replicating Oracle tables that contain BLOB, CLOB, or NCLOB columns only if these also have the `PRIMARY KEY` constraint. If the tables don;t have the `PRIMARY KEY` constraint, the streaming replication will only support INSERT operations.
```
gvasquezvargas, line 22: @yidian1997  I saw you had a conversation with @ksnavely about this in the other thread (https://github.com/EnterpriseDB/docs/pull/5908/files#r1703232801). Is this wording more accurate?

gvasquezvargas:
```suggestion
EDB Data Migration Service supports replicating Oracle tables that contain BLOB, CLOB, or NCLOB columns only if these also have the `PRIMARY KEY` constraint. If the tables don't have the `PRIMARY KEY` constraint, the streaming replication will only support INSERT operations.
```

product_docs/docs/transporter/ba_preview/upgrading.mdx
```diff
@@ -1,18 +1,17 @@
 ---
 title: "Upgrading"
-description: Learn how to upgrade EDB Transporter to a more recent version. 
+description: Learn how to upgrade EDB Data Migration Service to a more recent version. 
 ---
 
-Upgrades to Reader are straightforward but involve a temporary halt to any data migration.
+Upgrading the EDB Data Migration Reader is straightforward but involves a temporary halt to any data migration.
```
gvasquezvargas, line 6: @yidian1997 @Xander-run 
I would like to have more context around upgrades. When specifically would they take place? I don't think we should recommend interrupting a streaming migration for an upgrade, correct? We would rather wait until a migration is completed, then upgrade the reader, and then use the new reader for other migrations? 

I was a bit confused about the `1. Stop the EDB Data Migration Reader.` step below.

jperozoa: I'm agree, probably upgrading the reader in the middle of a migration makes not a lot of sense. I think this only makes sense if the reader is involved in a very long migration (weeks) and there is an important change in the reader that needs to be deployed (like a security vulnerability)

jperozoa: Stopping the reader is just kill the process

jperozoa (approved):

PR #5939, opened 2024-08-13
Fix PEM installer links

josh-heyer:
## What Changed?

One link (per installation target) was pointing to PEM 8. This was always incorrect; now it also does not work.


djw-m (approved):
LGTM

PR #5957, opened 2024-08-19
Changed folder names to reflect product name and version

gvasquezvargas:
## What Changed?

https://enterprisedb.atlassian.net/browse/DOCS-931 

Changed folder names to reflect the customer-facing product name and the major version we are planning on releasing. 

TODO: gatsby is not letting me build at the moment, I probably need to adapt a .js file for it to recognize the new path.


jperozoa (approved):
lgtm
I only see one ref to transporter here: https://github.com/EnterpriseDB/docs/blob/cebc457dd5c79fa0a75e829440b42b274feaacac/src/components/icon/iconNames.js#L78 but I'm not sure if it's used somehow

djw-m:
> lgtm I only see one ref to transporter here:
> 
> https://github.com/EnterpriseDB/docs/blob/cebc457dd5c79fa0a75e829440b42b274feaacac/src/components/icon/iconNames.js#L78
> 
> but I'm not sure if it's used somehow

It isn't that's the old icon entry.

gvasquezvargas:
@jperozoa good catch. That is the name of this icon the our internal library of available icons:

Screenshot 2024-08-20 at 14 26 06

But it is not visible to the customer at all, so there is no need to change the name there (it could mess up with the build and is a larger effort).

PR #5980, opened 2024-08-23
Rename directory and add readme for clarity

djw-m:
## What Changed?

Rename directory and add readme for clarity


gvasquezvargas (approved):
I was worried about cross-links but the checker did its job! 🥳

PR #5989, opened 2024-08-26
DOCS-961 Add Release Note For EPAS 15.8.1

julienmarcbrown:


josh-heyer (approved):

theadamwright (commented):
Minor suggestion based on internal feedback  "change "the old" to "the layout of previous database server versions"

Updated text: 
A previous release introduced a new in_create field to the LogicalDecodingContext structure and changed its memory layout. This means that any code that expects the old layout will no longer be compatible. The release fixed the issue to ensure that the layout of previous database server versions is still compatible with the creation of a logical replication slot.

PR #6007, opened 2024-08-29
Update working-with-ai-data-in-postgres.mdx

kunliuedb:
I believe that `Working without auto embedding` should be on the same level with `Working with auto embedding`. 

And there is a missing quick link of `Working without auto embedding` on the right side.


djw-m:
The link would be missing as it isn't at the right level and deepToC isn't enabled.

djw-m (approved):
LGTM

PR #6011, opened 2024-08-29
Pgd/durabiltity/docs-942

jpe442:
Moving Eager conflict resolution from Consistency to Durability on the Group Commit page as Eager conflict resolution is a conflict resolution option for the Group Commit commit scope.


djw-m:
This appears to be somewhat out of date and likely should be being applied to the 5.6 branch.

mwanner (approved):

PR #6015, opened 2024-08-30
Changed wording to remove references to Eager All-Node Replication from Known Issue

jpe442:
It looks like there was a Known Issue left over from PGD version 4 that made it into 5, even though things had changed for 5. Namely, it was the case in version 4 that Group Commit was incompatible with Eager Replication, but now it is actually a conflict resolution option of Group Commit, replacing the global scope: "The global scope no longer exists. To create scope with the same behavior, use [Group Commit](https://www.enterprisedb.com/docs/pgd/latest/durability/group-commit/)."


djw-m (approved):
LGTM

Don't forget to apply to 5.6 branch

djw-m (approved):
LGTM

PR #6025, opened 2024-09-04
Docs 757 - Adds `ORIGIN_GROUP` to the docs

jpe442:
We are introducing new syntax sugar such that instead of targeting a specific PGD data node group, you can target `ORIGIN_GROUP` to create a rule that applies to the node group of any transaction-originating data node in the cluster.


jpe442:
should have been aimed at 5.6 branch

PR #6026, opened 2024-09-04
Docs 757 - Adding ORIGIN_GROUP

jpe442:
5.6 introduces ORIGIN_GROUP "syntax sugar" such that instead of targeting a specific node_group with your commit scope you can target ORIGIN_GROUP to create a rule that applies to the node group of any transaction-originating data node in the cluster.


djw-m:
This apparently has 287 changed files. That doesn't sound good.

josh-heyer:
⚠Rebased! Pull latest before continuing work.

product_docs/docs/pgd/5.6/commit-scopes/commit-scopes.mdx
```diff
@@ -175,4 +176,40 @@ SELECT bdr.alter_node_group_option(
 );
 ```
 
+### ORIGIN_GROUP syntax sugar
```
djw-m, line 179: Do not refer to it as syntax sugar. From the users POV it is a useful function and they know nothing of the internals of the syntax that would make it sugary.

product_docs/docs/pgd/5.6/commit-scopes/commit-scopes.mdx
```diff
@@ -122,10 +124,9 @@ Set `bdr.commit_scope` to `local` to  use the PGD default async replication.
 
 ## Origin groups
```
djw-m, line 125: This may be large enough to be its own page now.

product_docs/docs/pgd/5.6/reference/commit-scopes.mdx
```diff
@@ -41,7 +45,24 @@ commit_scope_kind:
...
 ```
 
-Where `node_group` is the name of a PGD data node group.
+Where `node_group` is the name of a PGD data node group. 
+ 
+## Commit scope targets
+
+### ORIGIN_GROUP
+
+Instead of targeting a specific group, you can also use `ORIGIN_GROUP` syntax sugar, which dynamically refers to the bottommost group from which a transaction originates. Therefore, if you have a top level group, `top_group`, and two subgroups as children, `left_dc` and `right_dc`, then adding a commit scope like:
```
djw-m, line 54: Again, remove the syntax sugar reference.

product_docs/docs/pgd/5.6/commit-scopes/commit-scopes.mdx
jpe442: Oh I see good point.

product_docs/docs/pgd/5.6/reference/commit-scopes.mdx
jpe442: done.

product_docs/docs/pgd/5.6/commit-scopes/commit-scopes.mdx
jpe442: I see what you are saying in that it is long, but I am worried about its discoverability. As in It makes sense that it is on this page on commit scopes to me and I worry that if it was on its own page it would be hard to find.

djw-m (approved):
LGTM

PR #6046, opened 2024-09-11
DMS: added purl for UI link

gvasquezvargas:
## What Changed?

At the moment, the staging environment string links to a page in our _preview_ documentation. 
After we go live, the string should link to the _live_ documentation. 

I have set up a permanent URL that we can embed in the UI, so the link does not become outdated, and we can change the content/location of the page without worrying about breaking links. 

Screenshot 2024-09-04 at 14 09 00

The link the UI team can embed to the string will be: 
https://www.enterprisedb.com/docs/purl/dms/configure_source


jperozoa:
Updated the FF with this new URL: https://edb-starlight.slack.com/archives/C02CVAC3938/p1726046008626609

PR #6049, opened 2024-09-12
PGD: improve description of "transaction_id" pseudo-GUC (5.6)

ibarwick:
Current doc: https://www.enterprisedb.com/docs/pgd/latest/reference/functions/#transaction_id

## What Changed?

This is only queryable as a reported GUC via libpq's PQparameterStatus() function, and is never accessible at SQL level. Rewrite the description to clarify this, linking to a helpful code example while we're at it.

Also rework the introductory section to make it clearer that one of the GUC-like objects about to be described is not query-able via SQL.

5.6 port of PR #6048.

DOCS-1028.


djw-m (commented):
LGTM after fixing the link

PR #5917, opened 2024-08-06
feat(ET-480): Add documents for cdcreader installation

Xander-run:
## What Changed?

- Removed unused installation doc.
- Updated the installation process to utilize package managers.
- Added installation docs for different distros.

PR #6055, opened 2024-09-12
Templatized install for DMS Reader

dwicinas:
Revised install template for DMS Reader and generated updated files.


josh-heyer:
FYI: I rebased to remove commits that came along from old parent (front page redesign branch).

ksnavely:
Javier notes that we do not support all of these platforms yet. We won't consider the merge of this PR as blocking for our FF lift, so we can make some tweaks.

advocacy_docs/edb-postgres-ai/migration-etl/data-migration-service/getting_started/installing/linux_x86_64/index.mdx
gvasquezvargas, line 1: I don't think we need this folder/level. It basically just repeats exactly the same information we state in the index.mdx file of the installing folder. And since we only have x86-64 distros at this point, I think it would be clearer to delete it for now.
Is it possible to get rid of this folder, and nest all the installation tutorials under /installing ?

advocacy_docs/edb-postgres-ai/migration-etl/data-migration-service/getting_started/installing/index.mdx
```diff
@@ -1,30 +1,31 @@
 ---
-navTitle: Installing EDB DMS Reader
-title: Installing EDB DMS Reader on Linux
+navTitle: Installing
+title: Installing EDB Data Migration Service Reader on Linux
+
```
gvasquezvargas, line 4: Is it possible to add the following to the frontmatter of this index.mdx?  
```suggestion
indexCards: none

```
That way, we don't have both a Table of Content, and navigation tiles. That way, we can avoid unnecessary repetitions. 
Screenshot 2024-09-24 at 12 39 03 copy

advocacy_docs/edb-postgres-ai/migration-etl/data-migration-service/getting_started/installing/linux_x86_64/index.mdx
```diff
@@ -1,48 +1,47 @@
 ---
-title: "Installing EDB DMS Reader on Linux x86 (amd64)"
+title: "Installing EDB Data Migration Service Reader on Linux x86 (amd64)"
 navTitle: "On Linux x86"
```
gvasquezvargas, line 4: If for some reason we cannot get rid of this page as suggested in the previous comments: 

Can we add the following to the frontmatter of this index.mdx?  
```suggestion
indexCards: none

```
That way, we don't have both a Table of Content, and navigation tiles. That way, we can avoid unnecessary repetitions. 
Screenshot 2024-09-24 at 12 39 03 copy

gvasquezvargas:
> Javier notes that we do not support all of these platforms yet. We won't consider the merge of this PR as blocking for our FF lift, so we can make some tweaks.

@jperozoa could you please name the platforms we don't support yet? This PR contains the same Operating Systems @Xander-run added a while back in https://github.com/EnterpriseDB/docs/pull/5917/files.

jperozoa:
> > Javier notes that we do not support all of these platforms yet. We won't consider the merge of this PR as blocking for our FF lift, so we can make some tweaks.
> 
> @jperozoa could you please name the platforms we don't support yet? This PR contains the same Operating Systems @Xander-run added a while back in https://github.com/EnterpriseDB/docs/pull/5917/files.

@gvasquezvargas What we have in this PR is correct, when I fast reviewed it with Kyle I thought I saw installers for pppoe, but I don't see it now

install_template/templates/products/edb-data-migration-service-reader/ppc64le_index.njk
gvasquezvargas, line 1: @dwicinas  This is what Javier was referring to, since we don't support any ppc64te architectures. But as I look at the config.yaml, it seems like we are not using this njk file to build any ppc64le installation docs. Am I assuming correctly that you build this in advance so in case we decide to support it the base njk is already there? And it is not doing any "harm" by existing there?

advocacy_docs/edb-postgres-ai/migration-etl/data-migration-service/getting_started/installing/linux_x86_64/index.mdx
dwicinas: The idea behind the templates is that all products use the same format. At the simplest level, I could make the change you propose, but that change would then propagate to all other products, which I don't think we want. Or, I could make a one off change for this product, but I every time I generate new install instructions (which happens frequently!) I would have to remember to keep modifying the results to maintain the one-off nature of this product. Or, finally, I could enlist Josh to customize the template for this one product. I don't think any of those approaches are worthwhile. I think we should live with the duplication for now. At some point in the future they will probably support other platforms and then it will not be so obvious.

install_template/config.yaml
```diff
@@ -1,4 +1,33 @@
 products:
+  - name: EDB Data Migration Service Reader
+    platforms:
+      - name: RHEL 8 or OL 8
+        arch: x86_64
+        supported versions: [2.0]
```
gvasquezvargas, line 6: @dwicinas  is there any chance that we can get rid of this line for all platforms? The problem is that we have the 2.14.10 version for now, but that might get bumped regularly (without any changes to the installation path). 

Alternatively, if we do need a line for the version, could we do `2.x` ? That way the reference won't get outdated as quickly.

gvasquezvargas: PS: applies to all platforms below.

advocacy_docs/edb-postgres-ai/migration-etl/data-migration-service/getting_started/installing/index.mdx
dwicinas: This is a good idea. However, if I make this change it will apply to all products. I would probably make this change in a separate PR because it's going to affect a hundred topics or so. @nidhibhammar what do you think of this idea?

install_template/templates/products/edb-data-migration-service-reader/ppc64le_index.njk
dwicinas: @gvasquezvargas That's correct. We are not using this structure right now, but it's there when they do start suppporting the product on other architectures.

install_template/config.yaml
dwicinas: @gvasquezvargas We use this approach for a few products that have frequent upgrades that don't require new documentation.

advocacy_docs/edb-postgres-ai/migration-etl/data-migration-service/getting_started/installing/index.mdx
gvasquezvargas: Okay, that is a good idea. Let me open a Jira for this work https://enterprisedb.atlassian.net/browse/DOCS-1040

install_template/config.yaml
gvasquezvargas: Synced with David, and we'll change to 2.

gvasquezvargas (approved):
LGTM!

PR #6060, opened 2024-09-17
PGD 5.6: update description for "bdr.join_node_group()"

ibarwick:
## What Changed?

From PGD 5.6, database structure will never be synchronized to a witness server, which represents a minor behaviour change compared to previous versions, and means the "synchronize_structure" parameter, if explicitly set to "all", will be forcibly overridden (a NOTICE will be emitted noting that).

DOCS-1029.


djw-m (approved):
LGTM

PR #6083, opened 2024-09-23
Cloud Service title fix and image update

nidhibhammar:
## What Changed?

Edited the title and added a new image as per the suggestion provided by Yogesh jain


gabrosys (commented):
LGTM! Thanks 😄

josh-heyer:
Ok; two things happened here @nidhibhammar 

1. Netlify choked on the first build - just hung deploying. That meant this didn't actually deploy (or finish building and cache the results) until the second commit, where you renamed the image. 
2. That build resulted in the page without a valid image, which left a reference that wouldn't work. 
3. You fixed the image name in the next (third) commit, but the page didn't rebuild and the results were cached through subsequent builds.

Fix ended up being a no-op change to the file to trigger a rebuild.

This was sort of a freak occurrence; it wouldn't have affected anything if this had been merged (the cache for this branch only gets used on this branch), but I wanted to track down the cause just to be sure.

YogeshJain96 (approved):
lgtm!

nidhibhammar:
> Ok; two things happened here @nidhibhammar
> 
> 1. Netlify choked on the first build - just hung deploying. That meant this didn't actually deploy (or finish building and cache the results) until the second commit, where you renamed the image.
> 2. That build resulted in the page without a valid image, which left a reference that wouldn't work.
> 3. You fixed the image name in the next (third) commit, but the page didn't rebuild and the results were cached through subsequent builds.
> 
> Fix ended up being a no-op change to the file to trigger a rebuild.
> 
> This was sort of a freak occurrence; it wouldn't have affected anything if this had been merged (the cache for this branch only gets used on this branch), but I wanted to track down the cause just to be sure.

Thanks for your findings on this @josh-heyer.

PR #6090, opened 2024-09-24
Edits to Added note on non-ascii characters not being supported PR6024

ebgitelman:
Removed bold-italic emphasis. If there's something that needs to be called out here, use additional words rather than emphasis, which we don't have any standard for, and it doesn't seem apt in this case anyhow. For example, you could do something like this:

- If you're using the installation wizard to perform a system upgrade rather than a fresh install...

The added context will provide the emphasis if the concern is that readers might not see that this is referring to an upgrade. Alternatively, keep this text in its own note rather than combining with the new one and place each note more precisely within the content.


gvasquezvargas (approved):
LGTM!

PR #6116, opened 2024-09-30
Edits to Rework bdr.drop_node PGD pr6056

ebgitelman:

### DIFF
--- a/product_docs/docs/pgd/5/reference/functions-internal.mdx
+++ b/product_docs/docs/pgd/5/reference/functions-internal.mdx
@@ -165,7 +165,7 @@ This function removes the metadata for a given node from the local
 database. The node can be either:
 
 -   The local node, in which case it removes all the node metadata, including information about remote nodes.
--   A remote node, in which case it only removes metadata for that specific node.
+-   A remote node, in which case it removes metadata only for that specific node.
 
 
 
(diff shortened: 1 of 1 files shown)


product_docs/docs/pgd/5/reference/functions-internal.mdx
```diff
@@ -165,7 +165,7 @@ This function removes the metadata for a given node from the local
 database. The node can be either:
 
 -   The local node, in which case it removes all the node metadata, including information about remote nodes.
--   A remote node, in which case it only removes metadata for that specific node.
+-   A remote node, in which case it removes metadata only for that specific node.
```
ebgitelman, line 168: I read this and the previous bullet a few times and I believe this is what you intend? Otherwise it reads like it's only removing metadata. Please verify this is correct.

codepope: It literally only removes metadata and does nothing else.

ebgitelman, line 168:
```suggestion
-   A remote node, in which case it removes only metadata for that specific node.
```

ebgitelman: Thanks for clarifying. I revised it so there's no ambiguity as to what "only" is referring to.

djw-m (approved):
LGTM

PR #6118, opened 2024-09-30
Docs 958b - new observability SQL views

jpe442:
Adds new observability views across Manager Worker, Routing, Raft, Replication, and Commit scopes.


djw-m (approved):
LGTM

PR #6119, opened 2024-09-30
EFM-1698 Remove "Agent" column from cluster-status output.

EFM-Bobby:
This column was removed ~4 years ago, see EFM-993.

### DIFF
--- a/product_docs/docs/efm/4/06_monitoring_efm_cluster.mdx
+++ b/product_docs/docs/efm/4/06_monitoring_efm_cluster.mdx
@@ -23,11 +23,11 @@ The `efm cluster-status` [cluster properties file](07_using_efm_utility/#efm_clu
 The following status report is for a cluster named edb that has three nodes running:
 
 ```text
-Agent Type  Address               Agent   DB        VIP
------------------------------------------------------------------------
-Standby     172.19.10.2            UP     UP        192.168.225.190
-Standby     172.19.12.163          UP     UP        192.168.225.190
-Primary     172.19.14.9            UP     UP        192.168.225.190*
+Agent Type  Address               DB        VIP
+---------------------------------------------------------------
+Standby     172.19.10.2           UP        192.168.225.190
+Standby     172.19.12.163         UP        192.168.225.190
+Primary     172.19.14.9           UP        192.168.225.190*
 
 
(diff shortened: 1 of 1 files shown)


djw-m:
The column has gone but the surrounding text refers to the status of the agent and without that column it doesn't seem to make sense?

product_docs/docs/efm/4/06_monitoring_efm_cluster.mdx
```diff
@@ -55,11 +55,11 @@ Standby database(s) in sync with primary. It is safe to promote.
 The cluster status section provides an overview of the status of the agents that reside on each node of the cluster:
```
djw-m, line 55: Specifically, this says the status of the agents, but the table shown doesn't give an agent status, only an agent type. You may be able to infer the agent status from the DB status but....

product_docs/docs/efm/4/06_monitoring_efm_cluster.mdx
```diff
@@ -90,9 +90,9 @@ Standby  172.19.10.2     0/4000638          0/4000638
 If a database is down or if the database was restarted, but the resume command was not yet invoked, the state of the agent that resides on that host is idle. If an agent is idle, the cluster status report includes a summary of the condition of the idle node. For example:
```
djw-m, line 90:
```suggestion
If a database is down or if the database was restarted, but the resume command was not yet invoked, the state of the agent that resides on that host is idle. If an agent is idle, the cluster status report shows that status in the Agent Type column. For example:
```

product_docs/docs/efm/4/06_monitoring_efm_cluster.mdx
EFM-Bobby: Correct, the text is just referring to the general set of info about the agent: what type it is, does it have the VIP, whatever else. So I think the existing text is still ok.

product_docs/docs/efm/4/06_monitoring_efm_cluster.mdx
EFM-Bobby: Good thinking.

djw-m (approved):
LGTM (concerns addressed)

PR #6124, opened 2024-10-02
Migration Portal: workaround for snapshot too old error

gvasquezvargas:
## What Changed?

https://enterprisedb.atlassian.net/browse/MIG-5130


product_docs/docs/migration_portal/4/known_issues_notes.mdx
```diff
@@ -243,6 +245,59 @@ While using the Oracle default case, you may experience a lower compatibility ra
...
+    JOIN dba_tablespaces dt ON dfs.tablespace_name = dt.tablespace_name 
+    WHERE dt.contents = 'UNDO' GROUP BY dfs.tablespace_name; 2 3 4
+    ```
+
+Next, determine the number of MB your environment required and increase the allocated storage space:
+
+1.  Monitor the volume of undo data generated by transactions:
+    
+    ```sql
+    SELECT SUM(undoblks * 8192)/1024/1024 AS undo_mbs FROM v$undostat;
```
gvasquezvargas, line 282: @gabbasb  
I fear the customer might be confused about what to do with this value. Here, we get a value of used MB (per minute? per hour? per 24 hours?), and in the next step, we tell them to multiply the MB/min value by the number of required retention minutes. 

According to my research, the default `v$undostat` view displays the used MB in the last 24 hours. Does that mean that we have to divide this value by 1440 (total minutes in 24 hours) to get the MB/min value? 

And then, in the next step, we would have to multiply that value by the number of minute we want undo data to be retained?

product_docs/docs/migration_portal/4/known_issues_notes.mdx
```diff
@@ -243,6 +245,59 @@ While using the Oracle default case, you may experience a lower compatibility ra
...
+    ```
+
+    !!!note 
+    The resulting value reflects the used MB over a time interval specified in your system's `v$undostat` view.
```
gvasquezvargas, line 286: @gabbasb  I added this note to have some sort of guidance for the customer to understand the impact of the value. I wonder is we want to advise them to look up the time interval with something like `SELECT MIN(begin_time), MAX(end_time) FROM v$undostat;` or something similar?

product_docs/docs/migration_portal/4/known_issues_notes.mdx
gabbasb: @gvasquezvargas According to the docs
https://docs.oracle.com/en/database/oracle/oracle-database/19/refrn/V-UNDOSTAT.html
Each row in the view keeps statistics collected in the instance for a 10-minute interval.
The table holds start and end time as well.
However that's not the point.
What we are asking the user to do is to monitor the the total number of undo blocks consumed during peak usage hours using the V$UNDOSTAT.
This will give the user an idea of how much space the user needs to increase to avoid the error.

product_docs/docs/migration_portal/4/known_issues_notes.mdx
gabbasb: I don't think so.

gabbasb (approved):

djw-m (approved):
LGTM

PR #6146, opened 2024-10-10
PostGIS: 3.4 updates

gvasquezvargas:
## What Changed?

- Updated support matrix with a row for 3.4.2 and a column for EPAS 16. Removed EPAS 11 column. 
- Replaced `yum` instances with `dnf` as it is the current package manager
- Updated upgrade example and included a note clarifying that the example applies to all upgrades from versions after 3.1.4. 
https://enterprisedb.atlassian.net/browse/DB-2753


djw-m (approved):
LGTM

sagaryedida (approved):

PR #6158, opened 2024-10-16
Fix link with readable absolute link

djw-m:

## What Changed?

Fixed link for adding key to point into console using projects settings security with an absolute link as it was climbing up 3 directory levels and missing.


gvasquezvargas (approved):
LGTM!

PR #6170, opened 2024-10-21
MTK 55.10.0: computed columns

gvasquezvargas:
### Content 
https://enterprisedb.atlassian.net/browse/DOCS-1019 
Added `-mapVirtualToRegular` to the MTK command options. Added a description of the flag, use cases, and an example. Release notes entry will be handled separately. 

### Comments on release branch strategy 

This is a v55.10.0 feature. 
I have created a `DOCS-1088-Tracking-55.10-work` release branch to stage all 55.10.0 work.
This PR is branched off of said release branch, so once we get reviews and approvals, we can merge to the release branch, which will be staged and remain unmerged until release time.

### DIFF
--- a/product_docs/docs/migration_toolkit/55/07_invoking_mtk/08_mtk_command_options.mdx
+++ b/product_docs/docs/migration_toolkit/55/07_invoking_mtk/08_mtk_command_options.mdx
@@ -23,7 +23,7 @@ The command options that work with Migration Toolkit are grouped by their behavi
 | [Import options](#import-options)                                                           | -sourcedbtype, -targetdbtype, -schemaOnly, -dataOnly                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               |
 | [Schema creation options](#schema-creation-options)                                         | -dropSchema, -targetSchema                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         |
 | [Schema object selection options](#schema-object-selection-options)                         | -allTables, -tables, -excludeTables,<br /><br />-constraints, -ignoreCheckConstFilter,<br /><br />-skipCKConst, -skipFKConst,<br /><br />-skipColDefaultClause,<br /><br />-indexes, -triggers,<br /><br />-allViews, -views, -excludeViews,<br /><br />-allSequences, -sequences, -excludeSequences,<br /><br />-allProcs, -procs, -excludeProcs,<br /><br />-allFuncs, -funcs, -excludeFuncs,<br /><br />-checkFunctionBodies,<br /><br />-allPackages, -packages, -excludePackages,<br /><br />-allDomains,<br /><br />-allQueues, -queues, -excludeQueues<br /><br />-allRules,<br /><br />-allgroups, -groups |
-| [Migration options](#migration-options)                                                     | -truncLoad, -enableConstBeforeDataLoad,<br /><br />-retryCount, -safeMode, -fastCopy,<br /><br />-analyze, vacuumAnalyze, -replaceNullChar,<br /><br />-copyDelimiter, -batchSize,<br /><br />-cpBatchSize, -lobBatchSize,<br /><br />-fetchSize, -filterProp<br /><br />-customColTypeMapping, -customColTypeMappingFile                                                                                                                                                                                                                                                                                          |
+| [Migration options](#migration-options)                                                     | -truncLoad, -enableConstBeforeDataLoad,<br /><br />-retryCount, -safeMode, -fastCopy,<br /><br />-analyze, vacuumAnalyze, -replaceNullChar,<br /><br />-copyDelimiter, -batchSize,<br /><br />-cpBatchSize, -lobBatchSize,<br /><br />-fetchSize, -filterProp<br /><br />-customColTypeMapping, -customColTypeMappingFile<br /><br />-mapVirtualToRegular                                                                                                                                                                                                                                                          |
 | [Connection retry options](#connection-retry-options)                                       | -connRetryCount, -connRetryInterval, -abortOnConnFailure                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           |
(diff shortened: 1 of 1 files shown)


product_docs/docs/migration_toolkit/55/07_invoking_mtk/08_mtk_command_options.mdx
```diff
@@ -488,6 +488,24 @@ Specify each entry in the file on a separate line in a key=value pair. The left
 
 <div id="parallel_loading" class="registered_link"></div>
 
+### `-mapVirtualToRegular`
+
+Include the `-mapVirtualToRegular` option to migrate virtual, computed or generated columns in the source database to regular columns in the target database. 
+
+Turning virtual columns to regular columns can help speed up the migration process. Additionally, it can help you optimize the performance of queries by reducing complexity and computing resources.
```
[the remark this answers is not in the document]
gvasquezvargas: @zahid75 Thanks for clarifying, what do you think about the following: 
```suggestion
Turning virtual columns to regular columns is helpful when the expression used to compute the values in the source database column is not supported in the target database or cannot be translated into an expression supported by the target database. With this option, you can still migrate the column's values to the target, but the values will no longer be computed dynamically, as they are stored in a regular column.
```

product_docs/docs/migration_toolkit/55/07_invoking_mtk/08_mtk_command_options.mdx
```diff
@@ -488,6 +488,24 @@ Specify each entry in the file on a separate line in a key=value pair. The left
...
+
+For data migrations, including the option extracts the computed values from all computed columns in the source database and inserts those values into regular columns in the target database.
+
+For example, in: 
+
+```
+./runMTK.sh -sourcedbtype sqlserver -targetdbtype enterprisedb -mapVirtualToRegular -allTables <schema_scope>
+```
+
+A column called `FullName` that is computed from other two columns called `FirstName` and `LastName` in the source database, is migrated as a regular column called `FullName` in the target database. In the target, this column is no longer dynamically generated, but displays the `FullName` value the source database had at the time the migration was performed.
```
gvasquezvargas, line 507: @zahid75 @RaviKumarEDB 
I added a very easy example here for simplicity, but now that I understand the use case a bit better (migrating column's values when expressions are not supported in the target), I wonder if it would be more helpful something that exemplifies the use case better. 

For example, we could show a source database expression that is supported in Oracle but not in Postgres. And then explain how the values could be migrated nonetheless. WDYT?

gvasquezvargas: @salman-hameed-code tagging you here as you were working on the original issue [MTK-1389](https://enterprisedb.atlassian.net/browse/MTK-1389). 

Also tagging @mlewandowski-edb  as I think you have a good understanding of what is supported in Oracle that is not supported in Postgres and might know of a good example to use here.

product_docs/docs/migration_toolkit/55/07_invoking_mtk/08_mtk_command_options.mdx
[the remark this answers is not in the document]
zahid75: Looks good to me.

product_docs/docs/migration_toolkit/55/07_invoking_mtk/08_mtk_command_options.mdx
zahid75: While adding an example will elaborate it further, but in general one should get an understanding of the need to use a mapVirtualToRegular option with the explanation that you have added above.

product_docs/docs/migration_toolkit/55/07_invoking_mtk/08_mtk_command_options.mdx
```diff
@@ -450,13 +450,13 @@ The right side specifies a condition that must be true for each row migrated.
 
 For example, this code migrates only those countries with a `COUNTRY_ID` value that isn't equal to `AR`: 
 
-### `COUNTRIES=COUNTRY_ID<>'AR'`
+#### `COUNTRIES=COUNTRY_ID<>'AR'`
```
mlewandowski-edb, line 453: This appears to have been changed from an h3 to an h4 heading, but I'm not sure why it is a heading at all instead of just text.

Instead of:
> #### `COUNTRIES=COUNTRY_ID<>'AR'`

why not?
> `COUNTRIES=COUNTRY_ID<>'AR'`

product_docs/docs/migration_toolkit/55/07_invoking_mtk/08_mtk_command_options.mdx
mlewandowski-edb: I think the text is sufficient without an example.

product_docs/docs/migration_toolkit/55/07_invoking_mtk/08_mtk_command_options.mdx
```diff
@@ -450,21 +450,21 @@ The right side specifies a condition that must be true for each row migrated.
 
 For example, this code migrates only those countries with a `COUNTRY_ID` value that isn't equal to `AR`: 
 
-### `COUNTRIES=COUNTRY_ID<>'AR'`
+`COUNTRIES=COUNTRY_ID<>'AR'`
```
gvasquezvargas, line 453: @mlewandowski-edb, you are right. The heading formatting made me think this was a "suboption" under `-filterProp,` but I am now realizing the formatting was wrong in a few places. I turned this into normal text. 

I also changed the heading levels of `-customColTypeMapping` and `-customColTypeMappingFile` to be at the same level as the other migration options. At first, I thought these were "suboptions" too, but I now realize they are normal migration options that should be at the same level. 

Can you confirm my understanding is correct?

product_docs/docs/migration_toolkit/55/07_invoking_mtk/08_mtk_command_options.mdx
```diff
@@ -450,21 +450,21 @@ The right side specifies a condition that must be true for each row migrated.
...
 
 This constraint applies to the COUNTRIES table. 
 
 You can also specify conditions for multiple tables. However, the condition for each table must be on a new line in the property file.
 
-### Example
+#### Example
```
mlewandowski-edb, line 459: This seems to be the only case where "Example" is set as a heading within the context of the property descriptions. I would suggesting making this normal unbolded text as is the case for some other properties. For example, instead of
> #### Example

change it to
> Example:

mlewandowski-edb (commented):
One possible update to make. Otherwise, the page looks good

PR #6174, opened 2024-10-22
MTK: Privileges for using parallel loading in migrations from Oracle

gvasquezvargas:
## What Changed?

https://enterprisedb.atlassian.net/browse/MTK-392 

Adding note for migrations from Oracle, as additional user privileges are required.


zahid75 (approved):
Looks fine.

PR #6184, opened 2024-10-25
Relgen: Second gen relnotes generator first commit.

djw-m:
## What Changed?

Builds on relnoted but is independent of it.

relgen.js -p relnotedir

Reads config from src/meta.yml
Then processes all the release notes files in the directory
Creates index.mdx and releasenote files.


josh-heyer:
⚠Rebased to fix link errors

josh-heyer (approved):
This seems pretty solid; it's a clear improvement over what we have now, and it would be good to put this to use during releases as soon as possible.

Todo in the future: 
- more robust handling of input strings (currently possible to break table rendering with poor choice of date or version names)
- Pretty-printing for both Markdown and HTML tables

djw-m:
Noted on the changes. Feeling pretty, might make tickets later.

PR #6185, opened 2024-10-25
Link fixes and expansion of link checker, especially including PG4K/PGD4K APIs

josh-heyer:
## What Changed?

- Link checker now verifies links to yaml files, and does not skip URLs that *appear* to be internal links to resources with extensions (this was originally intended to avoid problems checking yaml files, which are handled in a weird way, but inadvertently skipped checking links to certain paths that contained periods)
- Fix broken links identified after ^
- Fix a bunch of broken slugs in the PG4K and PGD4K API docs. This ended up being a bigger deal, as I hadn't considered the fact that PGD4K depends on a version of PG4K other than the very latest - so its links to PG4K API docs were often broken due to changes in the PG4K API across minor versions. The only expedient way I could think of to address this was to publish the API docs for the version that PGD4K depends on; for completeness, I have expanded on this by publishing ALL versions of the PG4K API docs (with the exception of some very, very early alpha releases). I've also expanded the source script to automatically place new versions of the API docs in a sensible location. Existing URLs and navigation will continue to reference the very latest version of the API docs. 

...that last bullet probably deserves a sanity check. @jpe442 @gvasquezvargas take a look if you can.


gvasquezvargas (commented):
Hey @josh-heyer! 
This looks good to me, a few optional suggestions/comments:

[This index page](https://deploy-preview-6185--edb-docs-staging.netlify.app/docs/postgres_for_kubernetes/latest/pg4k.v1/) currently duplicates the content of the latest API version. We could turn it into a landing page explaining that each PG4k version comes with its dedicated API, and to see the respective API reference, they have to select the one corresponding to their PG4K version? 

I also wonder if we need a table/comment in the PGD4K docs stating the PG4K version that is uses? For example:
PDG4K 1.0.1 uses PG4K API version 1.22.6.
PDG4K 1.0.0 uses PG4K API version xxx. 
Unless the PG4K version is the same for both 1.x.x versions, which makes this obsolete.

josh-heyer:
@gvasquezvargas  The reason for the duplicate content is twofold: 

1. It doesn't break existing links this way - the old URL path to the API doc is the same as the new URL path to the index. Sure, I could update internal links, but - 
2. It provides a stable "latest" URL - e.g., `/postgres_for_kubernetes/latest/pg4k.v1/#postgresql-k8s-enterprisedb-io-v1-Cluster` will give the definition for `Cluster` as long as that exists, regardless of revision. 

That table effectively exists: https://www.enterprisedb.com/docs/postgres_distributed_for_kubernetes/latest/supported_versions/ - I assume we'll add another row if/when we do a minor revision for PGD4K.

gvasquezvargas:
@josh-heyer  gotcha, thanks for clarifying! Approving, then ✅

gvasquezvargas (approved):

djw-m (approved):
LGTM Merge merge merge

PR #6193, opened 2024-10-30
Update 04_using_descriptors.mdx

yuki-tei:
Hi team,

Unfortunately, there might be some issues in the sample code on this page.

## Summery
a) Descriptor name right after identifier DESCRIPTOR should not be enclosed by '', that would lead to error. 
b) Comment in the end of the program is not enclosed by */ .
c) Exec command and program content is in the same code block, that might confuse customers who want to copy program,
    using "Copy" button.

## What Changed?
Line 32 and 73 of exec_stmt.pgc: Erased unneeded ''.
Last line: added /
Extra: Divided the code block into 2.

Kind Regards,
Yuki Tei


josh-heyer (approved):

josh-heyer:
Looks good @yuki-tei - thanks!

PR #6197, opened 2024-11-01
Update updating_an_rpm_installation.mdx

miyuki-nakashima:
Hi team,

Recently I found that, 
Recently we found that dnf command is different from this page.
https://www.enterprisedb.com/docs/epas/latest/upgrading/minor_upgrade/05_performing_a_minor_version_update_of_an_rpm_installation/
As dnf upgrade and update have a little bit different behavior, I suggest to merge to 'dnf update'.

## What Changed?
dnf upgrade => dnf update

Kind Regards,
Miyuki Nakashima


yuki-tei:
Hi team,

Thanks for your confirmation. Please allow me to point out that,
your other pages uses 'dnf update' instead of 'dnf upgrade' on this page.

We think at least you might need to standardize the expression so
users won't be confused.

Kind Regards,
Yuki Tei

PR #6202, opened 2024-11-04
Docs 1095 pg failover slots v110

piano35-edb:


advocacy_docs/pg_extensions/pg_failover_slots/rel_notes/index.mdx
gvasquezvargas, line 1: Just a nit: 
You could add a `navigation` section in the frontmatter of the page to make the 1.1.0 version appear above the 1.0.0 version in the left navigation panel. (This would be consistent with the way we do it for other products). 

So: 
```
navigation:
  - pg_failover_slots_1.1.0_rel_notes
  - pg_failover_slots_1.0.0_rel_notes
  ```

gvasquezvargas (approved):
Left a suggestion for consistency with other products, otherwise, LGTM!

PR #6207, opened 2024-11-05
Remove BA catch-all redirect - is being too greedy

josh-heyer:
## What Changed?

Was preventing fine-grained redirects from working!


djw-m (approved):
LGTM

PR #6208, opened 2024-11-05
Release: 2024-11-05b

josh-heyer:
## What Changed?

- Remove catch-all redirect for /docs/biganimal/* path until we can ensure it isn't too greedy!


djw-m (approved):
LGTM

PR #6220, opened 2024-11-08
Add gen_slot_name note

djw-m:

## What Changed?

Added small note on slot name generator.

### DIFF
--- a/product_docs/docs/pgd/3.7/pglogical/subscriptions/index.mdx
+++ b/product_docs/docs/pgd/3.7/pglogical/subscriptions/index.mdx
@@ -184,6 +184,11 @@ alternately choose your own replication slot name instead of using
 `gen_slot_name` then pass it as the `slot_name` parameter to
 `create_subscription`.
 
+!!! Note pglogical.gen_slot_name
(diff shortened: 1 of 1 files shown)


DonaldKai (approved):

product_docs/docs/pgd/3.7/pglogical/subscriptions/index.mdx
```diff
@@ -184,6 +184,11 @@ alternately choose your own replication slot name instead of using
 `gen_slot_name` then pass it as the `slot_name` parameter to
 `create_subscription`.
 
+!!! Note pglogical.gen_slot_name
+Names passed to this function are normalized to lower-case only with no symbols. Upper-case and non-underscore symbols are replaced with underscores.
```
gvasquezvargas, line 188: So upper-case letters and symbols are converted to underscores, and lower-case letters are kept as they are? 
Or are upper-case letters converted to lower-case letters, and non-undersocre symbols are replaced by underscores? If it is the first option, maybe something like this would leave no doubt:
```suggestion
This function only accepts lower-case letters and underscores. When you pass names to this function, it converts upper-case text and non-underscore symbols to underscores.
```

PR #6232, opened 2024-11-14
add extension pgaudit

piano35-edb:
add extension pgaudit

### DIFF
--- a/advocacy_docs/pg_extensions/pgaudit/configuring.mdx
+++ b/advocacy_docs/pg_extensions/pgaudit/configuring.mdx
@@ -1,0 +1,24 @@
+---
+title: Configuring pgaudit 
+navTitle: Configuring
+---
+
+1. In `postgresql.conf`, add `pgaudit` to the `shared_preload_libraries` parameter: 
+
+   ```ini
+   shared_preload_libraries = 'pgaudit'
+   ```
+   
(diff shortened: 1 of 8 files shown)


advocacy_docs/pg_extensions/pgaudit/configuring.mdx
```diff
@@ -0,0 +1,24 @@
...
+1. In `postgresql.conf`, add `pgaudit` to the `shared_preload_libraries` parameter: 
+
+   ```ini
+   shared_preload_libraries = 'pgaudit'
+   ```
+   
+   If `shared_preload_libraries` has other extensions, then you can add `pgaudit` to the list. The order doesn't matter. 
+   
+
+2. Restart Postgres.
```
jpe442, line 15: Should you spell out how to do this for those who don't know. I imagine it is just restart with system D, but I am not sure if it depends on the setup or not. Either way it may be an easy way to make it a touch more explicit.

jpe442 (approved):
LGTM with a small note on restarting postgres step

advocacy_docs/pg_extensions/pgaudit/configuring.mdx
piano35-edb: I'll think about how to best accomplish that since so many extensions use the same method.  Thanks.

jpe442: 👍

PR #6258, opened 2024-11-19
MTK: added note on unquoted password for toolkit-properties

gvasquezvargas:
## What Changed?
https://enterprisedb.atlassian.net/browse/DOCS-1111
Originated from Slack convo, see Jira for more details.

### DIFF
--- a/product_docs/docs/migration_toolkit/55/06_building_toolkit.properties_file.mdx
+++ b/product_docs/docs/migration_toolkit/55/06_building_toolkit.properties_file.mdx
@@ -160,15 +160,13 @@ The URL contains the following information:
 
 -   `<database_id>` — The name of the target database.  Set this to the name of the database in your cluster that you want to use as your migration target database.  The name of the default database for your cluster is shown in the **Dbname** field on the **Connect** tab in the Cloud Service portal.  Often a separate database is created for use as the migration target.
 
+-   `sslmode` — Either "require" or "verify-full". See [Recommended settings for SSL mode](/edb-postgres-ai/cloud-service/using_cluster/connect_from_a_client/#recommended-settings-for-ssl-mode). Listed at the end of the **Service URI** value on the **Connect** tab in the Cloud Service portal.
+
 Ensure these values are set in the `toolkit.properties` file:
 
 -   `TARGET_DB_USER` — Specifies the name of a privileged database user. You can copy it from the **User** field on the **Connect** tab in the Cloud Service portal.
 
 -   `TARGET_DB_PASSWORD` — Contains the password of the specified user. Provide the password without any surrounding quotes.
-
-Ensure this value is set in Cloud Service:
-
--   `sslmode` — Either "require" or "verify-full". See [Recommended settings for SSL mode](/edb-postgres-ai/cloud-service/using_cluster/connect_from_a_client/#recommended-settings-for-ssl-mode). Listed at the end of the **Service URI** value on the **Connect** tab in the Cloud Service portal.
 
 ## URL and parameters for Oracle
 
--- a/product_docs/docs/migration_toolkit/55/06_building_toolkit.properties_file.mdx
+++ b/product_docs/docs/migration_toolkit/55/06_building_toolkit.properties_file.mdx
@@ -1,6 +1,6 @@
 ---
 title: "Specifying connection properities"
-
+DeepToC: true
 ---
 
 <div id="building_toolkit_properties_file" class="registered_link"></div>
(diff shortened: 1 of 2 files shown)


product_docs/docs/migration_toolkit/55/06_building_toolkit.properties_file.mdx
```diff
@@ -72,7 +72,7 @@ An Advanced Server URL contains the following information:
 
 -   `{TARGET_DB_USER|SRC_DB_USER}` — Specifies a user with privileges to create each type of object migrated. If migrating data into a table, the specified user might also require insert, truncate, and references privileges for each target table.
 
--   `{TARGET_DB_PASSWORD|SRC_DB_PASSWORD}` — Set to the password of the privileged Advanced Server user.
+-   `{TARGET_DB_PASSWORD|SRC_DB_PASSWORD}` — Set to the password of the privileged Advanced Server user. Do not wrap the password in quotes.
```
gvasquezvargas, line 75: This and the following edits refer to sections in the doc that mention setting the password in the scope of defining a database URL depending on the provider (PG, BigAnimal, Oracle, SQL, etc.).  

For some reason, even though the password is not mentioned in that URL, setting a password is required in all "Defining URL" section of this page. 

1. Do we really need to mention the password in the "Defining URL" sections, even though they are not implicitly included in the URL? 
2. If yes, I am assuming the "do not use quotes" rule applies here as well? 

@zahid75 @mlewandowski-edb

product_docs/docs/migration_toolkit/55/06_building_toolkit.properties_file.mdx
```diff
@@ -113,7 +113,7 @@ The URL contains the following information:
 
 -   `{SRC_DB_USER|TARGET_DB_USER}` — Specify a user with privileges to create each type of object migrated. If migrating data into a table, the specified user might also need insert, truncate, and references privileges for each target table.
 
--   `{SRC_DB_PASSWORD|TARGET_DB_PASSWORD}` — Set to the password of the privileged PostgreSQL user.
+-   `{SRC_DB_PASSWORD|TARGET_DB_PASSWORD}` — Set to the password of the privileged PostgreSQL user. Do not wrap the password in quotes.
```
zahid75, line 116: I would suggest to use "Provide the password without any surrounding quotes." to keep it consistent with earlier modifications you did above.

zahid75 (changes requested):

product_docs/docs/migration_toolkit/55/06_building_toolkit.properties_file.mdx
mlewandowski-edb: I think having the section titles saying "Defining ... URL", which is about the connection URL and then appending the migration toolkit SRC_DB_USER/TARGET_DB_USER and SRC_DB_PASSWORD/TARGET_DB_PASSWORD to the URL parameters discussion is very confusing. Would it make sense to retitled the sections to something like "Defining the URL and configuring other properties for <DB type connections>" and then more clearly separate the USER and PASSWORD property descriptions from the URL property descriptions.

mlewandowski-edb (commented):
See my comment about the confusion of the user and password properties in the defining URL sections.

product_docs/docs/migration_toolkit/55/06_building_toolkit.properties_file.mdx
gvasquezvargas: Sounds good, thanks for your replies. To sum up, on each of the "Defining ... URL" sections, I'll: 

- Change the title from "Defining ... URL" to "Defining the URL and configuring the required properties for ..."
- Separate the SRC_DB_USER/TARGET_DB_USER and SRC_DB_PASSWORD/TARGET_DB_PASSWORD bullet points from the other bullets with an introductory clause like "Ensure these properties are set in the `toolkit.properties` file:"

product_docs/docs/migration_toolkit/55/06_building_toolkit.properties_file.mdx
```diff
@@ -150,19 +154,23 @@ The URL contains the following information:
...
+Ensure these values are set in the `toolkit.properties` file:
 
--   `<database_id>` — The name of the target database.  Set this to the name of the database in your cluster that you want to use as your migration target database.  The name of the default database for your cluster is shown in the **Dbname** field on the **Connect** tab in the BigAnimal portal.  Often a separate database is created for use as the migration target. 
+-   `TARGET_DB_USER` — Specifies the name of a privileged database user. You can copy it from the **User** field on the **Connect** tab in the Cloud Service portal.
 
--   `TARGET_DB_USER` — Specifies the name of a privileged database user. You can copy it from the **User** field on the **Connect** tab in the BigAnimal portal.
+-   `TARGET_DB_PASSWORD` — Contains the password of the specified user. Provide the password without any surrounding quotes.
 
--   `TARGET_DB_PASSWORD` — Contains the password of the specified user. 
+Ensure this value is set in Cloud Service:
```
mlewandowski-edb, line 169: The `sslmode` parameter is part of the connection URL for cloud service connections. I think this should be moved up to the text that describes the connection URL settings.

mlewandowski-edb (commented):
I left a comment suggesting a minor change.

product_docs/docs/migration_toolkit/55/06_building_toolkit.properties_file.mdx
gvasquezvargas: Good catch, I thought it was a Console configuration. Updated ✅

PR #6272, opened 2024-11-21
Add hints on what's potentially generic upgrade instructions

mwanner:
## What Changed?

Add hints (Markdown comment?) to indicate PGD upgrade consideration that likely is generic and not specific to manual upgrades.

NOT TO BE MERGED AS IS. LIKELY JUST ME USING A WEIRD WAY TO COMMENT ON DOCUMENTATION.


mwanner:
I don't recall what this was all about, but can re-open if still needed.

PR #6288, opened 2024-11-25
PGD: Fix link to EDB Postgres Extended Server

ibarwick:
## What Changed?

Replace links to legacy, non-existent portal page with a direct reference to the public documentation, as is done with EPAS.

DOCS-1134


djw-m (approved):
LGTM

(Should have shown up on the link checker though, will investigate)

PR #6309, opened 2024-11-27
Docs/aidb/testlayout

djw-m:
## What Changed?

Alternate layout

NOTE - This was done on a snapshot for the main branch and will be out of sync. Do not proof this branch. It is for feedback on the layout and flattening only.


djw-m:
Applied the process separately to base branch

PR #6315, opened 2024-12-02
Content/docs/cloudservice/dec update

nidhibhammar:
## What Changed?

Added a deprecation note for EDB hosted cloud service


advocacy_docs/edb-postgres-ai/cloud-service/release_notes/index.mdx
```diff
@@ -18,6 +18,10 @@ redirects:
 indexCards: none
 ---
 
+!!!note "Deprecation"
+Please be advised that the EDB Hosted Cloud Service has been deprecated. Support will still be available for current customers. However the related documentation topics removal work is in progress and will be removed shortly. We recommend affected users transition to alternative solutions as soon as possible. Further updates will be provided as the transition progresses.
```
gvasquezvargas, line 22: I think the note is good as it is, just making some suggestions to make it more assertive, but feel free to accept or reject :) 
```suggestion
The EDB Hosted Cloud Service has been deprecated. Support is available for current customers. However, the related documentation topics will be removed shortly. We recommend that affected users transition to an alternative solution immediately. Further updates will be provided as the removal progresses.
``` 
I also think instead of "We recommend that affected users transition to an alternative solution inmediately." we should say something more helpful like: "Our Professional Services team can support your transition into an alternative solution on a case-by-case basis." (But I don't know if support services will actually help, so we'd have to check with PM.

advocacy_docs/edb-postgres-ai/cloud-service/index.mdx
```diff
@@ -16,6 +16,7 @@ redirects:
 - /purl/upm/home/
 - /biganimal/latest/
 - /biganimal/latest/overview/
+displayBanner: "Please be advised that the EDB Hosted Cloud Service has been deprecated. Support will still be available for current customers. However the related documentation topics removal work is in progress and will be removed shortly. We recommend affected users transition to alternative solutions as soon as possible. Further updates will be provided as the transition progresses."
```
gvasquezvargas, line 19: Not sure where this is supposed to appear, or it is only in the background. I cannot see any banners in the preview: 
https://deploy-preview-6315--edb-docs-staging.netlify.app/docs/edb-postgres-ai/cloud-service/ 

Also, I dropped an edited version of the message as suggestion in the comment below.

advocacy_docs/edb-postgres-ai/cloud-service/release_notes/index.mdx
gvasquezvargas, line 1: Should we add the RN for Nov here?

djw-m: Done

advocacy_docs/edb-postgres-ai/cloud-service/index.mdx
djw-m: Fixed issue with displayBanner.

gvasquezvargas (approved):
LGTM!

dbwagoner:
For this statement, "We recommend that affected users transition to an alternative solution immediately." - can you review what Support has in terms of phrasing? It seems that we would want to direct users to the EDB BYOA or other EDB service. This sentence can be removed, also, as Nidhi suggested. We may want to say that BYOA is still available and contact EDB Support with any questions.

PR #6345, opened 2024-12-12
add schema validation for relgen

josh-heyer:
## What Changed?

This may not be the nicest way of handling errors, but it has the distinct advantages of allowing for in-editor warnings when using VSCode, which is nicer than dealing with error messages anyway.

also:

- show warnings / errors in GitHub when running from Actions

- allow running from npm

## Reviewing

Pull locally, then from VSCode terminal run 

```shell
npm i
```

then

```shell
npm run releasenotes:generate-all
```

Should have clickable links to non-compliant files with vaguely relevant line and column numbers. Click them, get reasonable-ish errors shown inline.


djw-m (approved):
With the node version issue isolated, this is

LGTM

PR #6384, opened 2025-01-06
Fix lakehouse nav being out of order

djw-m:

## What Changed?

Small fix to lakehouse nav as it was out of order


jpe442 (approved):
LGTM

PR #6346, opened 2024-12-12
Clarify which values are optional.

sonotley:
Clarify which values are optional.

Explain that PEM will still create agent users even when pem-agent-user is provided.

PR #6385, opened 2025-01-07
Release 2025-01-07a

djw-m:
## What Changed?

#6383 : PG4K: Patch to correct LTS status and remove backslashes
#6384 : Fix lakehouse nav being out of order
#6346 : Clarify which values are optional.


jpe442 (approved):
LGTM

PR #6390, opened 2025-01-08
update firewall information for efm

EFM-Bobby:
This updates the firewall prereqs text to make it clear that:

1. the port that needs to be open is the one used with the bind.address property
2. the admin.port port doesn't need to be opened.


EFM-Bobby:
Note to self: close out https://enterprisedb.atlassian.net/browse/EFM-1715 when this is done.

jpe442 (approved):
LGTM

PR #6391, opened 2025-01-08
DOCS-790

jpe442:
## What Changed?

- fixed instances of using bdr.alter_node_group_config and replaced with bdr.alter_node_group_option where referenced.


djw-m (approved):
LGTM

PR #6393, opened 2025-01-09
Support Postgres 17 version

sergioenterprisedb:
Operator support Postgres 17 version


djw-m (commented):
Again, these changes are sourced from upstream.

djw-m:
version change has been confirmed, patches have been sent upstream to both repos. Applying change here now as it should not be overwritten in future.

djw-m (approved):
LGTM (after confirmation and upstream patching)

PR #6402, opened 2025-01-13
Migration Handbook: EPAS 17 updates

gvasquezvargas:
## What Changed?

https://enterprisedb.atlassian.net/browse/DOCS-1151 

Added some EPAS 17 updated to the Migration Handbook.

### DIFF
--- a/advocacy_docs/migrating/oracle/oracle_epas_comparison/non_relational_data.mdx
+++ b/advocacy_docs/migrating/oracle/oracle_epas_comparison/non_relational_data.mdx
@@ -8,8 +8,7 @@ title: "Comparison of nonrelational data support"
 | Spatial/location/graph     | Yes                          | Yes                           |
 | JSON support               | Yes <br/> Text based         | Yes <br/> Text and high-performance binary based |
 | Key-value store            | NoSQL database               | Yes                           |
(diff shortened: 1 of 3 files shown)


advocacy_docs/migrating/oracle/oracle_epas_comparison/non_relational_data.mdx
```diff
@@ -9,6 +9,7 @@ title: "Comparison of nonrelational data support"
 | JSON support               | Yes <br/> Text based         | Yes <br/> Text and high-performance binary based |
 | Key-value store            | NoSQL database               | Yes                           |
 | Support for XML namespaces, DOM, XQuery, SQL/XML, and XSLT | XML DB                        | No                            |
+| Support for XMLDOM API package | Yes   | Yes  |
```
mlewandowski-edb, line 12: I don't think we need a new separate row in the table for this. Instead, I think we should change the "No" in the row above ("Support for XML namespaces, DOM, XQuery, SQL/XML, and XSLT") to

> Partial
> XML DOM support is provided via the DBMS_XMLDOM package

mlewandowski-edb (commented):
See the comment I added for a suggested change.

mlewandowski-edb (approved):

josh-heyer (approved):

PR #5966, opened 2024-08-20
Replaces Basic Architecture section in PGD Overview page with an enhanced architecture overview

jpe442:
DOCS-824

Replaces Basic Architecture section in PGD Overview page with an enhanced architecture overview.

PR #6405, opened 2025-01-14
Docs 824 arch overview

jpe442:
## What Changed?

New PGD Architecture Overview section to replace the Basic Architecture section we have now. 

Originally started in this [PR](https://github.com/EnterpriseDB/docs/pull/5966), but got a bit too out of sync with the base branch.

Critical feedback from Petr I tried to answer in the Architectural Flexibility section:

"I kind of hate that we put this into HA topic, I much prefer us talking about flexibility, which high availability is just effect/result of.

I.e. flexibility in:

rolling maintenance operations (blue/green)
postgres minor/major upgrades
system upgrades
schema changes
vacuum
automated failover and self-healing (node coming back after failure/disconnect automatically reconnects to cluster and continues working)
selective replication
configurable durability (commit scopes)
multiple write nodes
read scaling with SO nodes
The text under mentions most of this, but it's all talking about HA which limits the scope of things IMO (i.e. it does not leave space for things like commit scopes)."


djw-m (commented):
The content is essentially solid, though I would note that if this appears with 5.7.
a) the CLI commands referenced will need to be updated
b) the roles are expanded in 5.7 with OT default on, and there should be some mention of subscriber only nodes.
c) I think there's a probably scope for more clearly explaining the bdr replication foundation and how that can work on its own, and then explaining the pgd proxy optiminsation for incoming traffic.

PR #6407, opened 2025-01-14
Docs 1162 collapse pgd 5.6 docs into 5.x

djw-m:
## What Changed?

Currently folding the tree down 5.6 -> 5.0 
Added new version display and support for "version" frontmatter
    (TODO: automate setting preciseVersion where there's relgen generated release notes)
Redirects done.


src/templates/doc.js
```diff
@@ -226,6 +227,8 @@ const DocTemplate = ({ data, pageContext }) => {
       ? frontmatter.showInteractiveBadge
       : !!katacodaPanel;
 
+  let preciseVersion = frontmatter.preciseVersion;
```
josh-heyer, line 230: Why not just use `frontmatter.version`?

djw-m: Trying to differentiate between the rounded path versions (which are in the code as version) and a precise specific version from the frontmatter.

jpe442 (approved):
LGTM

PR #6416, opened 2025-01-20
Fix `pgaa.create_storage_location` docs

mildbyte:
## What Changed?

When creating a storage location, all parameters are required in 0.1.x of pgaa, so make them explicit and use the `=>` syntax so that it's obvious which one goes where.


djw-m (approved):
LGTM

djw-m (approved):
LGTM

PR #6424, opened 2025-01-23
MP: terminated DDL assessment due to EPAS-reserved words

gvasquezvargas:
## What Changed?

https://enterprisedb.atlassian.net/browse/MIG-5001 
https://enterprisedb.atlassian.net/browse/DOCS-1178

### DIFF
--- a/product_docs/docs/migration_portal/4/known_issues_notes.mdx
+++ b/product_docs/docs/migration_portal/4/known_issues_notes.mdx
@@ -233,9 +233,9 @@ While using the Oracle default case, you may experience a lower compatibility ra
 
 ### Incomplete Object Definition
 
-While assessing source schemas with Migration Portal, Object Definitions can be incomplete or the DDL assessment is terminated when the source DDL contains reserved words as object names.
+While assessing source schemas with Migration Portal, in rare scenarios, the converted Object Definitions are incomplete when the source DDL contains reserved words.
 
-If you encounter an issue where some part of the Object Definition is missing or the definition gets terminated before reaching the actual end of the DDL, you are possibly using keywords reserved by EDB Postgres Advanced Server. For example, when you use `CONSTRAINT` as a column name and then use it in a `TABLE`, `VIEW` or `MVIEW` definition, the definition is terminated before reaching the end of the DDL.
+If you encounter an issue where some part of the Object Definition is missing or the definition gets terminated before reaching the actual end of the statement, you are possibly using keywords reserved by EDB Postgres Advanced Server. For example, when you use `CONSTRAINT` as a column name in a `TABLE` definition and then reference it in `VIEW` or `MVIEW` definitions, Migration Portal could terminate that DDL statement abruptly and skip to the next definition without successfully assessing or converting it.
 
 To work around this issue, you have these options: 
 
--- a/product_docs/docs/migration_portal/4/known_issues_notes.mdx
+++ b/product_docs/docs/migration_portal/4/known_issues_notes.mdx
@@ -231,6 +231,20 @@ While using the Oracle default case, you may experience a lower compatibility ra
    Character: 58
    ```
 
+### Incomplete Object Definition
+
+While assessing source schemas with Migration Portal, Object Definitions can be incomplete or the DDL assessment is terminated when the source DDL contains reserved words as object names.
+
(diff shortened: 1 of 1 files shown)


product_docs/docs/migration_portal/4/known_issues_notes.mdx
```diff
@@ -231,6 +231,20 @@ While using the Oracle default case, you may experience a lower compatibility ra
    Character: 58
    ```
 
+### Incomplete Object Definition
+
+While assessing source schemas with Migration Portal, Object Definitions can be incomplete or the DDL assessment is terminated when the source DDL contains reserved words as object names.
```
umangjtank, line 236: How about if we keep title like this:
"While assessing source schemas with Migration Portal, in very rare scenario some Object definition can be found incomplete when the source DDL contains reserved words."

product_docs/docs/migration_portal/4/known_issues_notes.mdx
```diff
@@ -231,6 +231,20 @@ While using the Oracle default case, you may experience a lower compatibility ra
...
+
+If you encounter an issue where some part of the Object Definition is missing or the definition gets terminated before reaching the actual end of the DDL, you are possibly using keywords reserved by EDB Postgres Advanced Server. For example, when you use `CONSTRAINT` as a column name and then use it in a `TABLE`, `VIEW` or `MVIEW` definition, the definition is terminated before reaching the end of the DDL.
```
umangjtank, line 238: "For example, when you use 'CONSTRAINT' as a column name in 'TABLE' definition and use its reference in 'VIEW' or 'MVIEW' definition, then there is a chance the definition gets terminated before reaching the end of the DDL."

product_docs/docs/migration_portal/4/known_issues_notes.mdx
gvasquezvargas: How about this: 
```suggestion
While assessing source schemas with Migration Portal, in rare scenarios, the converted Object Definitions are incomplete when the source DDL contains reserved words.
```

product_docs/docs/migration_portal/4/known_issues_notes.mdx
gvasquezvargas: @umangjtank Is the entire DDL file terminated or just that specific definition and then MP moves to the next definition? 

If it is just that specific definition we could say:
```suggestion
If you encounter an issue where some part of the Object Definition is missing or the definition gets terminated before reaching the actual end of the statement, you are possibly using keywords reserved by EDB Postgres Advanced Server. For example, when you use `CONSTRAINT` as a column name in a `TABLE` definition and then reference it in `VIEW` or `MVIEW` definitions, Migration Portal could terminate that statement and skip to the next definition without successfully assessing or converting it.
```

umangjtank: This will not impact the whole 'DDL file' but it will impact only single 'DDL' having such reserved keyword reference.

So the statement I would suggest:
"If you encounter an issue where some part of the Object Definition is missing or the definition gets terminated before reaching the actual end of the DDL statement, you are possibly using keywords reserved by EDB Postgres Advanced Server. For example, when you use `CONSTRAINT` as a column name in a `TABLE` definition and then reference it in `VIEW` or `MVIEW` definitions, Migration Portal could terminate that DDL statement abruptly and skip to the next definition without successfully assessing or converting it."

product_docs/docs/migration_portal/4/known_issues_notes.mdx
```diff
@@ -231,6 +231,20 @@ While using the Oracle default case, you may experience a lower compatibility ra
...
+
+To work around this issue, you have these options: 
+
+- Rename the column name in the Oracle database, then reextract the DDL and upload it to Migration Portal.
+
+or
+
+- Rename the column name in the extracted DDL file, and then upload it to Migration Portal.
+
```
mlewandowski-edb, line 247: I am wondering if another workaround is to simply update the target DDL with a copy of the source DDL for the object and manually repair that particular object directly in the Migration Portal.  I left a similar comment on the Jira ticket ([MIG-5001](https://enterprisedb.atlassian.net/browse/MIG-5001)).

gvasquezvargas: From what I can see in the Jira ticket, @umangjtank replied that this suggestion, unfortunately, won't fix the issue. 
I have corrected the other instances and think we are good to go. Can I get an approval, please?

umangjtank (approved):
Looks good.

djw-m (approved):
LGTM (assuming the solution is good)

PR #6436, opened 2025-01-29
DMS/Lakehouse Sync: removal notice

gvasquezvargas:
## What Changed?

As of 28.01.2025 Cloud Service no longer offers Data Migration Service and Lakehouse Sync capabilities. 

The Data Migration Service will only be available in HCP deployments. 
For the Lakehouse Sync, it is not yet clear if we'll eventually move it to the HCP or if there will be no support at all. 
Hence the differentiated Notice banners.  

https://enterprisedb.atlassian.net/browse/DOCS-1182


jperozoa (approved):

nidhibhammar (approved):
LGTM

PR #6458, opened 2025-01-31
Edits to MP: terminated DDL assessment due to EPAS-reserved words #6424

ebgitelman:
I didn't see any cases where object definition was capitalized, but let me know if it should be for some reason.


gvasquezvargas (approved):
LGTM!

PR #6468, opened 2025-02-05
EPAS: removing 11 and 12 templates

gvasquezvargas:
## What Changed?

- Removed EPAS 11 and 12 from install templates.
- PostgreSQL for Ubuntu 20 was missing, but I don't think there is a reason for it to be missing.


djw-m (approved):
LGTM

PR #6480, opened 2025-02-11
DOCS-1195 Failover slots 

djw-m:
## What Changed?

Added page and updated references.


product_docs/docs/pgd/5/logical-slot-failover.mdx
```diff
@@ -0,0 +1,94 @@
...
+
+## Logical Slot Failover Support with Postgres 17+
+
+For Postgres 17 and later, support for failover was added to allow standbys to be resumed, through an option in `pg_create_logical_replication_slot` named `failover`. This new setting requires that, no matter what the setting of `failover_slot_scope`, you must also set `failover` to `true`.
+
+```sql
+SELECT pg_create_logical_replication_slot('myslot',
+                                          'test_decoding',
+                                          failover=>true);
+```
```
pavanvd, line 82: We also support creation of failover slots via CREATE_REPLICATION_SLOT command on a replication connection.

product_docs/docs/pgd/5/logical-slot-failover.mdx
```diff
@@ -0,0 +1,94 @@
...
+This is a PGD 5.7 and later feature. It is not supported on earlier versions of PGD.
+!!!
+
+## Background
+
+Earlier versions of PGD have allowed the creation of logical replication slots on nodes that can provide a feed of the logical changes happening to the data within the database. These logical replication slots have been local to the node and not replicated. Apart from only replicating changes on the particular node, this has presented challenges when faced with node failover within the cluster. In that scenario, a consumer of the logical replication off a node which fails has no replica of the slot on another node to continue consuming from.
+
+While solutions to this can be engineered using a subscriber-only node as an intermediary, it does significantly raise the cost of logical replication.
+
+## Logical Slot Failover Support
```
pavanvd, line 17: I think we should also discuss the implications of consumer trying to start decoding from multiple nodes at the same time. PGD prevents that by ensuring that a Logical Failover Slot is only decoded on one node at a time. If the consumer connects to another node (or another instance of the consumer is started), then previous connection will be closed by PGD. This requires functional Raft on both the nodes though. So in case of split brain situations, PGD won't be able to guarantee that.

product_docs/docs/pgd/5/logical-slot-failover.mdx
```diff
@@ -0,0 +1,94 @@
...
+Once enabled you can use:
+
+```sql
+SELECT pg_create_logical_replication_slot('myslot',
+                                          'test_decoding');
+```
+
+To create a new globally replicated slot.
+
+Note that logical replication slots created before the option is set to `global` will not be replicated. Only new slots will be replicated.
```
pavanvd, line 72: `bdr.failover_replication_slots` catalog stores the information about all the Logical Failover Slots. User can query that catalog to check if a slot is created with failover support or not.
```
CREATE TABLE bdr.failover_replication_slots
(
    slot_name name NOT NULL PRIMARY KEY,
    slot_id oid NOT NULL,
    node_group_id oid NOT NULL,
    plugin name NOT NULL,
    twophase boolean NOT NULL,
    active_node oid,
    active_pid int,
    prev_node oid,
    UNIQUE(slot_id)
) WITH (user_catalog_table=true);
```

product_docs/docs/pgd/5/cdc-failover.mdx
```diff
@@ -0,0 +1,105 @@
...
+To ensure you are setting the correct, top-level group’s option.
+
+Once enabled you can use:
+
+```sql
+SELECT pg_create_logical_replication_slot('myslot',
+                                          'test_decoding');
+```
+
+To create a new globally replicated slot.
```
pavanvd, line 77: Should we add a note about dropping the slot? It can be done via PG function or DROP_REPLICATION_SLOT command

product_docs/docs/pgd/5/cdc-failover.mdx
```diff
@@ -0,0 +1,105 @@
...
+                                          'test_decoding',
+                                          failover=>true);
+```
+
+## Limitations
+
+The CDC Failover Slot support comes with certain limitations. 
+
+* CDC Failover slot support requires the latest versions of PGD (5.7+) and the latest minor releases of Postgres-Extended or EPAS (available Feb 2025\).
+* CDC Failover support is a global option and cannot be set on a per-slot basis.
```
pavanvd, line 100: Though you can influence the behaviour by changing `failover_slot_scope` in between. Note that `failover_slot_scope` option is only consulted when the slot is created. Changing it later does not influence the behaviour of already created slots

djw-m:
```suggestion
* CDC Failover support is a global option and cannot be set on a per-slot basis. It is possible though, because changing the enabled status of CDC Failover does not affect previously provisioned slots, to enable it (set to `global`), create a replicated slot, then disable it (set to `local`), to create a singular replicated slot.
```

product_docs/docs/pgd/5/cdc-failover.mdx
djw-m: Feels unnecessary; it behaves like a slot - if the behaviour was different, it may be worth mentioning?

pavanvd: Ok, makes sense

product_docs/docs/pgd/5/cdc-failover.mdx
pavanvd: Sounds good

pavanvd (approved):
CDC Failover Slot section LGTM

PR #6483, opened 2025-02-13
PDF rendering misc bugs

josh-heyer:
## What Changed?

- Embed SVGs (will not render in wkhtmltopdf otherwise). Also reference other images directly instead of embedding.
- Discourage page breaks between code block and output block
- Get rid of stray padding on the first line of code blocks
- Get rid of excess whitespace after headings


djw-m (approved):
LGTM

PR #6486, opened 2025-02-13
PostgreSQL - Locale update

nidhibhammar:
## What Changed?

Updated the step for Locale selection while installing on Windows as per [DBP-1239](https://enterprisedb.atlassian.net/browse/DBP-1239)


gvasquezvargas (approved):
LGTM!

PR #6504, opened 2025-02-17
EPRS 7.11: logical replication slots for HA and failover

gvasquezvargas:
## What Changed?

https://enterprisedb.atlassian.net/browse/XDB-2345
https://enterprisedb.atlassian.net/browse/DOCS-1209

Hi all! I used the text proposed by the team, but made some additional alterations: 

- The page was getting cluttered and long, so I divided the content into thematic headings (naming, number, HA and failover, other configs) 
- I called the new section "high availability and failover" as I believe this describes the use case. 
- I am proposing a new diagram, with a red x on the primary database, and a link from standby to rep servers, so the failover use case is more evident. 
- I added an intro paragraph and reworded the text into active voice. 
- I formatted the list as bulleted list because I believe the items are guidelines more than steps. 
- I turned the last item into a note, as it applies to a specific sub-use-case. 

Merge tentatively on Feb 21st.


zahid75:
@gvasquezvargas Regarding the use of X in the new diagram, I would suggest using two versions of the diagram:
Before failover
===========
First without X where primary is active and replicating to EPRS, "Logical Replication" from standby to be removed

Post-failover
==========
2nd where primary is down with X, the "Logical Replication" arrow originating from failed primary should also be greyed/marked X, and "Logical Replication" to be introduced originating from newly-promoted primary

zahid75 (changes requested):
Please see my comments related with the use of two diagrams.

product_docs/docs/eprs/7/images/eprs_logical_rep_failover_600w.svg
mlewandowski-edb, line 1: I suggest the following changes in the After diagram

- Update the label for the original primary diagram should be updated to be something like “Original Publication Database Primary”
- Remove the “Logical Replication” arrow between the original Publication Database Primary to the Replication Server

mlewandowski-edb (commented):
Suggested a change to the After Replication diagram

product_docs/docs/eprs/7/images/eprs_logical_rep_failover_600w.svg
mlewandowski-edb: For the second point, an alternative to removing the arrow between the original primary and Replication Server, would be to put an X through it like you did on the original primary

zahid75 (changes requested):
The requested changes look fine. However I just noticed (something that I should have pointed out earlier) that Replication Server is represented by Database specific diagram that should be replaced with a process specific diagram (round rectangle or circle).

zahid75 (approved):
The updated diagram incorporates the suggested changes.

djw-m (approved):
LGTM (But why so many sections here talking about Postgres 9.x...?)

gvasquezvargas:
> LGTM (But why so many sections here talking about Postgres 9.x...?)

@djw-m stop being in my head, I already logged a Jira for future Gaby to handle that 😆 
https://enterprisedb.atlassian.net/browse/DOCS-1212

product_docs/docs/eprs/7/02_overview/02_replication_concepts_and_definitions/10_sync_replication_log_based/02_logical_replication_slots.mdx
```diff
@@ -2,21 +2,61 @@
...
+## High availability and failover
+
+EPRS supports logical replication slot synchronization between primary and standby databases on PostgreSQL and EDB Postgres Advanced Server versions 17.x and later. 
+
+When enabled, the logical replication slot is replicated to the standby. If the primary fails, the standby can take over, allowing for replication to continue in failover scenarios as the slot is already there.
+
+### Before failover 
+
+![Logical replication slots with publication database standby before failover](/eprs/latest/images/eprs_logical_replication_600w.svg)
+
```
djw-m, line 34:
```suggestion

---

```

PR #6511, opened 2025-02-18
Use precise version (when available) for PDF download filename

josh-heyer:
## What Changed?

The PDF generator already did this, resulting in a 404 when precise version was used - e.g. download link points to `pgd_v5_documentation.pdf` but actual filename is `pgd_v5.6.1_documentation.pdf` (reported by @pnkrepps in Slack)

Also: use path version for identifying "active" version in selector - don't round precise version, as this would fail to produce the correct results when the path version already included a minor or patch version.

Added precise version for the other published PGD versions, primarily as a more thorough testbed for this stuff, but also... a bit of consistency is nice!

### DIFF
--- a/product_docs/docs/pgd/3.6/index.mdx
+++ b/product_docs/docs/pgd/3.6/index.mdx
@@ -4,6 +4,8 @@ navigation:
 - release-notes
 - rel_notes-pglogical
 hidePDF: true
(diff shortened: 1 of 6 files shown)


jpe442 (approved):
LGTM

PR #6519, opened 2025-02-19
Fix package name for dnf

martinmarques:
The rpm package name for edb_wait_states has underscores instead of dashes


gvasquezvargas (approved):
LGTM! 
@martinmarques shall we merge and publish this right away or wait for the 1.8 release?

PR #6544, opened 2025-02-25
AI: clarify model dimensions

timwaizenegger:
## What Changed?

I found a mistake/ambiguity in the docs for model dimensions at https://www.enterprisedb.com/docs/edb-postgres-ai/ai-accelerator/models/openai-api-compatibility/#configuration

- this implies that the setting `"dimensions":2000` will limit how many dimensions we get from the model; which isn't the case (i.e. `So we pass a dimension value of 2000 in the configuration, to get the maximum dimensions available with pgvector.`)
- text mentions that pgvector only support 2000 dimensions but that only applies to the standard index type. We don't use indexing right now so it doesn't apply to us at all. Even with indexing, you'd have to use the "halfvec" types to get more dimensions


djw-m (approved):
LGTM

PR #6564, opened 2025-02-27
add index for sections

piano35-edb:
index for section progression


josh-heyer (approved):

josh-heyer:
Note that you can also set `indexCards: simple` in the frontmatter header to automatically generate an index like this

nidhibhammar (approved):
LGTM

PR #6587, opened 2025-03-05
Sequence range in node2

lucianobotti:
There is a typo in the range used for NODE2. 

## What Changed?
The correct value should be the one in the example that is coming in the next paragraph (as fixed here):

"Node Node2 is using range from 2000334 to 4000333."


djw-m (approved):
LGTM

PR #6594, opened 2025-03-07
Efm v5.0 changes

EFM-Bobby:
Changes for EFM 5.0

See story for status on which changes are in this PR already and which are still waiting:
https://enterprisedb.atlassian.net/browse/EFM-1730
    
This first commit makes general version bumps across the docs without any new content yet. A placeholder release notes file has been added; am pushing now to get some help with figuring out e release notes ordering and various "version 4" bits in the generated pages that I don't know how to find.

### DIFF
--- a/product_docs/docs/efm/4/efm_quick_start/index.mdx
+++ b/product_docs/docs/efm/4/efm_quick_start/index.mdx
@@ -31,7 +31,7 @@ Start the configuration process on a primary or standby node. Then, copy the con
 1. Create working configuration files. Copy the provided sample files to create Failover Manager configuration files, and correct the ownership and version number if you are installing a different version:
 
    ```shell
-   cd /etc/edb/efm-4.10
+   cd /etc/edb/efm-5.0
 
    cp efm.properties.in efm.properties
 
@@ -45,7 +45,7 @@ Start the configuration process on a primary or standby node. Then, copy the con
 1. Create an [encrypted password](/efm/latest/04_configuring_efm/02_encrypting_database_password/) needed for the properties file:
 
    ```shell
-   /usr/edb/efm-4.10/bin/efm encrypt efm
+   /usr/edb/efm-5.0/bin/efm encrypt efm
    ```
 
    Follow the onscreen instructions to produce the encrypted version of your database password.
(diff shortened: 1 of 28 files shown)


product_docs/docs/efm/4/efm_quick_start/index.mdx
```diff
@@ -83,38 +83,38 @@ Start the configuration process on a primary or standby node. Then, copy the con
...
    ```
 
 1. Start the agent on the other nodes. Run the `efm cluster-status efm` command on any node to see the cluster status.
 
-  If any agent fails to start, see the startup log for information about what went wrong:
+  If any agent fails to start, see the systemctl status for information about what went wrong. If using Failover Manager version 4.x, see the startup log instead:
 
    ```shell
-   cat /var/log/efm-4.10/startup-efm.log
+   cat /var/log/efm-5.0/startup-efm.log
```
josh-heyer, line 109: shouldn't this be `cat /var/log/efm-4.<x>/startup-efm.log` ?

EFM-Bobby: Yep, good catch. I haven't swept everything yet since there's a lot to do. I hopefully would have caught it. :)

product_docs/docs/efm/5/04_configuring_efm/01_cluster_properties.mdx
```diff
@@ -121,7 +122,7 @@ Use the properties in the `efm.properties` file to specify connection, administr
...
 | [script.load.balancer.attach](#script_load_balancer)          |       |       |                                       | Example: script.load.balancer.attach= /&lt;path>/&lt;attach_script> %h %t                                                                                      |
 | [script.load.balancer.detach](#script_load_balancer)          |       |       |                                       | Example: script.load.balancer.detach= /&lt;path>/&lt;detach_script> %h %t                                                                                      |
-| [detach.on.agent.failure](#detach_on_agent_failure)           |       |       | true                                  | Set to false if you want to keep a running primary database attached.  **Available in Failover Manager 4.2 and later.**                                                                                   |
+| [detach.on.agent.failure](#detach_on_agent_failure)           |       |       | false                                 |                                                                                                                                                                |
```
EFM-Bobby, line 125: Because we now have a sane value for this property, I don't think we need any explanatory info here. It's a property that no one should use except in very specific circumstances. If even then....

product_docs/docs/efm/5/efm_rel_notes/29_efm_50_rel_notes.mdx
```diff
@@ -0,0 +1,33 @@
...
+| Enhancement | Removed local database checks after an agent cannot connect that produced confusing logging. |
+| Enhancement | Added more information in some cases when an agent cannot connect to a database. |
+| Enhancement | Failover Manager was upgraded to use JGroups version 5.4.4. |
+| Enhancement | Failover Manager was upgraded to use PostgreSQL JDBC Driver version 42.7.3. |
+| Enhancement | Failover Manager was upgraded to use Log4J version 2.24.3. |
+| Enhancement | Failover Manager was upgraded to use Jakarta Mail version 2.0.1. |
+| Enhancement | Failover Manager was upgraded to use Json-Simple version 1.1.1. |
+| Bug Fix | Fixed agent shutdown code that could cause a 'systemctl stop' call to finish before the agent completes shutdown, resulting in the agent process not exiting gracefully. [Support ticket: #37270] |
+| Bug Fix | Handle a case where an agent on a failed/rebooted standby node attempts to join the cluster before the previous agent from the same host has timed out. [Support ticket: #43162] |
+| Bug Fix | Removed 'efm upgrade-conf' support for very old Failover Manager version passwords. In rare cases, this support was causing upgrade failures. |
```
EFM-Bobby, line 30: I looked it up and "very old" means before version 4.0. If you think it would be better to call out the version here, that might look better than what I have.

product_docs/docs/efm/5/04_configuring_efm/01_cluster_properties.mdx
gvasquezvargas: I still think we should say something here as the empty field might generate more questions. 

We could either say this property is not to be used in production. 

Or say this is an outdated value replaced by [sane value here].

Or we should remove the line completely if we really don't want users using it.

EFM-Bobby: I've added some text here. Hopefully it makes sense.

theadamwright (approved):

PR #6609, opened 2025-03-14
Update 03_sql_profiler.mdx

yuki-tei:
Hi team,

## What Changed?
Since currently each extension details are on another manual, it is better to link it so users can access it easily.

Kind Regards,
Yuki Tei


nidhibhammar (approved):
LGTM

PR #6612, opened 2025-03-18
PGD SiX draft branch for feedback

djw-m:
## What Changed?

Created a new PGD branch (Suggest isolating it so users can't roll back time easily).
Moved existing docs temporarily to the root.

Built out a basic framework for the new content which could be easily bifurcated into solo manuals.


djw-m:
Deleting branch as unreviewed and DOCS-1249 is lead strategy now

PR #6629, opened 2025-03-20
Update primitives.mdx

bilge-ince:
The parameter "my_bert_model" refers to a sentence transformer model, which is incapable of generating embeddings from images or summarising and translating text. I believe this information was mistakenly copied from another source. I have corrected it using the default models supported by AIDB.

## What Changed?

- `aidb.encode_image_batch` function used my_bert_model and was replaced with a clip model capable of generating embeddings from images.

- `aidb.decode_text_batch` function used my_bert_model and was replaced with the t5 model, which is capable of performing summarising and translating tasks.


djw-m (approved):
LGTM

PR #6647, opened 2025-03-26
Update usage.mdx

bilge-ince:
## What Changed?
Replace the parameter of `source_table_name` with `source_table`. It's correct in the page of `Pipelines retrievers reference` but it's been missed in this page.


djw-m (approved):
LGTM

PR #6592, opened 2025-03-07
DOCS-1265 PGAIDB 2.2.1

djw-m:
## What Changed?

Release notes stubbed
Prepaper docs added
NIM updates applied
New Using with section created

PR #6651, opened 2025-03-28
Edits to DOCS-1265 PGAIDB 2.2.1 #6592

ebgitelman:


ebgitelman:
I didn't edit the release notes but I'd like to. We should move away from first person there and also be consistent with how I've capitalized preparers and retrievers in the text (among other things).

djw-m (approved):
LGTM

PR #6673, opened 2025-04-03
quick and dirty product name replacement

josh-heyer:
## What Changed?

Allows expressions like `{{name.ln}}` or `{{name(epas).abbr}}` or `{{version.minor}}` to be used in Markdown, title, description.

Demo: https://deploy-preview-6673--edb-docs-staging.netlify.app/docs/playground/1/01_examples/expression-replacement/


djw-m (approved):
Looks pretty clean for quick and dirty.

This would/would not apply to content in advocacy_docs? I'm assuming not which is more about us needing to do the docsingularity on the tree.

I would back out the PGD 5.7 docs change before merging though.

But generally, I approve.

PR #6676, opened 2025-04-03
EPRS: replaces em dashes in code snippets with hyphen in commands

gvasquezvargas:
## What Changed?

There is a bug where in some documented command flags/arguments, they are introduced by an em dash instead of a hyphen. This is causing customers to get the `Error: Invalid combination of arguments.` when copy-pasting from the docs. 

https://enterprisedb.atlassian.net/browse/XDB-2369

I did a global search in the */eprs/* documentation and found more instances.

PS: The difference is not that visible here in GH, but VSCode displayed it better: 

Screenshot 2025-04-03 at 18 15 40


jpe442 (approved):
So the snippets had em-dashes in them? Weird. LGTM

zahid75 (approved):
Looks good.

PR #6688, opened 2025-04-09
Update installing.mdx

yuki-tei:
Hi team,
Greetings.

## What Changed?
SQL Profiler Extension might be supported since version 13, not 12.
According to your archive of EPAS documentation v12, the instruction indicates to invoke scripts instead of create extension.

Kind Regards,
Yuki Tei


nidhibhammar (approved):
LGTM

PR #6700, opened 2025-04-16
Update 07_using_efm_utility.mdx

yuki-tei:
Hi team,

## What Changed?
Since "Superuser privilege" usually indicates the "Superuser"  Attribute for roles, I think it is better to explicitly say "as a root" user, so this would avoid any potential misunderstanding.

### DIFF
--- a/product_docs/docs/efm/5/07_using_efm_utility.mdx
+++ b/product_docs/docs/efm/5/07_using_efm_utility.mdx
@@ -116,7 +116,7 @@ If the `-prompt` option is specified, the command will output the steps it will 
 - `db.port` specifies the port used to access the primary database
 - `application.name` specifies the `application_name` to use (if set)
 
-This command must be invoked with superuser privileges in order to perform database and Failover Manager operations.
+This command must be invoked as root user in order to perform database and Failover Manager operations.
 
 !!!Note
     This command was introduced in Failover Manager 5.0. More options will be added in future releases.


nidhibhammar:
@EFM-Bobby, please review and let us know what you think?

EFM-Bobby:
We use "superuser privileges" in other parts of the doc (including this page for the 'efm upgrade-conf'), so if we make the change it should be consistent. Can't other users also have the privileges to run these commands with sudo besides root?

yuki-tei:
Hi,
Greetings. Just for your reference of below:

> Can't other users also have the privileges to run these commands with sudo besides root?
=> I tried this before raising this pull request. Since create-standby option is new and require using pg_basebackup instead of /usr/edb/efm-5.0/bin/efm_db_functions, thus the default sudoers.d setting is not sufficient for this alternative.
We must add this if wee need to sudo.

efm    ALL=(enterprisedb)  NOPASSWD:   /usr/edb/as17/bin/*
efm    ALL=(enterprisedb)  NOPASSWD:   /usr/bin/rm -rf /home/enterprisedb/data

Kind Regards,
Yuki Tei

EFM-Bobby:
> Hi, Greetings. Just for your reference of below:
> 
> > Can't other users also have the privileges to run these commands with sudo besides root?
> > => I tried this before raising this pull request. Since create-standby option is new and require using pg_basebackup instead of /usr/edb/efm-5.0/bin/efm_db_functions, thus the default sudoers.d setting is not sufficient for this alternative.
> > We must add this if wee need to sudo.
> 
> efm ALL=(enterprisedb) NOPASSWD: /usr/edb/as17/bin/* efm ALL=(enterprisedb) NOPASSWD: /usr/bin/rm -rf /home/enterprisedb/data
> 
> Kind Regards, Yuki Tei

But that's not related to whether the term "superuser privileges" or "as root user" is used in the text, is it?

EFM-Bobby:
In the current documentation workspace, I don't see any instances of "invoked as root" but I see nine instances of "invoked with superuser" (in the efm and epas products). I would prefer the documentation stay consistent unless there is some strong reason to change this.

EFM-Bobby:
Here is the PR that added the current text. Note that this change came from a customer request and was approved by the docs team: https://github.com/EnterpriseDB/docs/pull/5256

So far I can't find any other information about this or who the customer was.

yuki-tei:
Hi.

Nah, I just answered your question. So it is currently not available to run these commands with sudo besides root, at least, not without extra setups which is not described in the document.

> > Hi, Greetings. Just for your reference of below:
> > > Can't other users also have the privileges to run these commands with sudo besides root?
> > > => I tried this before raising this pull request. Since create-standby option is new and require using pg_basebackup instead of /usr/edb/efm-5.0/bin/efm_db_functions, thus the default sudoers.d setting is not sufficient for this alternative.
> > > We must add this if wee need to sudo.
> > 
> > 
> > efm ALL=(enterprisedb) NOPASSWD: /usr/edb/as17/bin/* efm ALL=(enterprisedb) NOPASSWD: /usr/bin/rm -rf /home/enterprisedb/data
> > Kind Regards, Yuki Tei
> 
> But that's not related to whether the term "superuser privileges" or "as root user" is used in the text, is it?

Kind Regards,
Yuki tei

yuki-tei:
Hi,

This is just a precaution for those are not used to reading technical documents or Database newbies.
If the document's term is not the general one, it will be easier to trapped by confusion or overlooking, whether the real context is correct or not. Especially for those who doesn't use English as the 1st language, if they use PostgreSQL daily, "superuser" for "roles" will come up first in their minds. We all not want this kind or unconscious term picking and all confusions might happen, right?

Please think of change all 9 parts to "as root user". This is most explicit because it is the general term and familiar for Linux users. 

> In the current documentation workspace, I don't see any instances of "invoked as root" but I see nine instances of "invoked with superuser" (in the efm and epas products). I would prefer the documentation stay consistent unless there is some strong reason to change this.
> 
> I also don't understand what this sentence means: 'Since "Superuser privilege" usually indicates the "Superuser" Attribute for roles...' We're talking about running a linux command.

Kind Regards,
Yuki Tei

nidhibhammar:
@EFM-Bobby, do we have any action item here?

EFM-Bobby:
> @EFM-Bobby, do we have any action item here?

I don't really know. This isn't an efm question; it's about whether the documentation should use "as root user" instead of "superuser privileges." I just want things to be consistent, not just across the efm docs but across all of the product documentation.

FWIW I don't think "as root user" is technically correct, as you can do things with superuser privileges (e.g. install something) without being root. It's unfortunate that the database and operating system both have the term "superuser" but we can't change that. As written, I do not approve of this PR that changes one instance of the wording.

My suggestion is to close this PR, and @yuki-tei can start a different conversation with the docs team about this terminology. A quick search shows that "superuser privileges" is used ~100 times across multiple products.

nidhibhammar:
Thanks, @EFM-Bobby. I am closing this PR. @yuki-tei we can have separate discussion on this if needed.

PR #6705, opened 2025-04-16
DOCS-1432-Update-Compatibility-Table-2

jpe442:
Updated the compatibility table to remove compats for pre v14 PG and updated upgrades compatibility page to say that Upgrading from 4.x to 6.x is possible.


djw-m (approved):
LGTM

PR #6713, opened 2025-04-21
Update using.mdx

yuki-tei:
Hi team,

## What Changed?
Changed some argument values because part of them cannot be implicitly type casted, which leads to "No function matches the given name and argument types" error.

Your original SQL sample is:
SELECT sp_activate(tracefortestuser,0011,134,25,10000,30);

tracefortestuser => 'tracefortestuser' : Cannot be without single-quotation or will be see as unknown identifier.
0011 => '0011': Cannot be casted to oidvector as long as it is integer.
134 => '134': Cannot be casted to oidvector as long as it is integer.
30 => '30 minutes' : Cannot be casted to interval as long as it is integer. If you means 30 minutes then it should be as string like this.

Kind Regards,
Yuki Tei

### DIFF
--- a/advocacy_docs/pg_extensions/sqlprofiler/using.mdx
+++ b/advocacy_docs/pg_extensions/sqlprofiler/using.mdx
@@ -59,7 +59,7 @@ sp_activate(
 Start a new trace named 'tracefortestuser', for user with OID 0011, on database with OID 134, with a maximum trace log size of 25MB, for queries that run at least 10 seconds, up to 30 minutes:
 
   ```sql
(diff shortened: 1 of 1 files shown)


piano35-edb:
Pending confirmation of the correction from sqlprofiler dev.

S-SrinathReddy:
@piano35-edb  LGTM but to be clearer and safer for the last argument we can set as " interval '30 minutes' ".
ex:- select * from sp_activate('ironman', '10', '15489', 0, -1, interval '30 minutes');

piano35-edb:
Added fix to branch docs-small-fixes-071825

PR #6714, opened 2025-04-22
PGD: remove reference to "pause_in_standby"

ibarwick:
## What Changed?

This deprecated parameter was removed from function bdr.join_node_group() in PGD 6.

DOCS-1449.


djw-m (commented):
Misses the matching edit to the reference entry.

I will tidy up.

PR #6719, opened 2025-04-22
Clarify a safe scenario for bdr.drop_node()

eatonphil:


product_docs/docs/pgd/5.7/reference/functions-internal.mdx
```diff
@@ -170,7 +170,9 @@ database. The node can be either:
 
 
 !!! Important When to use `bdr.drop_node()`
-Do not use `bdr.drop_node()` to drop node metadata and reuse node names. PGD can reuse existing node names providing the node name belongs to a node in a `PARTED` state. Use `bdr.part_node()` to remove the original node and place it in a `PARTED` state.
+Do not use `bdr.drop_node()` just to drop node metadata and reuse node names. PGD can reuse existing node names providing the node name belongs to a node in a `PARTED` state. Use `bdr.part_node()` to remove the original node and place it in a `PARTED` state.
+
+A scenario you might need to use `bdr.drop_node()` is if a node gets stuck in `PART_START` state. Running `bdr.drop_node()` will force the part to continue which means that any data on the node being dropped will be lost. If you have already re-imaged the node being dropped (you've already deleted the data on that node) then there is no additional risk to running `bdr.drop_node()`.
```
vaijayanti-b, line 176: My understanding is that if a node is stuck in PART_START or so, doing a force_part will cause it to move to PARTED skipping intermediate states.
drop_node with force will force a drop even if a node is not parted, but perhaps used when there is a bug and force_part is not working.
Mostly even drop_node is also not needed. Rejoining the same node even when it is in PARTED state works very well as the documentation rightly says, unless sometimes there is some leftover state not cleaned up.

PJMODOS (approved):
I think this is fine for 5, obviously 6 will change this

product_docs/docs/pgd/5.7/reference/functions-internal.mdx
```diff
@@ -161,16 +161,25 @@ bdr.difference_fix_xact_set_avoid_conflict()
 
 Drops a node's metadata.
 
+After a node has been `PARTED` its info remains around. For example, the node will remain in the `bdr.node_sumary` results, marked as `PARTED`, until the node is dropped.
```
djw-m, line 164: "its info remains around" is kinda vague.

When a node leaves the cluster by being `PARTED`, the node's metadata remains on the other nodes enabling the nodes to handle any synchronization issues. This state is maintained until the node is dropped. Once a node is `PARTED`, its node name is available to be reused.

product_docs/docs/pgd/5.7/reference/functions-internal.mdx
```diff
@@ -161,16 +161,25 @@ bdr.difference_fix_xact_set_avoid_conflict()
...
+
+Calling `bdr.drop_node('some node', force := true)` can be necessary
+and appropriate when a node becomes stuck while parting. Note that it
+skips past syncing any data out of the node being dropped, if there is
+any data on that node that still needs to be synced out. If a node
+stuck parting has already been reimaged or deleted, there is no harm
+in calling `bdr.drop_node` on it. Note that this must be called for
+this stuck node on all nodes in the cluster so they all have a
+consistent view that the node has been dropped.
```
djw-m, line 173: The force parameter can be used with this command if a node becomes stuck while parting. Forcing the metadata drop means that the node being dropped will not get a chance to synchronize its data. If the node has already been reimaged or deleted, dropping it will not cause harm.  If you are removing a stuck node, you will have to drop the node on all the nodes in the cluster to ensure the cluster's view of nodes is consistent across the cluster.

product_docs/docs/pgd/5.7/reference/functions-internal.mdx
```diff
@@ -161,16 +161,25 @@ bdr.difference_fix_xact_set_avoid_conflict()
...
 database. The node can be either:
 
 -   The local node, in which case it removes all the node metadata, including information about remote nodes.
 -   A remote node, in which case it removes only metadata for that specific node.
 
-
-
 !!! Important When to use `bdr.drop_node()`
-Do not use `bdr.drop_node()` to drop node metadata and reuse node names. PGD can reuse existing node names providing the node name belongs to a node in a `PARTED` state. Use `bdr.part_node()` to remove the original node and place it in a `PARTED` state.
+It is not necessary to use `bdr.drop_node()` to drop node metadata just for the sake of reusing node names. PGD can reuse existing node names providing the node name belongs to a node in a `PARTED` state. Use `bdr.part_node()` to remove the original node and place it in a `PARTED` state.
```
djw-m, line 182:
```suggestion
It is not necessary to use `bdr.drop_node()` to drop node metadata just to reuse node names. PGD 5 and later can reuse existing node names as long as the node name in question belongs to a node in a `PARTED` state. Instead of dropping the node, use `bdr.part_node()` to remove the original node and place it in a `PARTED`.
```

eatonphil: Good suggestions!

product_docs/docs/pgd/5.7/reference/functions-internal.mdx
djw-m:
```suggestion
After a node has been `PARTED` its metadate remains present in the cluster's nodes. For example, the node will remain in the `bdr.node_sumary` results, marked as `PARTED`, until the node is dropped.
```

djw-m (approved):
LGTM

PR #6720, opened 2025-04-23
[DOCS-1451] Creating landing page and content seed for warehousepg and support

djw-m:
## What Changed?

Added pages as draft


BillSmith-EDB:
I took a look at the initial draft of the docs and have the following comments:

- Installing on RHEL/CentOS, there's this installation command: sudo dnf install -y warehouse-pg-6 warehouse-pg-clients. This is applicable only for RH 8/9.  For RH7, it's sudo yum install instead of sudo def install. 
- We need to have something that tells customers if they're installing version 6, it's "warehouse-pg-6".  For v7, it's "warehouse-pg-7".
- v6 is only available on RH7/8
- v7 is only available on RH8/9
- We should add whpg-backup to the list of packages are to be installed.

I think this was the only issue that I saw.

BillSmith-EDB (approved):
The changes made after my feedback looks good. I don't see any issues.

PR #6724, opened 2025-04-25
Update updating_an_rpm_installation.mdx

yuki-tei:
Hi team,

## What Changed?
As long as your repository repos1.0 is upgraded to repos1.0 and is no longer available, the new repository name should be 'enterprisedb-enterprise'. Also, your gpg key is now managed by curl online script, not by local key. so you don't need to modify your credential in your /etc/yum.repos.d/ configuration file.

Kind Regards,
Yuki Tei


tusharahuja (approved):
looks good.

PR #6725, opened 2025-04-25
Update installing_epas_using_local_repository.mdx

yuki-tei:
Hi team,

## What Changed?
Since historically in Repos1.0 edb-repo rpm package generates /etc/yum.repos.d/edb.repo, it is better to change the sample name to the same, or the Repos2.0's enterprisedb-enterprise.repo, so it will not cause confusion to the package name.

Kind Regards,
Yuki Tei


nidhibhammar (approved):
LGTM

PR #6729, opened 2025-04-28
remove redundant note on wal_sender_timeout

EFM-Bobby:
We (ok, I) somehow ended up with two different notes about wal_sender_timeout when using efm's sync standby properties. Only one of these was now correct; I removed the other.


nidhibhammar (approved):

EFM-Bobby:
@nidhibhammar Thank you!

PR #6740, opened 2025-05-02
Don't update redirected links when running checker locally

josh-heyer:
## What Changed?

Automatically updating links that would result in redirects caused unexpected behavior - if the redirect was inappropriate, the link would be updated in a way that might lose information needed to correct the original link.

Relegate this behavior to a new npm script, `links:fix` that update links (`node tools/automation/actions/link-check --fix` will also work when running directly)

Also surface Frontmatter errors, since it's easy to do and those are hard to debug right now.


djw-m (approved):
LGTM

(not tested as too many precious branches in flight)

PR #6754, opened 2025-05-06
MTK 55.11.2: new errors

gvasquezvargas:
## What Changed?
https://enterprisedb.atlassian.net/browse/MTK-1465

### DIFF
--- a/product_docs/docs/migration_toolkit/55/09_mtk_errors.mdx
+++ b/product_docs/docs/migration_toolkit/55/09_mtk_errors.mdx
@@ -117,9 +117,9 @@ The JDBC URL for the target database (EDB Postgres Advanced Server) specified in
 
 To resolve this error, edit the `toolkit.properties` file, specifying valid connection information for the target database in the `TARGET_DB_URL` property. For information about forming a JDBC URL for EDB Postgres Advanced Server, see [Defining an EDB Postgres Advanced Server URL](06_building_toolkit.properties_file/#url-and-parameters-for-an-advanced-server).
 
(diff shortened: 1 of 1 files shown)


product_docs/docs/migration_toolkit/55/09_mtk_errors.mdx
```diff
@@ -117,6 +117,49 @@ The JDBC URL for the target database (EDB Postgres Advanced Server) specified in
...
+DB-604: java.sql.SQLException: ORA-00604: error occurred at recursive SQL level 1
+ORA-01882: timezone region not found
+```
+
+```text
+MTK-11009: Error Connecting Database "Oracle"
+DB-17056: java.sql.SQLException: Non supported character set (add orai18n.jar in your classpath): ZHS16GBK
+```
+
+To resolve this issue you must align the time zone Migration Toolkit is using for the migration with the time zone of the Oracle database. You can do this by specifying the time zone of the Oracle database in the `runMTK.sh` file.
```
zahid75, line 135: Since MTK is also supported on Windows, we need to provide similar guidelines for runMTK.bat script.

zahid75 (changes requested):
While the issue is originally reported on *nix OS and thus the suggested workaround guidelines specifically apply to runMTK.sh, however a similar issue when encountered when running MTK on Windows will require the same work-around.

gvasquezvargas:
I implemented @salman-hameed-code's feedback to the best of my knowledge and am merging this today as the release goes live. If necessary, we can follow up with a quick docs fix.

PR #6758, opened 2025-05-07
Fix links from edb_otel to pg_exporter

tureba:
The paths are valid for `upm-metrics-pg-exporter`, but not for `pg_exporter`.

Since the former is not public-facing, it probably shouldn't be in public docs anyway, so remove it and fix the paths into `pg_exporter` appropriately, which is to remove the `/collector/` component and `0410-pg_activity.yml` instead of `410-pg_activity.yml`.


tureba:
Looks like the repo was moved from Vonng to pgsty, so I changed it too in the last push.

piano35-edb (approved):

PR #6761, opened 2025-05-09
pgd: clarify that "transaction_id" is not accessible in SQL.

ibarwick:
## What Changed?

Also add a missing heading for the unrelated function references which follow immediately after.

DOCS-1528.


djw-m (approved):
LGTM

PR #6765, opened 2025-05-09
pgd: add some notes about pgd_bench, scripts and "-m camo"

ibarwick:
## What Changed?

Explicitly note that pgd_bench does not automatically initiate database transations, so any custom scripts will need to include SQL transaction statements, particularly for running in CAMO mode.

DOCS-1529


djw-m (approved):
LGTM

PR #6768, opened 2025-05-09
Fixes for bdr.run_on...

djw-m:
## What Changed?

Added requested notes


lucianobotti (approved):
Changes look good to me from a Technical POV. All 3 functions are clearly stating that transactions are not being replicated on the other nodes..

PR #6770, opened 2025-05-09
Docs 1426 remove deprecated gucs and functions in pgd 6 0

djw-m:
## What Changed?

Small change, follows up on previous adjustments during the big sweep.


djw-m:
Created in error

PR #6774, opened 2025-05-09
edb_otel: add requirement on EPEL for RHEL 9

tureba:
In RHEL 9, we rely on packages provided by EPEL, such as gRPC or protobuf. Other operating systems may have additional requirements that we may have to add as support of edb_otel extends further.


nidhibhammar:
@piano35-edb, please review and do the needful

nidhibhammar:
@piano35-edb, do we have any progress on this?

tusharahuja (approved):
looks good. Thanks.

nidhibhammar (approved):

PR #6808, opened 2025-05-14
Update monitoring_failover_manager.mdx

ToontjeM:
Changed path from `/usr/efm-<X>/bin` to `/usr/edb/efm-<X>/bin`


EFM-Bobby:
@gvasquezvargas Thank you for checking up -- yes, this change is correct. Good catch @ToontjeM . The existing path has been wrong for a long time and I didn't notice.

PR #6813, opened 2025-05-15
Adds a card entry limit with a counted more prompt

djw-m:
## What Changed?

Small tweaks to full and extra index card layouts to avoid assymetrical "huge" index cards when surfacing large folders of content.

### DIFF
--- a/src/components/card-decks.js
+++ b/src/components/card-decks.js
@@ -27,7 +27,7 @@ const FullCard = ({ card }) => {
         <p className="card-text">{card.description}</p>
 
         <div className="d-grid gap-2">
-          {card.items.map((child) => (
+          {card.items.slice(0, 4).map((child) => (
             <Link
               key={child.path}
               to={child.path}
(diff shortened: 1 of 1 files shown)


djw-m:
Reviewers should check the ai accelerator release notes in the index for an example.

src/components/card-decks.js
```diff
@@ -37,6 +37,11 @@ const FullCard = ({ card }) => {
               {child.interactive && <KatacodaBadge />}
             </Link>
           ))}
+          {card.items.length > 4 && (
```
josh-heyer, line 40: Should "or" this such that if card length is exactly 5, it just adds the 5th item here instead of a "more" link.

djw-m: This is of course the wrong code to do this on, as this is the Full card. I've done the same to the Extra card below. Which one should have the behavior is an open question. I think Extra ATM.

josh-heyer (approved):

PR #6818, opened 2025-05-19
PGD: updates for PGD 6 relnotes

ibarwick:
## What Changed?

Add three items which were noted in the release notes in the BDR source tree, but not here.

DOCS-6550.


djw-m:
Manually merged as concurrent work was ongoing.

PR #6822, opened 2025-05-19
DOCS-1544 Block DDL on SO Nodes

djw-m:
## What Changed?

Should be enough to cover the issue in 5.8?


djw-m:
Closing current PR as functionality of fix has changed in meantime. Will return to it when included in RC/release.

PR #6823, opened 2025-05-20
Add PGD6 release note for BDR-3767.

ibarwick:
## What Changed

Late addition to BDR source tree relnotes, see BDR commit f86b846d4.

Note: seems this update was intended for 5.8.0 as well, but seems to have missed the cut-off.


djw-m:
Closing PR as ticket has changed. Will return to it when functionality is confirmed.

PR #6833, opened 2025-05-22
DOCS-1560-__output__-content-blocks-eat-up-spaces-but-not-co

djw-m:
## What Changed?

Simplified regex so it stops eating spaces.


josh-heyer:
So there were ~310 new failures (code blocks with `__OUTPUT__` dividers that weren't recognized) with the new regex. The *vast* majority of these were code blocks that **only** contain output.

Riffing on our chat the other day, I beefed up the regex to accomplish what you're going for here while still being recognized in existing uses. 

Also added a warning for situations where two __OUTPUT__ markers appear in a code block, as there were a few cases where code blocks were already broken for various reasons with this hallmark.

PR #6867, opened 2025-06-05
Update degrading.mdx

vaijayanti-b:
the examples say group commit can degrade to but it cannot. Sync commit cannot "abort on" and group commit cannot "degrade to". Made changes to reflect this in this page.


djw-m (approved):
LGTM

PR #6870, opened 2025-06-06
Add a link from the high-level HA PEM Patterns 

sonotley:
Add a link from the high-level HA PEM Patterns page to the detailed instructions


nidhibhammar (approved):
Edited the link as required. LGTM

PR #6872, opened 2025-06-10
Fix PDF build hang for PGD6

josh-heyer:
## What Changed?

Avoid hang triggered by nested tab containers
...also add a bunch of useful redirects before this product release goes live
...yes i kinda had the 2nd thing in flight when the 1st thing came up.


djw-m (approved):
LGTN

PR #6896, opened 2025-06-18
Add odbc connector v17 to installer config

josh-heyer:
## What Changed?

...and regenerate installer files 
This should fix the build warnings about conflicting redirects. Should also allow version selection to work properly for the connector docs.


gvasquezvargas (approved):
lgtm! thanks a lot for this!

PR #6899, opened 2025-06-24
Updated sentence about professional services

mlewandowski-edb:
Changed the sentence to not list specific service offerings as I believe the packaged services offerings may have changed.  Instead, I updated the text to more generically describe the types of migration related services are available. This change will reduce the need to maintain a list of packaged services that are susceptible to change.


gvasquezvargas (approved):
LGTM!

PR #6906, opened 2025-07-03
remove duplicated download instructions

fmr-disy:
The script was already downloaded in step 2

## What Changed?

Removed redundant paragraph


fmr-disy:
Hi @nidhibhammar 

FYI: Both links (../installing/assets-op/eks-fix-auto.sh and ../installing/assets-helm/eks-fix-auto.sh) download the same script, the code only differs by an empty line at the end.

PR #6948, opened 2025-08-21
Add clarification to the local.timeout.final property

EFM-Bobby:
The point is that this property defines a maximum amount of time for an agent to wait IF NECESSARY, not a length of time that it waits before doing things normally.


gvasquezvargas:
Closing in favor of https://github.com/EnterpriseDB/docs-editorial/pull/297

PR #6953, opened 2025-08-26
Clarify that only some logging parameters are fixed

nick-ivanov-edb:
The sentence erroneously states that _all_ logging parameters are fixed.

## What Changed?

"all" -> "some"; add link to the fixed parameters section


gvasquezvargas:
Before rebasing, checking with the team whether these kind of changes have to be done upstream to avoid them being overwritten.

PR #6976, opened 2025-09-22
Add max_active_replication_origins on PG 18+

irionr:
The max_active_replication_origins parameter was introduced in PostgreSQL 18 and is not available in earlier versions.

BDR-7069


gvasquezvargas (approved):
LGTM!

PR #6981, opened 2025-10-01
Update edb_loader_control_file_parameters.mdx

yuki-tei:
Hi team,
Greetings. We are contacting you as per below change.

## What Changed?
`The PRESERVE BLANKS option works only with the OPTIONALLY ENCLOSED BY clause. `

This phrase is not correct because, OPTIONALLY ENCLOSED BY clause is only needed when input file is a
Delimiter-separated field data file, but no needed if it is a Fixed-width field data file.

And also, when using Delimiter-separated field data file, only TERMINATED clause is needed and it will work
without OPTIONALLY ENCL.OSED BY clause

Kind Regards,
Yuki Tei


yuki-tei:
Hi team,
Greetings. Is anyone able take a look at this?

Kind Regards,
Yuki Tei

nidhibhammar (approved):

PR #7000, opened 2025-10-28
fix: cnp hibernate status was deprecated

josemore:
## What Changed?
As part of [v1.26.0](https://www.enterprisedb.com/docs/postgres_for_kubernetes/latest/installation_upgrade/#declarative-hibernation-in-the-cnp-plugin), "the hibernate status command has been removed, as its purpose is now fulfilled by the standard status command".


gvasquezvargas (dismissed):
LGTM!

Edit: I think we might need to do this change on https://github.com/EnterpriseDB/cloud-native-postgres-maintenance  according to this [Confluence wiki](https://enterprisedb.atlassian.net/wiki/spaces/DCBC/pages/3805413377/Open+Source+Import+Processes#Transformation%3A). Checking with @josh-heyer .

josh-heyer (approved):

PR #7003, opened 2025-10-30
Changes for supported upgrade paths for 5.9 for review

sumarjee:
Changes for supported upgrade paths for 5.9 for review.

Assuming 5.8 doc is stating things upto 5.8 latest release hence can keep 
https://www.enterprisedb.com/docs/pgd/5.8/upgrades/upgrade_paths/
as is which has To Version with "or later"

Assumption: Here or later does not include 5.9 as this is 5.8 doc.

Fixes BDR-7237

### DIFF
--- a/product_docs/docs/pgd/5.9/upgrades/upgrade_paths.mdx
+++ b/product_docs/docs/pgd/5.9/upgrades/upgrade_paths.mdx
@@ -18,7 +18,7 @@ Upgrades from PGD 4 to PGD 5 are supported from version 4.3.0 or later.
 The following table provides a full list of possible upgrade paths from PGD 4 to PGD 5. Check the "From version" column to confirm the minimum required PGD 4 release necessary before upgrading to your des
 ired PGD 5 target version.
 
-See [Upgrading within 4](/pgd/4/upgrades/upgrade_paths/#upgrading-within-version-4) for instructions on upgrading within PGD 4 to reach the required version.
+See [Upgrading within version 4](/pgd/4/upgrades/upgrade_paths/#upgrading-within-version-4) for instructions on upgrading within PGD 4 to reach the required version.
 
 | From version | To version (range)
 | ----- | -------------- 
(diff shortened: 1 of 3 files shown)


product_docs/docs/pgd/5.9/upgrades/upgrade_paths.mdx
```diff
@@ -9,24 +9,20 @@ All changes within the same major version are backward compatible, lowering the
...
+4.4.0 or later, the following upgrade paths are possible.
 
 | From version | To version |
 | ---- | -- |
-| 4.3.0 | 5.0.0 or later |
-| 4.3.1 | 5.1.0 or later |
-| 4.3.2 | 5.1.0 or later |
-| 4.3.3 | 5.1.0 or later |
-| 4.3.7 | 5.7.0 or later |
+| 4.4.0 | 5.9.0 or later |
```
mwanner: How about PGD 6? Doesn't 4.4.0 also allow to upgrade to PGD 6.1.0?

mwanner (commented):
I'm not sure how to best handle the different versions or whether these docs can really eliminate information about previous versions. I would not trust customers to always select the exact version they are interested in.

And even if, they may select - for example - 4.2.0 from which they happen to be upgrading from.

product_docs/docs/pgd/5.9/upgrades/upgrade_paths.mdx
```diff
@@ -9,24 +9,20 @@ All changes within the same major version are backward compatible, lowering the
 
 You can upgrade from any version 5.x release to a later 5.x release. 
 
-## Upgrading from version 4 to version 5
+## Upgrading from version 4 to version 5.9.0 or later
```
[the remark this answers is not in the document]
sumarjee: @irionr / @mwanner does above looks fine? Then I shall change it accordingly.

product_docs/docs/pgd/5.9/upgrades/upgrade_paths.mdx
```diff
@@ -11,22 +11,26 @@ You can upgrade from any version 5.x release to a later 5.x release.
...
-
-| From version | To version |
-| ---- | -- |
-| 4.3.0 | 5.0.0 or later |
-| 4.3.1 | 5.1.0 or later |
-| 4.3.2 | 5.1.0 or later |
-| 4.3.3 | 5.1.0 or later |
-| 4.3.7 | 5.7.0 or later |
+Generally, we recommend that you upgrade to the latest version 4 release
+before upgrading to the latest version 5 release.
```
mwanner: I'd state "before upgrading to version 5 release". Otherwise, one could read this as "not necessary for older v5 releases than the latest". Nit-picking, but let's try to get this as clear as possible.

PR #7005, opened 2025-10-30
Fix SQL query to use 'id' instead of 'server_id'

asheshv:
Corrected the typo in the suggestion.


nick-ivanov-edb:
> what do you think of using `<id>` or similar here, to better signify this is a placeholder?

It's not a placeholder, it's the actual column name. This PR is a dupe of https://github.com/EnterpriseDB/docs/pull/6999

gvasquezvargas:
Hey @asheshv ! Since this change was already made in https://github.com/EnterpriseDB/docs/pull/6999, after rebasing this branch, the changes diff disappeared (as the changes are already in `develop`), hence why this PR was automatically closed.

PR #7025, opened 2025-11-21
Rename edb_enable_pruning to enable_partition_pruning

amacvar:
This should be a simple edit. this doc page is 2+yrs old Updated parameter name and deprecated notice.

## What Changed?
Rename edb_enable_pruning to enable_partition_pruning


amacvar:
@gvasquezvargas amazing attention to detail 👏🏽 
Ai comments can only add so much 😆 
Since the original doc says:
  This parameter is deprecated from version 15 and later.
We should backport it to 15 & 16 too.
For 100% certainty, hopefully there is a process to verify these changes as mentioned in slack

PR #7060, opened 2026-01-06
Update namespace for spire-server commands

ftouserkani-edb:
% kubectl  get ns | grep -i spire

spire-system                 Active   13d


nfaust-edb (approved):
image

PR #7061, opened 2026-01-08
Update CVE-2025-14038 advisory date

goto-satoru:
Updated: 2025/15/03 
should be a typo.

need to update to:
 Updated: 2025/12/15


nidhibhammar (approved):
LGTM

PR #7081, opened 2026-02-03
Fix indentation in RDS configuration example

ftouserkani-edb:
yaml example needs to be indented appropriately.


jpe442:
Fixed this in both versions in your other PR 7085.

PR #7083, opened 2026-02-04
Update rds.mdx

tswartz-edb:
## What Changed?

Fix incorrect command to delete secret


jpe442 (commented):
@tswartz-edb Does this only apply to IR version? Does it need to be backported to the LTS docs?

PR #7098, opened 2026-02-25
Update conflict_functions.mdx

yuki-tei:
Hi,

## What Changed?
We found by chance that these links are not correct for the current directory structure, which lead to a wrong link when opening without a new tab.

Kind Regards,
Yuki Tei


nidhibhammar:
I have created a JIRA for this to handle a larger number of such fixes within PGD together: https://enterprisedb.atlassian.net/browse/DOCS-3430 and assigned it to @mpfuster

PR #7108, opened 2026-03-09
Update upgrading_major_rolling.mdx

yuki-tei:
Hi,

#### Suggested fix 1
The PGD package name is no longer bdr. Modify to the same syntax as https://www.enterprisedb.com/docs/pgd/latest/lifecycle/deploying/deploy-manual/03-installing-database-and-pgd/

#### Suggested fix 2
Often postgresql.conf is different between major versions, if you not modify manually but copy the original file,
new cluster shall not start due to errors like: "postgresql.conf contains errors"

Kind Regards,
Yuki Tei


yuki-tei:
Hi @josh-heyer ,
Thanks for your modification. I agree to that.

Kind Regards,
Yuki Tei

PR #7146, opened 2026-04-03
efm doesn't use "superuser" database permissions

EFM-Bobby:
This section is about executing scripts as different users, so "superuser" doesn't make any sense here. (When executing sql queries in the database, efm also doesn't use superuser privileges.)


irmabenitezzapata-creator (approved):

irmabenitezzapata-creator:
@EFM-Bobby, the changes are now live in [the documentation](https://www.enterprisedb.com/docs/efm/latest/04_configuring_efm/04_extending_efm_permissions/).

PR #7163, opened 2026-05-01
Clarify that this is specific to TDE

josh-heyer:
## What Changed?

...pg_upgrade is just fine for major version upgrades in general, but cannot be used in this scenario for TDE-enabled databases


pnkrepps (approved):
Perfect! Thanks

PR #7164, opened 2026-05-06
[DBP-1857]

sagaryedida:
Update postgis docs as RHEL 9 No Longer Requires EPEL because EDB providing those deps in the EDb repos only now.

## What Changed?
Removed epel repo installation steps of the EL9 architectures from postgis installation guide


jpe442 (approved):
LGTM

PR #7187, opened 2026-06-30
docs(pgd): Clarify origin_node_group matching semantics for 5.9 and 6.4

irionr:
In 5.9 the GUC-override path matched only the node's own group, while the default_commit_scope path walked ancestors — an asymmetry that caused silent async fallback when SET LOCAL bdr.commit_scope was used with a scope defined on a parent group. In 6.4 the GUC path was fixed to also walk ancestors (commit 776e243b13), so both paths now behave consistently.

Replace the vague "directly or indirectly related" sentence in both versions with precise, version-accurate descriptions. The 5.9 entry explains the route-dependent split and adds a warning callout for the silent-fallback trap. The 6.4 entry describes the unified ancestor-walk behaviour.

BDR-7611

> [!TIP]
> Draft deployment: https://deploy-preview-7187--edb-docs-staging.netlify.app/docs/

### DIFF
--- a/product_docs/docs/pgd/5.9/commit-scopes/commit-scopes.mdx
+++ b/product_docs/docs/pgd/5.9/commit-scopes/commit-scopes.mdx
@@ -40,7 +40,14 @@ If not specified, the system will search for a default commit scope. Default com
 
 If no default_commit_scope is found then the node's GUC, bdr.commit_scope is used. And if that isn't set or is set to `local` then no commit scope applies and PGD's async replication is used.
 
-A commit scope will not be used if it is not local and the node where the commit is being run on is not directly or indirectly related to the origin_node_group.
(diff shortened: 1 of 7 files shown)


mpfuster:
Thanks @irionr , do you need we need to port the note you added for 5.9 all throughout the following versions up to 6.4?

irionr:
> Thanks @irionr , do you need we need to port the note you added for 5.9 all throughout the following versions up to 6.4?

probably yes, the change is only from 6.4 onwards

irionr:
> > Thanks @irionr , do you need we need to port the note you added for 5.9 all throughout the following versions up to 6.4?
> 
> probably yes, the change is only from 6.4 onwards

event hough , we have the latest minors updated I think it's fine like this. 
No port to the other versions.

mpfuster (approved):
LGTM

PR #7200, opened 2026-07-29
Update registration connection string availability note

nick-ivanov-edb:
Clarified the availability of the registration connection string in PEM 10.4 and later.

## What Changed?

Added notes when the new command line argument became available, to address a customer's complaint. See https://edb.slack.com/archives/CDU4GKHL5/p1785319158167079


nidhibhammar (approved):
LGTM

PR #7210, opened 2026-08-11
Update modifying_agent_configuration.mdx

david-olivari:
If max_connection - 0 is unlimited 
why then has to be set to > 1 ?

should it not be anything but 1 
Seems illogical to say > 1 when it is unlimited , but perhaps I miss something in this


josh-heyer (approved):
This seems true, though I would note that "must be set to 0 or be greater than 1" is probably more accurate - 0 is a special value here. But your version is true enough without being pedantic.

PR #7213, opened 2026-08-17
Update authentication details in API documentation

nick-ivanov-edb:
Clarified authentication method for the operation in the API documentation. The current note is confusing.

## What Changed?

The note says the call to obtain the API token "does not require authentication", which is false; it simply use a different authentication method.

### DIFF
--- a/product_docs/docs/pem/10/pem_rest_api/index.mdx
+++ b/product_docs/docs/pem/10/pem_rest_api/index.mdx
@@ -230,7 +230,7 @@ Status Code **201**
 
 !!! info
 
-    Unlike other API calls, this operation does not use token-based authentication; supply username and password instead.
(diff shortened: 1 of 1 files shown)


product_docs/docs/pem/10/pem_rest_api/index.mdx
```diff
@@ -230,7 +230,7 @@ Status Code **201**
 
 !!! info
 
-    This operation does not require authentication
+    Unlike other API calls, this operation does not use token-based authentication; supply username and password instead.
```
gvasquezvargas:
```suggestion
    Unlike other API calls, this operation doesn't use token-based authentication; instead, supply the username and password.
```

gvasquezvargas (approved):
LGTM!

PR #7214, opened 2026-08-19
Update index.mdx

yuki-tei:
Hi,
Greetings. 
## What Changed?
We found only Windows installation page mentions java, which shall be the same prerequisite for both Linux and Windows.
I added the same description to the common section.

Please also consider to discard the duplicate from [this page](https://www.enterprisedb.com/docs/edb_plus/latest/installing/windows/) if my request is accepted.

Kind Regards,
Yuki Tei

### DIFF
--- a/product_docs/docs/edb_plus/41/installing/index.mdx
+++ b/product_docs/docs/edb_plus/41/installing/index.mdx
@@ -17,6 +17,9 @@ navigation:
   - windows
   - configuring_linux_installation
 ---
+## Prerequisites
+
+Before installing EDB\*Plus, you must first install Java version 1.8 or later. For Windows, Java installers and instructions are available at the [Java download page](http://www.java.com/en/download/manual.jsp).
 
 Select a link to access the applicable installation instructions:
 


yuki-tei:
Sorry, my bad. Linux Current Java dependency shall be >= java11, ">=1.8.0" is condition only for windows.
So the alternative is :

Prerequisites
Before installing EDB*Plus, you must first install Java version 11 or later for Linux , and Java 1.8 or later for Windows. For Windows, Java installers and instructions are available at the [Java download page](http://www.java.com/en/download/manual.jsp).

josh-heyer:
I'm gonna decline this one for now, and ask what prompted it? 

See... We had this as part of each individual Linux-platform install page, and [then removed it](https://github.com/EnterpriseDB/docs/pull/3861) - the package should have a Java runtime as a dependency (e.g. Debian & Ubuntu packages have `default-jre`), and thus whatever system you install should automatically obtain the proper dependency and install it; you shouldn't need to go out of your way to do this first (and might not want to if, for instance, this is the ONLY package you've installed that has a Java dependency & you want that to be cleaned up on package removal).

If you ran into issues in spite of this, we should dig into that before attempting to document around it.

yuki-tei:
Hi @josh-heyer 
Thx for your response. Well, the main reason we need it is exactly the auto-dependency of java.
Specifically, RHEL and similar OS's runtime Java dependency will be >=11 but excludes 21. However, you know, the latest java is 21 and many customers rely on it. Then what happens? While our customer alternative config Java to 21, some EDB rpms still depends on for example Java17. So when you'd like to integrate your Java version to 21 and remove Java 17, all EDB products depends on Java will be removed, too.

This is why we have to publish the Java limitation clearly as prerequisite.
Please tell me if this makes sense.

Kind Regards,
Yuki Tei

